PR #1, opened 2020-09-25
Change telemetry prefix

cevian:
Safe to do because this extension has never been published under this name


Harkishen-Singh (approved):
AFAIK, lgtm.

PR #3, opened 2020-09-25
fix: remove autogenerated file

### LINKED ISSUE #2: fails to compile due to hardcoded paths
Hi, just tried to compile this via make, and it's failed with this warning:

```sh
$ make
Makefile:57: warning: overriding recipe for target 'clean'
/usr/lib/postgresql/pgxs/src/makefiles/pgxs.mk:355: warning: ignoring old recipe for target 'clean'
make: *** No rule to make target '/Users/arye/Development/postgres/12.2/install/include/server/postgres.h', needed by 'src/support.o'.  Stop.
```

This [file](https://github.com/timescale/promscale_extension/blob/805d13d893c4c45a823ab0ce009a04c544497c47/src/support.d) seems to be the culprit. Isn't the file `support.d` automatically generated during the `make` process?

grawlinson:
This removes support.d, as it has hardcoded paths specific to one user
and will fail compilation if run on a different computer.

The file has also been added to gitignore for posterity.

Closes #2


cevian (approved):
Thanks a bunch

PR #13, opened 2021-03-29
Prepare for the 0.1.2 release with PG13 support

cevian:


VineethReddy02 (approved):

sql/promscale--0.1.2.sql
```diff
@@ -0,0 +1 @@
+promscale--0.1.1.sql
```
VineethReddy02, line 1: Is this right? `0.1.1.sql` ?

PR #14, opened 2021-06-11
Switch pgx

cevian:

PR #15, opened 2021-06-14
Add vectorSelector

cevian:
a vector selector aggregate has the same semantics as parse.VectorSelector processing
in Prometheus. Namely, for all timestamps ts in the series:
     ts := ev.startTimestamp; ts <= ev.endTimestamp; ts += ev.interval
we return the last sample.value such that the time sample.time <= ts and sample.time >= ts-lookback.
if such a sample doesn't exist return NULL (None).
thus, the vector selector returns a regular series of values corresponding to all the points in the
ts series above.

Note that for performance, this aggregate is parallel-izable, combinable, and does not expect ordered inputs.

Relies on #14

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -17,6 +17,8 @@ type Microseconds = i64;
 const USECS_PER_SEC: i64 = 1_000_000;
 const USECS_PER_MS: i64 = 1_000;
 
+const STALE_NAN: u64 = 0x7ff0000000000002;
+
 // prom divides time into sliding windows of fixed size, e.g.
 // |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |
 // we take the first and last values in that bucket and uses `last-first` as the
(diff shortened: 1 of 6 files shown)


src/lib.rs
```diff
@@ -301,3 +302,262 @@ impl GapfillDeltaTransition {
...
+                    let s = state2.clone();
+                    Some(s.into())
+                }
+                (Some(state1), None) => {
+                    let s = state1.clone();
+                    Some(s.into())
+                } //should I make these return themselves?
+                (Some(state1), Some(state2)) => {
+                    let mut s1 = state1.clone(); // is there a way to avoid if it doesn't need it
+                    s1.combine(&state2);
```
JLockerman, line 381: if you make `combine()` take `&self` and be in charge of allocating you can remove the clone outside it.

JLockerman (approved):
Generally looks good.
I think you can remove some of the `type_builder.rs` code if you want.

PR #29, opened 2021-11-11
Port support.c to Rust, add initial tests

JamesGuthrie:


JamesGuthrie:
@JLockerman I've created this draft PR with preliminary changes which 1. compiles, and 2. generates a schema.

The extension is not functional yet, because the schema generation is incomplete. I intend on adding tests against all of the extension functionality, and using that as a basis to validate that the generated schema is complete and correct.

Could you please take a look and point out any obvious issues that you see?

src/prom_agg.rs
```diff
@@ -0,0 +1,156 @@
...
+            ('2000-01-02 15:15:00 UTC', 150),
+            ('2000-01-02 15:20:00 UTC', 200),
+            ('2000-01-02 15:25:00 UTC', 200),
+            ('2000-01-02 15:30:00 UTC', 150),
+            ('2000-01-02 15:35:00 UTC', 100),
+            ('2000-01-02 15:40:00 UTC', 50),
+            ('2000-01-02 15:45:00 UTC', 0);
+    "#,
+        name = "initialize_prom_agg_test_setup"
+    );
```
JLockerman, line 129: You should be able to run these commands using the SPI at test start, see eg. [this](https://github.com/timescale/timescaledb-toolkit/blob/3e9beb1935bb6d889c55fe82dbbc7c2c785266d2/extension/src/uddsketch.rs#L657-L658).
(I personally prefer to use sub-selects to create the data, [like so](https://github.com/timescale/timescaledb-toolkit/blob/3e9beb1935bb6d889c55fe82dbbc7c2c785266d2/extension/src/time_series/pipeline/aggregation.rs#L725-L737) because it sidesteps issues with table iteration order)

src/raw.rs
```diff
@@ -0,0 +1,87 @@
+#![allow(non_camel_case_types)]
+
+use pgx::*;
+
+// TODO: Is this the right approach to declaring `bytea` and `TimestampTz`?
+extension_sql!(
+    r#"
+        SELECT 1;
+    "#,
```
JLockerman, line 9: You don't need actual contents you can just do
```suggestion
    ""
```
but otherwise 👍

src/support.rs
```diff
@@ -0,0 +1,87 @@
...
+
+#[pg_extern(immutable, strict)]
+#[search_path(@extschema@)]
+pub unsafe fn make_call_subquery_support(input: Internal) -> Internal {
+    make_call_subquery_support_inner(input.to_inner()).internal()
+}
+
+pub unsafe fn make_call_subquery_support_inner(
+    input: Option<Inner<*mut pg_sys::Node>>,
+) -> Inner<*mut pg_sys::Node> {
```
JLockerman, line 14: I _think_ this should be
```suggestion
    input: Option<Inner<pg_sys::Node>>,
) -> Internal {
```
`Inner` is already a pointer, and is guaranteed non-null, so it's not a valid return type for this.
(Honestly, `Inner<...>` is very overfit to aggregate transition states, I suspect just working with `Internal` for both input and output is right)

src/support.rs
```diff
@@ -0,0 +1,87 @@
...
+    let input: *mut pg_sys::Node = input.unwrap().cast();
```
JLockerman, line 15: If you end up using `input: Inner<pg_sys::Node>` I think this should be
```suggestion
    let input: *mut pg_sys::Node = (&mut *input.unwrap()) as _;
```
or if it's `input: Internal` then
```suggestion
    let input: *mut pg_sys::Node = input.unwrap().unwrap() as _;
```

src/support.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    let original_args = PgList::<pg_sys::Node>::from_pg((*expr).args);
+
+    /* Check that these are expressions that don't reference
+    any vars, i.e. they are constants or expressions of constants */
+    if !original_args
+        .iter_ptr()
+        .all(|arg| arg_can_be_put_into_subquery(arg))
+    {
+        return (0 as *mut pg_sys::Node).into();
```
JLockerman, line 46: on a pure rustic-ness level this might be better written as
```suggestion
        return ptr::null_mut::<pg_sys::Node>().internal();
```

JLockerman (commented):
The stuff schema for schema changes look good.
I have some comments on the rustified support function which _might_ be a correctness issue.

Cargo.toml
```diff
@@ -1,26 +1,29 @@
...
 [lib]
-crate-type = ["staticlib"]
+crate-type = ["cdylib", "rlib"]
 
 [profile.release]
 lto = true
 
+[profile.dev]
+lto = "thin"
+
```
JLockerman, line 14: The recommended config is something like

```suggestion
[profile.dev]
panic = "unwind"
lto = "thin"

[profile.release]
panic = "unwind"
opt-level = 3
lto = "fat"
debug = true
codegen-units = 1
```
basically:
1. want to ensure unwinding so we don't abort a postgres process
2. fat lto does a better job cleaning up unused symbols, and helps ensure that we don't leak extra ones into the binary (though it makes compiling slower)
3. leaving in debug info makes it easier to get info when things go wrong
4. everything else is just an optimization setting which tbh isn't important

src/prom_agg.rs
JamesGuthrie: Thanks! I know that I can run the setup in the SPI at start. I was trying to work around the fact that all of the tests have the same setup SQL, and that I want to use the `#[pg_test(error = "time is null")]` macro to assert failure of the function (so I can't put all of the asserts into one function).

Coming from the Java/JavaScript world, I'd rely on some setup functionality which the testing framework provides me. In Rust it doesn't seem that that exists in the built-in testing setup, and I wasn't sure if I should pull in a library for that.

Do you have any recommendations?

I have reworked this test setup a bit. Curious to hear your thoughts.

src/support.rs
JamesGuthrie: I have reworked this to just use `Internal` as you suggested.

src/support.rs
JamesGuthrie: Yes indeed!

src/support.rs
JamesGuthrie: Thanks for this suggestion! I have adopted it.

Cargo.toml
JamesGuthrie: Thanks, I have taken on these changes

src/raw.rs
JamesGuthrie: 👍

src/prom_agg.rs
JLockerman: (haven't looked through the new code yet, so these are fairly generic thoughts)
Common setup code is a perennial issues for the rust testsuite, and there isn't really a canonical solution yet. For stuff like this, I tend towards the simplest and dumbest thing: copy-paste if the tests are logically distinct and just happen to have similar setup, or a function called at the start of each test if the inputs are logically related.

In general my expectation is that most people will encounter test code when they're trying to do a bug fix: they don't really know the surrounding code, and want to be in and out of the codebase as quickly as possible. I think keeping tests cases self-contained assists with this: you only need to read a single function to get context, and it's usually clearer what the unit under test is when using dedicated setup code than when there's common code designed to be general. I often find that code designed to factor out common test tasks can be misplaced: the time spent writing the test is usually less painful than the time spent debugging a patch or tracking down a time-sensitive bug, and after a certain point common code gets complicated enough that it can obfuscate the unit-under-test making it difficult to determine what the test means, and in really bad cases causing test failures do to things unrelated to the features or, even worse, hiding test failures. That said, I know @cevian has differences of opinion with me on testing philosophy, so he may want to comment as well.

(For full clarity and future readers, in this particular case I think a common data table has some unique issues:
1. The aforementioned ordering issue: table scans in postgres return values in an arbitrary order by default, and we have seen cases where the actual order changes across versions, or even just different runs. Since it's very easy to accidentally make a test rely on the specific order of tuples returned, I find tests using such scans to be needlessly brittle (though not enough to outright ban them).
2. Concurrency issues: I'm not entirely sure how this interacts with the pgx shims, but rust tests usually run in parallel by default; accessing the same table from multiple tests _can_ result in weird nondeterminism, though I think that's less likely here where it should be read-only.

)

src/prom_agg.rs
```diff
@@ -107,49 +107,69 @@ CREATE AGGREGATE @extschema@.prom_delta(
...
-            ('2000-01-02 15:20:00 UTC', 200),
-            ('2000-01-02 15:25:00 UTC', 200),
-            ('2000-01-02 15:30:00 UTC', 150),
-            ('2000-01-02 15:35:00 UTC', 100),
-            ('2000-01-02 15:40:00 UTC', 50),
-            ('2000-01-02 15:45:00 UTC', 0);
-    "#,
-        name = "initialize_prom_agg_test_setup"
-    );
+    fn setup() {
```
JLockerman, line 110: ☺️

JLockerman (approved):
Looks good to me!
@cevian did you also want to review?

JamesGuthrie:
@JLockerman Thanks for the great feedback. I have added one more commit to this chain (and rebased against the latest `develop` branch). The newest commit fixes the ported support function, and adds tests. I would appreciate it if you could take a look at that too.

src/support.rs
```diff
@@ -97,5 +96,131 @@ pub unsafe fn arg_can_be_put_into_subquery(arg: *mut pg_sys::Node) -> bool {
...
+                  "Parallel Aware": false,
+                  "Async Capable": false,
+                  "Relation Name": "gfs_test_table",
+                  "Alias": "gfs_test_table"
+                }
+              ]
+            }
+          }
+        ]
+    "#;
```
JLockerman, line 170: I'd inline these `const`s into their respective functions; they seem like they're only useful there anyway, so there's I don't think they need to be in such a high-level scope

JLockerman (approved):
👍

src/support.rs
JamesGuthrie: 👍 done

src/lib.rs
```diff
@@ -19,68 +26,48 @@ const USECS_PER_MS: i64 = 1_000;
...
+    )
+    .internal()
 }
 
-#[pg_extern(immutable, parallel_safe)]
-pub fn prom_rate_transition(
-    state: Option<Internal<GapfillDeltaTransition>>,
-    lowest_time: TimestampTz,
-    greatest_time: TimestampTz,
+pub fn prom_rate_transition_inner(
```
cevian, line 53: do we really need this split between the inner and main function all over the place? Was this just for schema generation?

src/lib.rs
```diff
@@ -314,40 +316,41 @@ impl GapfillDeltaTransition {
...
-        unsafe { TimestampTz::from_datum(args[5].value, args[5].isnull, pg_sys::TIMESTAMPTZOID) }
-            .unwrap_or_else(|| error!("time is null"));
-
-    let value = unsafe { f64::from_datum(args[6].value, args[6].isnull, pg_sys::FLOAT8OID) }
-        .unwrap_or_else(|| error!("value is null"));
-
+) -> Internal {
+    vector_selector_transition_inner(
+        unsafe { state.to_inner() },
+        start_time.into(),
```
cevian, line 330: will errors here be useful? The old code was explicit in what errors were thrown (e.g. error!("start_time is null"))

src/lib.rs
```diff
@@ -568,24 +582,37 @@ impl VectorSelector {
...
+//     deserialfunc = vector_selector_deserialize,
+//     parallel = safe
+// );
+// "#,
+//     name = "create_aggregate_vector_selector"
+// );
+
+#[cfg(test)]
+#[pg_schema]
+pub mod pg_test {
```
cevian, line 609: this looks like it does nothing?

cevian (changes requested):
Some questions and comments. Also where is the generated SQL? Are we thinking  that it won't be committed as a file and re-generated every time?

src/type_builder.rs
```diff
@@ -49,7 +49,7 @@ macro_rules! do_serialize {
                 let len = writer.position().try_into().expect("serialized size too large");
                 ::pgx::set_varsize(writer.get_mut().as_mut_ptr() as *mut _, len);
             }
-            writer.into_inner().as_mut_ptr() as pg_sys::Datum
+            bytea::from(writer.into_inner().as_mut_ptr() as pg_sys::Datum)
```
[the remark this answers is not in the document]
JamesGuthrie: Yes

src/support.rs
```diff
@@ -0,0 +1,230 @@
...
+        )
+        .expect("SQL query failed");
+
+        assert_eq!(
+            result.0,
+            serde_json::from_str::<Value>(
+                // Note: This output can be obtained directly from postgres
+                r#"
+                    [{
+                        "Plan": {
```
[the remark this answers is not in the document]
JamesGuthrie: Me too! Being able to write them helped me to figure out the mechanics of the support function.

Cargo.toml
```diff
@@ -1,25 +1,35 @@
 [package]
-name = "promscale_rs"
-version = "0.3.0"
+name = "promscale"
+version = "0.4.0"
```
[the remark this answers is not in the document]
JamesGuthrie: Yes, can do.

src/lib.rs
```diff
@@ -568,24 +582,37 @@ impl VectorSelector {
...
-    finalfunc = vector_selector_final,
-    combinefunc = vector_selector_combine,
-    serialfunc = vector_selector_serialize,
-    deserialfunc = vector_selector_deserialize,
-    parallel = safe
-);
-"#,
-    name = "create_aggregate_vector_selector"
-);
+// extension_sql!(
```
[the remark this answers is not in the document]
JamesGuthrie: The answer to this question (and a few others that you've asked) is that this PR is not a complete picture of all of the changes which need to be merged to get schema generation working. I wanted to have some eyes on the work that I'm doing and get early feedback without overwhelming reviewers with a code dump.

I commented out this code in this change to allow the extension to be built and installed. In a follow-on change it will be un-commented and tested.

src/lib.rs
```diff
@@ -19,68 +26,48 @@ const USECS_PER_MS: i64 = 1_000;
...
-// prom divides time into sliding windows of fixed size, e.g.
-// |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |  5 seconds  |
-// we take the first and last values in that bucket and uses `last-first` as the
-// value for that bucket.
-//  | a b c d e | f g h i | j   k |   m    |
-//  |   e - a   |  i - f  | k - j | <null> |
 #[pg_extern(immutable, parallel_safe)]
-pub fn prom_delta_transition(
-    state: Option<Internal<GapfillDeltaTransition>>,
+pub fn prom_rate_transition(
```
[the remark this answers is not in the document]
JamesGuthrie: Same as above. I will be refactoring the remainder in follow-on changes.

src/lib.rs
JamesGuthrie: This is a really good question, and I learned something by answering it (or maybe re-learned something I had already forgotten).

The main reason why I refactored this block with the `error!()` calls was because it was manually unpacking the args from fcinfo. When using the `#[pg_extern]` macro this is not necessary. In fact the macro re-writes the function and auto-generates _the same_ unpacking code which was implemented manually, _with_ the error warnings.

It's possible to expand macros and inspect the generated code. The following is the auto-generated code for the new function signature (there is a fair amount of magic going on here, but basically `pgx` autogenerates a `vector_selector_transition_wrapper` function which wraps `vector_selector_transition`):

```
pub unsafe extern "C" fn vector_selector_transition_wrapper(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
    #[allow(non_snake_case)]
    #[allow(unused_variables)]
    #[allow(clippy::missing_safety_doc)]
    #[allow(clippy::redundant_closure)]
    unsafe fn vector_selector_transition_wrapper_inner(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
        let result = {
            let state_ = pgx::pg_getarg::<Internal>(fcinfo, 0usize).unwrap_or_else(|| panic!("{} is null", stringify! {state }));
            let start_time_ = pgx::pg_getarg::<TimestampTz>(fcinfo, 1usize).unwrap_or_else(|| panic!("{} is null", stringify! {start_time }));
            let end_time_ = pgx::pg_getarg::<TimestampTz>(fcinfo, 2usize).unwrap_or_else(|| panic!("{} is null", stringify! {end_time }));
            let bucket_width_ = pgx::pg_getarg::<Milliseconds>(fcinfo, 3usize).unwrap_or_else(|| panic!("{} is null", stringify! {bucket_width }));
            let lookback_ = pgx::pg_getarg::<Milliseconds>(fcinfo, 4usize).unwrap_or_else(|| panic!("{} is null", stringify! {lookback }));
            let time_ = pgx::pg_getarg::<TimestampTz>(fcinfo, 5usize).unwrap_or_else(|| panic!("{} is null", stringify! {time }));
            let value_ = pgx::pg_getarg::<f64>(fcinfo, 6usize).unwrap_or_else(|| panic!("{} is null", stringify! {value }));
            let fcinfo_ = fcinfo;
            vector_selector_transition(state_, start_time_, end_time_, bucket_width_, lookback_, time_, value_, fcinfo )
        };
        result.into_datum().unwrap_or_else(|| panic!("returned Datum was NULL"))
    }
    pg_sys::guard::guard(|| vector_selector_transition_wrapper_inner(fcinfo ))
}

```

JamesGuthrie: Ah, I see that it doesn't `error!()`, but `panic!()`. Do you know if the unpacking with `errror!()` was done on purpose to avoid `panic!()`?

src/lib.rs
JamesGuthrie: I think that schema generation is part of the motivation, yes. As far as I understand, we need the `Internal` type in the API of the method (as an arg and return type), for correct schema generation. Additionally, we have the generic `Inner<T>` wrapper which we use internally. In the main function we translate `Internal` to `Inner<T>`.

I assume that @JLockerman can give a better answer here.

src/raw.rs
```diff
@@ -0,0 +1,84 @@
...
+        }
+    };
+}
+
+#[derive(Clone, Copy)]
+pub struct bytea(pub pg_sys::Datum);
+
+raw_type!(bytea, pg_sys::BYTEAOID, pg_sys::BYTEAARRAYOID);
+
+/// We define our own `TimestampTz` data type which wraps and has interop with the underlying \
```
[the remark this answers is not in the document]
JamesGuthrie: Honestly I'm going to claim innocence here, this is code that I copied from toolkit.

I believe that they're here for schema generation. I'm not sure if there are suitable alternatives, and it's something that I wanted to investigate once the schema generation works cleanly (after the follow-on changes). From what I understand, `pgx` does offer a `TimestampWithTimeZone` type which _might_ be an alternative to our `TimestampTz`, although from the underlying datatypes it doesn't seem to be _quite_ the same thing. It may also offer an alternative to our `bytea` type.

@JLockerman can you perhaps shed some light on this?

src/lib.rs
JamesGuthrie: It does do something, and what it does is interesting.

Basically when you write `pgx` tests, the test functions are executed _inside_ the database. pgx compiles the test function down, creates SQL schema for it, and installs the functions into the db. The functions are all "installed" into the `pg_test` DB schema. The `#[pg_schema]` here ensures that `pub mod pg_test` gets interpreted as "create a new DB schema named `pg_test`", for the test functions to get installed into.

Cargo.toml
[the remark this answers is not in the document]
JamesGuthrie: Actually coming back to this: I _had_ to bump the version number because otherwise the way that PGX auto-generates and installs the schema conflicted with the existing SQL schema scripts. I chose to bump it by another minor version because I was unsure how backwards compatible we would be able to remain.

I will bump the patch version as you suggested. We can always bump the minor before release if we realise that it's necessary.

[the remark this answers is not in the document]
cevian: Yeah let's always bump be patch until release. We can always choose to bump more at realease time but not really go backwards

src/lib.rs
cevian: I don't think that's purposeful panic should be fine. and the errors look ok. Can we just add a test for this.

cevian (approved):
I think this is ok to merge as long as we address the other issues in followup PRs.

PR #37, opened 2021-11-29
Add tests for aggregate functions

JamesGuthrie:


JamesGuthrie:
This change restructures our aggregate function structure and adds tests. I have one more change that I'm working on in the area of schema generation to finish up this refactoring.

src/lib.rs
```diff
@@ -1,609 +1,14 @@
-use std::collections::VecDeque;
```
cevian, line 1: do we need this file anymore? seems mostly empty now

src/aggregates/mod.rs
```diff
@@ -0,0 +1,189 @@
...
+
+#[pg_extern()]
+pub fn prom_delta_final(state: Internal) -> Option<Vec<Option<f64>>> {
+    prom_delta_final_inner(unsafe { state.to_inner() })
+}
+pub fn prom_delta_final_inner(
+    state: Option<Inner<GapfillDeltaTransition>>,
+) -> Option<Vec<Option<f64>>> {
+    state.map(|mut s| s.to_vec())
+}
```
cevian, line 28: why are these here and not in prom_delta.rs

src/aggregates/mod.rs
```diff
@@ -0,0 +1,189 @@
+use std::collections::VecDeque;
```
cevian, line 1: should this be called gapfill_delta.rs?

src/aggregates/vector_selector.rs
```diff
@@ -0,0 +1,404 @@
...
+            first_bucket_max_time: start_time,
+            last_bucket_max_time: last_bucket_max_time,
+            end_time: end_time,
+            bucket_width: bucket_width,
+            lookback: lookback,
+            elements: vec![None; num_buckets as usize],
+        }
+    }
+
+    fn combine(&mut self, other: &Inner<VectorSelector>) {
```
cevian, line 188: is this function ever hit in the tests? i.e. are we ever using a parallel plan in tests? can we force parallel plans?

src/aggregates/vector_selector.rs
```diff
@@ -0,0 +1,404 @@
...
+#[pg_extern(immutable, parallel_safe)]
+pub fn vector_selector_final(
+    state: Internal, /* Option<Inner<VectorSelector>> */
+) -> Option<Vec<Option<f64>>> {
+    let state: Option<Inner<VectorSelector>> = unsafe { state.to_inner() };
+    state.map(|s| s.to_pg_array())
+}
+
+#[pg_extern(immutable, parallel_safe)]
+pub fn vector_selector_serialize(state: Internal) -> bytea {
```
cevian, line 77: is this ever hit in tests?

cevian (changes requested):
Mostly looks good. Some cleanup requested. Also is there an easy/standard way to get code coverage stats on this repo?

src/lib.rs
JamesGuthrie: Ya, we do. It serves as the entrypoint into the lib, kinda like a `main.go` does for a go binary.

src/aggregates/mod.rs
JamesGuthrie: Good question. `prom_delta_final` is used as the `finalfunc` for all aggregates. Perhaps it would make more sense to rename it, but I don't want to break compatibility (yet).

src/aggregates/vector_selector.rs
JamesGuthrie: It isn't, thanks for catching that! I've added a test which forces execution of a parallel query, which should hit this function.

src/aggregates/vector_selector.rs
JamesGuthrie: I _think_ (based on my reading of the PG docs on this topic https://www.postgresql.org/docs/14/xaggr.html#XAGGR-PARTIAL-AGGREGATES) that now that we have a parallel execution test, this function should now also be hit.

src/aggregates/mod.rs
JamesGuthrie: Kinda maybe, but probably not directly. Rust's module system expects there to be a module for `aggregates` somewhere (because we have defined `mod aggregates;` in `lib.rs`. The `aggregates` module implementation can either be in `aggregates/mod.rs` (this file) or in `aggregates.rs` (which doesn't exist). This means that we can't rename `mod.rs` to `gapfill_delta.rs`. We could move the `GapfillDelta...` stuff into a separate source file named `gapfill_delta.rs`, but I'm not sure that that adds any value.

@bitwalker maybe you could add your thoughts here. I'm not a Rust guru.

JamesGuthrie:
@cevian I looked into code coverage. It's possible to get a code coverage report. It does require using the `nightly` compiler though. I have added an issue for it.

src/aggregates/mod.rs
bitwalker: As a general rule, I use the `<module>/mod.rs` convention whenever there are multiple submodules, and use `mod.rs` to hold module-global stuff (like custom error/result types, constants, etc), and to re-export submodule types/functions/etc., that the module aims to provide; it's also a good place to define any `macro_rules!` macros used within the module. For anything else, I use the single-file `<module>.rs` convention. 

The approach @JamesGuthrie used here feels right to me. I do agree though that the `GapfillDeltaTransition` bits probably make sense to define in a submodule; the rest of it seems like it is in the right place, at least for now.

src/aggregates/vector_selector.rs
JamesGuthrie: Based on the coverage report that I was able to generate, these functions are not being hit by the tests. I'm having some issues with debugging this at the moment though. Will keep digging.

src/aggregates/vector_selector.rs
```diff
@@ -0,0 +1,493 @@
...
+use pgx::error;
+
+use crate::aggregate_utils::in_aggregate_context;
+use crate::aggregates::{Milliseconds, STALE_NAN, USECS_PER_MS};
+use serde::{Deserialize, Serialize};
+
+use crate::palloc::{Inner, InternalAsValue, ToInternal};
+use crate::raw::{bytea, TimestampTz};
+
+// a vector selector aggregate has the same semantics as parse.VectorSelector processing
```
bitwalker, line 12: We should probably change comments like these to be doc comments, i.e. `///` rather than `//`, and then control whether or not they show up in the generated crate docs using doc attributes, e.g. `#[doc(hidden)]`

bitwalker (approved):
I'm still getting caught up, but this LGTM!

src/aggregates/mod.rs
cevian: ok let's just put in a code-comment todo to rename in the future

src/aggregates/vector_selector.rs
cevian: Have you tried breaking the logic and seeing if tests fail? It may be a misreport by the coverage tool

src/aggregates/mod.rs
cevian: Ok I do think we should move GapfillDeltaTransition but will leave it up to you in this PR.

cevian (approved):
All this code is looking lots better.

src/aggregates/vector_selector.rs
JamesGuthrie: The coverage was correct. I managed to figure out how to force PG into performing a partial aggregation, and it surfaced a bug. Partial aggregation on `vector_select` is broken (also in previous releases - I tested `0.2.0` and `0.3.0` and has probably never worked. That might be interesting in itself, because obviously we're not generating queries in `promscale` that _would_ use it.

src/aggregates/vector_selector.rs
JamesGuthrie: Agree. I have done this already in some places. IMO it requires more than just converting a `//` to a `///`, as I think that the quality of the code documentation needs to be improved. I've opened #42 for this.

src/aggregates/vector_selector.rs
JamesGuthrie: And I found out how to fix it.

src/aggregates/mod.rs
JamesGuthrie: done

cevian (approved):
LGTM

bitwalker:
@JamesGuthrie I've rebased develop on latest master, and rebased your branch for this PR on top of develop. As mentioned in a different PR, I think we should try and avoid merging changes to master without going through develop first, and base all of our feature branches on develop (like you've done here), that way we can avoid needing to do the rebase dance like this in the future - but at least now we can merge your ongoing work into develop, and then merge develop into master when we're ready to cut a new release. If you have other local branches based on develop, just be aware that you'll need to rebase them on develop as well.

JamesGuthrie:
Awesome, thanks for cleaning that up @bitwalker

PR #38, opened 2021-11-29
Run lint checks on PRs and merge commits

JamesGuthrie:


antekresic (commented):
Is there a reason why fmt and clippy cannot be in the same job? Is it mainly to run them in parallel?

JamesGuthrie:
Having them in separate jobs allows them to be run in parallel, and gets separate failures for each.

.github/workflows/lint.yaml
```diff
@@ -0,0 +1,61 @@
+name: lint
+on:
+  pull_request:
+  push:
```
bitwalker, line 4: **Question:** Is there any particular reason to run this workflow on anything but PRs? I'm assuming we're requiring changes to be made via PR, so if that's the case, we can require that these lint status checks pass before merge, at which point master/develop will be guaranteed to pass the checks. Running them on a schedule wouldn't be necessary for the same reason; and unless we're triggering this workflow via the GitHub API, the `workflow_dispatch` event won't be used.

JamesGuthrie: I've adjusted this to only run the workflow on PRs.

src/lib.rs
```diff
@@ -438,10 +441,10 @@ impl VectorSelector {
...
             first_bucket_max_time: start_time,
-            last_bucket_max_time: last_bucket_max_time,
-            end_time: end_time,
-            bucket_width: bucket_width,
-            lookback: lookback,
+            last_bucket_max_time,
+            end_time,
+            bucket_width,
+            lookback,
             elements: vec![None; num_buckets as usize],
```
antekresic, line 448: Disclamer: I don't know much about Rust style but should these be grouped by style? Named and unnamed parameters.

antekresic (approved):

bitwalker (approved):

bitwalker:
@JamesGuthrie Should this branch be based on develop instead? Since a lot of the ongoing work is flowing from `feature branch -> develop -> master` right now, and we probably want this PR in `develop` too, it might make things a bit more straightforward

PR #44, opened 2021-12-06
ci: run tests on pull_request events

bitwalker:
Add GHA workflow to run tests for the extension against Postgres 14. We can add support for other Postgres versions once we've sorted out how to handle test output differences that currently cause the tests to fail when run under versions other than 14.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,47 @@
+name: ci
+on:
+  pull_request:
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        postgres:
+        - version: "14"
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install latest nightly
+        uses: actions-rs/toolchain@v1
+        with:
+            toolchain: stable
+            override: true
+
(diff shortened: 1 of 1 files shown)


JamesGuthrie (approved):
LGTM!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,39 @@
...
+
+      - uses: Swatinem/rust-cache@v1
+
+      - name: Install cargo-pgx
+        uses: actions-rs/cargo@v1
+        with:
+          command: install
+          args: cargo-pgx
+
+      - name: Initialize pgx
```
JamesGuthrie, line 28: It would be great if we could cache this step somehow. Also for the linting workflow.

Harkishen-Singh (approved):

.github/workflows/ci.yaml
bitwalker: Good call, I think we should be able to cache the `~/.pgx` directory and be more or less good to go, I'll give that a shot here in a few

JamesGuthrie:
I retriggered the build and caching worked. Fantastic 🎉

PR #45, opened 2021-12-06
Schema generation and binary compatibility

JamesGuthrie:
This series of changes provides:

- schema generation with PGX
- binary backwards compatibility of the shared object
- explicitly configure extension schema to `_prom_ext`

The schema can be generated with `cargo pgx schema -f pg14`, and should be effectively API-compatible with the schema that we used in the 0.3.0 release, except for the omission of the following snippets:

```PLpgSQL
DO $$
    DECLARE
        current_version TEXT;
        original_message TEXT;
    BEGIN
        BEGIN
            SELECT version
              INTO STRICT current_version
              FROM public.prom_schema_migrations;
        EXCEPTION WHEN OTHERS THEN
            GET STACKED DIAGNOSTICS original_message = MESSAGE_TEXT;
            RAISE EXCEPTION 'could not determine the version of the Promscale connector that was installed due to: %', original_message
            USING HINT='This extension should not be created manually. It will be created by the Promscale connector and requires the connector to be installed first.';
            RETURN;
        END;

        IF current_version = '' THEN
            RAISE EXCEPTION 'the requisite version of the Promscale connector has not been installed'
            USING HINT='This extension should not be created manually. It will be created by the Promscale connector and requires the connector to be installed first.';
        END IF;
    END
$$;
```

and

```PLpgSQL
-- Set the search path to one that will find all the definitions provided by the
-- connector. Since the connector can change the schemas it stores things
-- in we cannot just hardcode the searchpath, instead we switch the search path
-- based on the schemas declared by the connector.
DO $$
    DECLARE
        ext_schema TEXT;
        prom_schema TEXT;
        metric_schema TEXT;
        catalog_schema TEXT;
        new_path TEXT;
    BEGIN
        SELECT value FROM public.prom_installation_info
         WHERE key = 'extension schema'
          INTO ext_schema;
        SELECT value FROM public.prom_installation_info
         WHERE key = 'prometheus API schema'
          INTO prom_schema;
        SELECT value FROM public.prom_installation_info
         WHERE key = 'catalog schema'
          INTO catalog_schema;
        new_path := format('public,%s,%s,%s', ext_schema, prom_schema, catalog_schema);
        PERFORM set_config('search_path', new_path, false);
    END
$$;
```

The motivation for hard-coding the schema name is as follows:

> Without the schema configured in the control file, pgx's generated SQL
is not schema-qualified, meaning that objects are created in the schema
defined by CREATE EXTENSION.

> Promscale creates the extension in the _prom_ext schema, and expects all
objects to be in that schema.

> By hard-coding this in the control file, we explicitly ensure that the
extension's objects are only installed into the _prom_ext schema.

Binary backwards-compatibility is provided by declaring `extern C` functions and the required `pg_finfo` info functions.


JamesGuthrie:
@cevian could you perhaps weigh in on the SQL snippet which I have removed. I have added some workarounds which allow the extension to be installed without promscale being present. This has been necessary to even be able to test the extension in isolation. We can probably work around this to make the extension testable, but make it fail if Promscale is not running. What do you think?

cevian:
@JamesGuthrie I think that message is nice for people from a UX perspective. However I don't feel strongly and we can add an option to bypass for tests. But, honestly not sure we can require a flow where we in don't install Promscale. reason is our current plan to create view in the extension for the tracing stuff. Those views will require certain Promscale version installed (with access to those tables). So, I don't see how we can make this truly independent of one another.

src/schema/setup.rs
```diff
@@ -0,0 +1,39 @@
+use pgx::*;
+
+// First we need to define some objects which should be defined by `promscale`
+// This is mostly so that the extension can be installed standalone, which is
+// convenient for development.
+extension_sql!(
+    r#"
+CREATE OR REPLACE FUNCTION swallow_error(query text) RETURNS VOID AS
```
bitwalker, line 8: This should specify a schema we own right? Just like we do further down below. Otherwise it might potentially replace, then drop, a function of the same name. I haven't checked to see if we're implicitly creating this in a schema other than `public`, but it caught my eye anyway.

src/support.rs
```diff
@@ -86,6 +85,26 @@ pub unsafe fn make_call_subquery_support(input: Internal) -> Internal {
...
+);
+
+/// Backwards compatibility
+#[no_mangle]
+pub extern "C" fn pg_finfo_make_call_subquery_support() -> &'static pg_sys::Pg_finfo_record {
+    const V1_API: pg_sys::Pg_finfo_record = pg_sys::Pg_finfo_record { api_version: 1 };
+    &V1_API
+}
+
+#[allow(dead_code)]
```
bitwalker, line 101: Shouldn't this function be marked `#[no_mangle]`? If not, and the compiler is seeing it as dead code, then it's going to be stripped from the shared library when optimized, and even if it isn't, the symbol will be impossible to reference by name (Rust symbols are mangled, so `make_call_subquery_support` will be something like `_ZN7example26make_call_subquery_support17h9fb4fbedaed4f767E` in the binary. If it's marked `#[no_mangle]`, then rustc will always include it in the final binary, and the symbol will be referenceable as `make_call_subquery_support`. You can also remove the `pub` visibility modifier in this case, since its not actually called anywhere in our rust code (AFAICT), and rustc will always include `#[no_mangle]` extern functions.

bitwalker (changes requested):
Overall looks good! Just one suggestion on where to place some of the compatibility shims, and a couple small things that look like they might be unintentional, but may just be me missing some context - let me know if that's the case and I'll change this to an approval.

src/support.rs
JamesGuthrie: doh. That was a total oversight. Nice catch.

src/schema/setup.rs
JamesGuthrie: 👍

src/aggregates/mod.rs
```diff
@@ -17,16 +17,35 @@ pub const STALE_NAN: u64 = 0x7ff0000000000002;
...
+    &V1_API
+}
+
+#[no_mangle]
+unsafe extern "C" fn gapfill_delta_final(fcinfo: pg_sys::FunctionCallInfo) -> pg_sys::Datum {
+    prom_extrapolate_final_wrapper(fcinfo)
+}
+
+#[no_mangle]
+pub extern "C" fn pg_finfo_prom_delta_final_wrapper() -> &'static pg_sys::Pg_finfo_record {
```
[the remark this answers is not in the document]
JamesGuthrie: 👍

src/aggregates/mod.rs
```diff
@@ -17,16 +17,35 @@ pub const STALE_NAN: u64 = 0x7ff0000000000002;
...
 pub const USECS_PER_MS: i64 = 1_000;
 
-// TODO: Rename this function.
-// It is used by multiple aggregates, not only `prom_delta`. Renaming it will
-// break binary compatibility with old extension installs, so we need to figure
-// that out first.
 #[pg_extern()]
-pub fn prom_delta_final(state: Internal) -> Option<Vec<Option<f64>>> {
-    prom_delta_final_inner(unsafe { state.to_inner() })
+pub fn prom_extrapolate_final(state: Internal) -> Option<Vec<Option<f64>>> {
```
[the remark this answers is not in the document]
JamesGuthrie: 👍

bitwalker (commented):
Looks good! Not sure why the test run bails out the way it does, we might want to run tests with `RUST_BACKTRACE=1` or something

bitwalker (approved):

cevian:
I think we must have 

```-- Set the search path to one that will find all the definitions provided by the
-- connector. Since the connector can change the schemas it stores things
-- in we cannot just hardcode the searchpath, instead we switch the search path
-- based on the schemas declared by the connector.
DO $$
    DECLARE
        ext_schema TEXT;
        prom_schema TEXT;
        metric_schema TEXT;
        catalog_schema TEXT;
        new_path TEXT;
    BEGIN
        SELECT value FROM public.prom_installation_info
         WHERE key = 'extension schema'
          INTO ext_schema;
        SELECT value FROM public.prom_installation_info
         WHERE key = 'prometheus API schema'
          INTO prom_schema;
        SELECT value FROM public.prom_installation_info
         WHERE key = 'catalog schema'
          INTO catalog_schema;
        new_path := format('public,%s,%s,%s', ext_schema, prom_schema, catalog_schema);
        PERFORM set_config('search_path', new_path, false);
    END
$$;```

I don't think we can get rid of that since we need to set the search_path. 

Also once we move more things into the extension we'll have more than on schema so setting the schema in the control file may not be good. I believe you can specify the schema explicitly on a per-object basis in pgx (using the schema trait in pg_extern).

File: src/aggregates/gapfill_delta.rs [hunk 5]
```diff
@@ -1,8 +1,31 @@
...
+) -> Option<Vec<Option<f64>>> {
+    state.map(|mut s| s.as_vec())
+}
+
+/// Backwards compatibility
+#[no_mangle]
+pub extern "C" fn pg_finfo_gapfill_delta_final() -> &'static pg_sys::Pg_finfo_record {
+    const V1_API: pg_sys::Pg_finfo_record = pg_sys::Pg_finfo_record { api_version: 1 };
+    &V1_API
+}
```
cevian, line 22: should we consider making this a macro?

JamesGuthrie: I thought about it, but I didn't think it would be worth it for code that we would probably drop when we have versioned SOs anyway.

JamesGuthrie:
@cevian 
> I don't think we can get rid of that since we need to set the search_path.

If the extension was standalone then I would agree with you, but as-is, we set all of the search paths correctly from Promscale. Setting them here is irrelevant and doesn't seem to do anything.

> Also once we move more things into the extension we'll have more than on schema so setting the schema in the control file may not be good. I believe you can specify the schema explicitly on a per-object basis in pgx (using the schema trait in pg_extern). But I'll leave this up to you.

Yes, you're right. I think that we will have to look into this a bit more. Can we punt those changes to the next step?

On a related note, I was finally able to test this change against promscale's e2e tests, and it passed.

cevian:
| If the extension was standalone then I would agree with you, but as-is, we set all of the search paths correctly from Promscale. Setting them here is irrelevant and doesn't seem to do anything.

cevian:
> If the extension was standalone then I would agree with you, but as-is, we set all of the search paths correctly from Promscale. Setting them here is irrelevant and doesn't seem to do anything.

Yes you are right we don't need that.

> Yes, you're right. I think that we will have to look into this a bit more. Can we punt those changes to the next step?
Yes we can punt for now.

> On a related note, I was finally able to test this change against promscale's e2e tests, and it passed.
Great. Can we test this with an older version of promscale that still uses the old functions to make sure binary compatibility still works?

cevian (approved):
LGTM. Would like to have the backward binary compatibility tested (manually is fine)

JamesGuthrie:
I tested the binary backward compatibility, it worked fine.

FYI (and for future reference) what I did was:
```
docker run --platform linux/amd64 -e POSTGRES_PASSWORD=password -p 5432:5432 -d --rm --name promscale-ext timescaledev/promscale-extension:0.1.1-2.0.1-pg12
# in promscale dir
./dist/promscale  -db-ssl-mode=disable -db-user=postgres -db-password=password -db-name=postgres --migrate only
docker exec -i promscale-ext psql -U postgres -d postgres <<'EOF'
            CREATE TABLE gfs_test_table(t TIMESTAMPTZ, v DOUBLE PRECISION);
            INSERT INTO gfs_test_table (t, v) VALUES
                ('2000-01-02T15:00:00+00:00',0),
                ('2000-01-02T15:05:00+00:00',12),
                ('2000-01-02T15:10:00+00:00',24),
                ('2000-01-02T15:15:00+00:00',36),
                ('2000-01-02T15:20:00+00:00',48),
                ('2000-01-02T15:25:00+00:00',60),
                ('2000-01-02T15:30:00+00:00',0),
                ('2000-01-02T15:35:00+00:00',12),
                ('2000-01-02T15:40:00+00:00',24),
                ('2000-01-02T15:45:00+00:00',36),
                ('2000-01-02T15:50:00+00:00',48);
            ANALYZE;
            CREATE OR REPLACE FUNCTION arbitrary_function(key text, value text)
                RETURNS text
                AS $func$
                    SELECT key || value
                $func$
                LANGUAGE SQL STABLE PARALLEL SAFE
                SUPPORT make_call_subquery_support;
EOF
docker exec -i promscale-ext psql -U postgres -d postgres <<'EOF'
            EXPLAIN  SELECT * FROM gfs_test_table WHERE arbitrary_function('const','value') = 'constvalue';
EOF
docker run -i --platform linux/amd64 -d --rm --name promscale-ext-new timescaledev/promscale-extension:0.3.1-2.5.0-pg14 cat -
docker cp promscale-ext-new:/usr/local/lib/postgresql/promscale.so ./
docker cp ./promscale.so promscale-ext:/usr/local/lib/postgresql/promscale.so
docker exec -i promscale-ext psql -U postgres -d postgres <<'EOF'
            EXPLAIN  SELECT * FROM gfs_test_table WHERE arbitrary_function('const','value') = 'constvalue';
EOF
```

PR #52, opened 2022-01-05
Move build logic into GitHub actions

JamesGuthrie:


JamesGuthrie:
The spirit of this change has been adopted in the most recent docker image building.

PR #53, opened 2022-01-05
Ensure that pgx generates schema on Alpine Linux

JamesGuthrie:
The missing *-unknown-linux-musl targets meant that pgx's magic linker
script was not used, breaking schema gen on Alpine


JamesGuthrie:
We _could_ merge this change now, but if we re-run `cargo pgx schema -f <pgversion>` then it will overwrite this file, because pgx hasn't released this upstream yet.

bitwalker:
Since you've removed the `--force-default` flag, that shouldn't be an issue right? It's only a problem if one runs that manually, unless I've misunderstood. I'll leave it up to you, but I think we're clear to merge this

JamesGuthrie:
I've created a branch named `promscale-staging` in the https://github.com/timescale/pgx repo to track any of our changes which we're trying to upstream. It currently has the same contents as upstream `develop`. Switching this change to use that branch gives us the alpine fix which has been upstreamed but which is not in master yet. With that, this change is ready to be merged.

Cargo.toml
```diff
@@ -26,11 +26,11 @@ pg_test = ["serde_json"]
 
 [dependencies]
 bincode = "1.3.1"
-pgx = "0.2.5"
```
bitwalker, line 29: There's a slightly easier way to do this generally, using the `[patch]` section:

```toml
[patch.crates-io]
pgx = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
pgx-macros = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
pgx-tests = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
```

When placed in a workspace `Cargo.toml` (or in a single crate's `Cargo.toml`), you only have to specify the override in one place, and it will replace all uses of those crates in any of the workspace crates. In this case it's 6 of one, half a dozen of the other; but probably worth following the convention anyway.

JamesGuthrie: TIL. Thank you!

bitwalker (approved):
LGTM!

jgpruitt (approved):

PR #57, opened 2022-01-11
feat: provide multi-platform packaging automation

bitwalker:
See the individual commits for more details on the changes included here. I've broken out the workflow from the underlying Dockerfiles and packaging script to keep the changes smaller and easier to review.

This PR introduces a GitHub Actions workflow and supporting infrastructure to automatically generate a GitHub Release when new tags are pushed to this repository. It handles the following:

* Build packages for each combination of supported OS and Postgres version controlled by a build matrix defined in the workflow
* Automatically parse release notes from `CHANGELOG.md` and attach them to the release
* Create the GitHub Release, with all of the package artifacts attached
* Upload all of the package artifacts to PackageCloud

The workflow also runs on PRs that modify any of the release tooling.

**NOTE:** The generation of release notes from the CHANGELOG file assumes that prior to pushing the new tag, that the CHANGELOG has been updated to move all of the `UNRELEASED` items into a section for the tag. If no section exists for the tag that was pushed, then the items in the unreleased section are used instead. This should give us reliably useful release notes without having to do anything manually, and we can always manually update the release notes if needed.

@JamesGuthrie  I've ported over the post-install automation that is used by the toolkit just to get the versioned .so stuff in place, just to test things out, but if you have work in this area going on we can make changes to this PR accordingly.

---

Originally we were thinking we would add our scripts to the `release-build-scripts` repo, but there are a number of awkward aspects to doing so, and this approach feels both cleaner and easier to work with and maintain for our team. Just to list what comes to mind:

* Despite being used by both the database and toolkit projects, only a minimal amount of the scripts in that repo are actually shared, there does not appear to be much benefit to sharing a single repo for this purpose as a result.
* Any change to the scripts in that repo on the master branch trigger almost all of the workflows to run, rebuilding virtually every package managed there. This is both wasteful of our CI budget (including things like available cache, artifact storage, etc.), but extremely time consuming.
* The structure of the scripts in that repo is not well documented and is awkward to maintain. The solution implemented in this changeset consists of a single Dockerfile for each package type (deb/rpm) and a single script that handles the common aspects of generating a package across all platforms and shakes out to ~300 LOC total while preserving all of the same functionality.

It's possible the other teams can benefit from moving their release workflows out of that repo, but since the database and toolkit teams are more tightly bound to one another, it may be more involved for them. For us, there is just no reason to tie our release process up with theirs.


JamesGuthrie (dismissed):

.github/workflows/release.yaml
```diff
@@ -0,0 +1,123 @@
+name: release
+on:
+  pull_request:
+    paths:
+    - '.github/workflows/release.yaml'
+    - 'dist/*'
+    - 'tools/*'
+  push:
+    branches:
+    - main
```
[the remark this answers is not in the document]
bitwalker: We can remove that for the time being, but my feeling is that we'd want to publish these sort of "nightly" snapshot packages somewhere, maybe to GitHub Packages or something. They aren't really nightly either, since they only get built on a merge to main, so I guess `-snapshot` is a better convention for those, if we decide to produce them.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,123 @@
...
+
+    - name: Setup QEMU
+      uses: docker/setup-qemu-action@v1
+
+    - name: Setup Docker Buildx
+      uses: docker/setup-buildx-action@v1
+
+    - name: Build Package
+      uses: docker/build-push-action@v2
+      with:
```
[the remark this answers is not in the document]
bitwalker: Yeah for sure! I plan to enable that (that's why I added the qemu action), but wanted to see how long a full run of the workflow takes before doing so, in case it's pushing the limits for a single workflow run. If we can do it all in a single workflow run, that would be ideal. If it takes too long to build all of the packages in a single run, then I'd probably do all of the amd64 packages in the primary workflow, and then trigger a second run on completion of the first that appends the arm64 packages to the release that gets created.

I'll make sure that gets addressed by the time this is merged though!

Makefile
```diff
@@ -1,42 +1,86 @@
...
+ifndef $(DOCKERFILE)
+	$(error Unsupported OS_NAME! Expected one of debian,ubuntu,centos,rhel,fedora)
+endif
+	docker buildx build --load \
+		--build-arg OS_NAME=$(OS_NAME) \
+		--build-arg OS_VERSION=$(OS_VERSION) \
+		--build-arg PG_VERSION=$(PG_VERSION) \
+		--build-arg RUST_VERSION=$(RUST_VERSION) \
+		-t $(RELEASE_IMAGE_NAME) \
+		-f $@ \
```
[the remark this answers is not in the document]
bitwalker: Ah nice catch! Forgot to update this after refactoring the target from referencing the actual filename

dist/deb.dockerfile
```diff
@@ -0,0 +1,108 @@
+# syntax=docker/dockerfile:1.3-labs
+
+## Build base system
+ARG OS_NAME=debian
+ARG OS_VERSION
+FROM ${OS_NAME}:${OS_VERSION}-slim as base
```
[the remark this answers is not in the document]
bitwalker: Ah right, it's probably not essential to use `-slim` anyway considering the number of packages we end up installing on top of the base

.github/workflows/release.yaml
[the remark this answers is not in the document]
bitwalker: I've modified this a bit so that we still run the workflow on pushes to main, but no release is created and no packages are uploaded anywhere (except as artifacts associated to the workflow run, which might be handy). We can remove this entirely if it's not useful, or if it is expensive to run. I'll leave it up to you all on whether or not to keep it in.

tools/package
```diff
@@ -0,0 +1,236 @@
...
+echo "Building.."
+cd "${ROOT_DIR}"
+
+PATH="${pg_config_dir}:${PATH}" cargo pgx package
+
+echo "Post-processing build.."
+cargo run --manifest-path tools/post-install/Cargo.toml -- --dir "target/release/promscale-pg${pg_version}"
+
+echo "Packaging.."
+fpm --log info -s dir -t "${pkg_type}" \
```
cevian, line 223: should we add --meta-file flag and a debian directory with the copyright file like https://github.com/timescale/release-build-scripts/blob/master/debian/timescaledb-base-deb/debian/copyright

tools/changelog
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env bash
+
```
cevian, line 2: is this script taken from somewhere? Also should the file be named  changelog.sh

.github/workflows/release.yaml
```diff
@@ -0,0 +1,150 @@
...
+          RUST_VERSION=${{ env.RUST_VERSION }}
+        tags: ${{ steps.metadata.outputs.image }}
+
+    - name: Extract Package
+      env:
+        PLATFORM: ${{ matrix.os.name }}/${{ matrix.arch.platform }}
+      run: |
+        mkdir -p artifacts
+        container="$(docker create --platform "${PLATFORM}" "${{ steps.metadata.outputs.image }}")"
+        docker cp ${container}:/dist/${{ steps.metadata.outputs.filename }} artifacts/${{ steps.metadata.outputs.outfile }}
```
cevian, line 102: how are you sure steps.metadata.outputs.filename matches the filename created by Build Package step since that's never passed into that step?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,150 @@
...
+        docker rm -f "${container}"
+        ls -la artifacts/*
+
+    - name: Upload Artifact for GitHub Release
```
cevian, line 106: Nit: this won't be associated with a release 
```suggestion
    - name: Upload Artifact for Job
```

cevian (commented):
Looks good. This isn't a full review. Will finish tomorrow. But looks like really solid engineering work.

Questions:
1) Is it worth running lintian on the deb package
2) Can we test these packages somehow as part of the build runs?

tools/post-install/src/main.rs
```diff
@@ -0,0 +1,361 @@
+use std::{
```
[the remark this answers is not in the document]
bitwalker: Yeah exactly, I'm not a fan either. This is basically just a placeholder for the work @JamesGuthrie was doing in this direction

tools/package
bitwalker: Sure, can do!

tools/changelog
bitwalker: No, just something handwritten for our use case. As a general rule, I prefer to omit the extension on scripts which are marked executable and meant to be used as utility programs, as this and `package` are; and add the extension when they are meant to be sourced/included in other scripts (i.e. like libraries). It's a subjective rule, but I find it works well since it doesn't matter to the consumer of the script what it's written in, and virtually every editor I know of understands how to syntax highlight shell scripts with the shebang, so there is little objective reason to require it. That said, it doesn't matter too much to me, so if we prefer a different approach, just let me know

.github/workflows/release.yaml
bitwalker: It's implicit, since the rules to construct that output are the same rules used by the `package` script. We can be more explicit, but I also wanted to make sure that `fpm` was producing outputs that matched my assumptions on all the targets. Now that I've validated those assumptions, I can pass the output file name using the appropriate option to the `package` script so we always know they match up.

tools/post-install/src/main.rs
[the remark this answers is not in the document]
JamesGuthrie: If it's just intended to be a placeholder, can you remove it from this PR? It's not necessary for the packaging work.

tools/package
```diff
@@ -0,0 +1,236 @@
...
+    fi
+fi
+
+# If no package format was set, select an appropriate one for the current OS
+if [ -z "$pkg_type" ]; then
+    case "$os_name" in
+        debian|ubuntu)
+            pkg_type="deb"
+            ;;
+        rhel|centos|fedora)
```
JamesGuthrie, line 163: Does this actually work with RHEL or Fedora?

.github/workflows/release.yaml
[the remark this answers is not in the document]
JamesGuthrie: 👍

JamesGuthrie:
@bitwalker I added a couple of commits to get the build working. Please squash them into your existing commits.

tools/post-install/src/main.rs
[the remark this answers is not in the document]
bitwalker: @JamesGuthrie It's serving a function, by "placeholder" I mean that we have other work in progress to handle versioned artifacts in a cleaner way, so it is intended that the `post-install` crate will be removed when that work is done, or at least when its ready to serve the same function. The reason it's worth having in the interim is to avoid packaging the extension without versioned artifacts.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,155 @@
...
+        - name: linux
+          distro: centos
+          version: 7
+          codename: 7
+          pkg_type: rpm
+        - name: linux
+          distro: centos
+          version: 8
+          codename: 8
+          pkg_type: rpm
```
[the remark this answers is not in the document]
bitwalker: Not that I'm aware of. I have my own informal rules for deciding things like that, but this initial matrix was just to make sure both package types work, likewise across distros.

CentOS announced last year that it is changing from its previous RHEL-like release process to a more Fedora-like model (which is completely the opposite of why people run CentOS, but it is what it is), which is why 8 is marked EOL even though 7 isn't.

In general though, I would recommend that we support the latest stable release, and the latest LTS release for the distros we care about. We may also be able to ship only Debian releases for all Debian-derivative distros if they install Postgres in the same location, in which case we could support the most commonly encountered versions (Debian 9, 10, and 11, which correspond to Ubuntu 18.04, 20.04, and 21.04). With CentOS it would be 7, RHEL would be 7 and 8, and future CentOS and Fedora would track the latest release (since they both follow a continuous release model). Again though, we could check and see if a single .rpm package could work across all RHEL/Fedora-like distros, and try to simplify our build matrix that way.

As far as supporting arm64 builds, I suggest we only produce those for the latest stable release in each OS, since those are predominately going to be used for development (ARM servers are still very rare in practice). I'd guess the vast majority of such users would be people running Docker on an M1-series laptop. We can use exclusions in the build matrix to avoid trying to produce arm64 versions for all combinations.

tools/package
bitwalker: Yes, I've tested on both.

.github/workflows/release.yaml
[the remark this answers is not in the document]
JamesGuthrie: 👍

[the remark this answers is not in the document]
cevian: what is the difference between this matrix and TimescaleDB itself?

.github/workflows/release.yaml
cevian: Cool. Yeah I think that would be better

tools/changelog
cevian: ok that works.

cevian: ok that works

cevian (commented):
Ok finished review. Everything looks good. Things left to do/discuss:

- [x]  lets finalize the matrix. Is it the same as timescaleDB?
- [x]  the copyright stuff for debian
- [x] do we want to run lintian?
- [x] can we test these artifacts in an automates way at least for some subset?

tools/package
JamesGuthrie: Hmmm, okay. Maybe I misunderstood how this is supposed to work. I assumed that it would be possible to build this on fedora 35, like so:

```
> docker buildx build --platform linux/amd64 --load \                    ~/Development/promscale_extension
        --build-arg OS_NAME=fedora \
        --build-arg OS_VERSION=35 \
        --build-arg PG_VERSION=14 \
        --build-arg RUST_VERSION=1.57.0 \
        -t promscale-extension-package:0.3.1-fedora35-pg14 \
        -f dist/rpm.dockerfile \
        .
        
```

.github/workflows/release.yaml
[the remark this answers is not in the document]
bitwalker: TimescaleDb builds for PostgreSQL 12-14, on the following (x86_64 only):

- Debian 9, 10, 11
- Ubuntu 18.04, 20.04, 21.04, and 21.10
- Centos 7 and 8

I'm not sure why they build a separate package for Ubuntu 21.10.

The matrix in the PR currently targets Debian 11, Ubuntu 21.04, and Centos 7 and 8.

tools/package
bitwalker: Ah oops, forgot to update the rpm dockerfile with my fedora tweaks. Fixed.

Easiest way to build a target is `OS_NAME=fedora OS_VERSION=35 PG_VERSION=14 ARCH=x86_64 make release`, little easier to type out than the full `docker buildx build` invocation :)

bitwalker:
@cevian The copyright file is present, and lintian is run by default when built in Docker. I commented on the thread about the build matrix, but bottom line is that I think we can probably shoot for the same set of targets initially (except maybe Ubuntu 21.10, not sure there is any value in that).

As far as testing goes, we can easily set up automatic testing that covers each supported Postgres version and package types (e.g. deb/rpm). I don't think we need to worry about testing every single distro combination, we really just want to make sure the supported package types properly install/upgrade/uninstall, and that they are known to work on each of our supported Postgres versions. 

For a first pass, I think validating that we can install the package via the package manager, then create the extension in a Postgres session and call at least one of the functions, should give us plenty of confidence. The next phase would test upgrades by grabbing a previous release and testing that the new release can upgrade from that previous release successfully - this would be a bit more complicated to orchestrate, but would give us a lot of confidence that our upgrade process and migrations are working as expected. I think we can tackle that once we have the upgrade infrastructure in place though.

Issue #61, opened 2022-01-14
Test pgx upstream first-class aggregate support and give feedback


JamesGuthrie:
pgx is working on upstream first-class aggregate support. We should test it out and give feedback where we can.

PR #62, opened 2022-01-14
Switch to first-class aggregate support in pgx

JamesGuthrie:
Note this is a WIP, and removes backwards-compatibility functions.

Part of https://github.com/timescale/promscale_extension/issues/61


src/aggregates/prom_increase.rs
```diff
@@ -1,109 +1,50 @@
...
-                )
-                .into();
-                state
-            });
-
-            state.add_data_point(sample_time, sample_value);
-
-            Some(state)
-        })
+pub struct PromIncrease {}
```
bitwalker, line 6: NOTE: You can declare zero-sized structs using just `pub struct PromIncrease;`

src/aggregates/prom_increase.rs
```diff
@@ -1,109 +1,50 @@
...
+
+#[pg_aggregate]
+impl Aggregate for PromIncrease {
+    type State = Option<GapfillDeltaTransition>;
+    type Args = (
```
bitwalker, line 11: It's unfortunate that this trait requires an extra allocation just to pack the arguments into a tuple in order to call the `state` function, but I guess that was a tradeoff they were forced to make. I wonder what the impact of that is in a hot aggregate.

bitwalker (commented):
Definitely cleaner! I think we'll want to try and benchmark the overhead of wrapping arguments in a tuple, since I could see that potentially being an issue in hot aggregates, but we'll have to concoct a decent benchmark setup for that (we probably should be thinking about that anyway, especially with performance being one of our main objectives this year). 

Rust does have facilities for that, but we may have to upstream some support in pgx for it, since it would need to build on top of the pgx infra, just like how testing does.

PR #64, opened 2022-01-21
WIP: Generate SQL migration scripts

JamesGuthrie:
This change:

1. Moves SQL from promscale repo to the extension
    This will need to be updated and coordinated with the promscale repo
2. Provides a skeleton for the SQL generation
    This currently _only_ provides support for migrations, no idempotent/preinstall separation.
    
I don't think that it's ready to merge at all yet, but it's a starting point around which we can coordinate work in the SQL move effort.


migration-table.sql
```diff
@@ -0,0 +1,7 @@
+-- migration-table.sql
+CREATE SCHEMA IF NOT EXISTS _ps_catalog;
+CREATE TABLE IF NOT EXISTS _ps_catalog.migration(
+  id TEXT NOT NULL PRIMARY KEY
```
cevian, line 4: `id` implies an integer type to me. I'd rename to `filename` or `name`

migration/001-utils.sql
```diff
@@ -0,0 +1,50 @@
+ --perms for schema will be addressed later;
```
cevian, line 1: I think there is some confusion. These are the preinstall files not the migration files. Migration files are under 
`promscale/pkg/migrations/sql/versions/dev/**` By the way, I don't think we need to port the current migration files over. But we should have some files that test the migration infrastructure.

migration-table.sql
```diff
@@ -0,0 +1,7 @@
...
+, applied_version TEXT
```
cevian, line 5:
```suggestion
, applied_at_version TEXT
```

templates/migration-wrapper.sql
```diff
@@ -0,0 +1,25 @@
+-- {{filename}}
```
cevian, line 1: This looks good

cevian (changes requested):

templates/migration-wrapper.sql
```diff
@@ -0,0 +1,25 @@
...
+DO
+$block1$
```
jgpruitt, line 3: Thinking maybe we should use more obscure string delimiters so we don't potentially clash with the body of the migration. Maybe `$outer_migration_block$` and `$inner_migration_block$`?

jgpruitt (changes requested):

build.rs
```diff
@@ -0,0 +1,115 @@
...
+
+#[derive(Template)]
+#[template(path = "idempotent-wrapper.sql", escape = "none")]
+struct IdempotentWrapper<'a> {
+    filename: &'a str,
+    body: String,
+}
+
+fn main() {
+    println!("cargo:rerun-if-changed=build.rs");
```
bitwalker, line 25: **Suggestion:** No need to specify this for `build.rs` or `Cargo.lock`, as any change to either of those triggers recompilation/re-run anyway (for `build..rs` that's always the case. I haven't confirmed, but I think changes to `Cargo.lock` only trigger build script recompilation if build dependencies change, as you'd expect). If you want to re-run the build script when dependencies change because of expected changes to codegen output, then its better to specify the path to the codegen outputs themselves, rather than trigger a re-run if any of our dependencies change.

I'm fairly sure you'll need to be more specific about the `migration` and `templates` folders. If you want to rebuild/rerun the build script when their contents change, then you have to specify each file individually. At least that was the case when I last investigated the semantics of rerun-if-changed.

build.rs
```diff
@@ -0,0 +1,115 @@
...
+    println!("cargo:rerun-if-changed=Cargo.lock");
+    println!("cargo:rerun-if-changed=migration");
+    println!("cargo:rerun-if-changed=templates");
+    println!("cargo:rerun-if-changed=migration-table.sql");
+
+    let mut sql = include_str!("migration-table.sql").to_string();
+    let rendered_sql = render_sql().expect("unable to render migrations");
+    sql.push_str(rendered_sql.as_str());
+
+    let mut out_path = env::current_dir().expect("current directory invalid");
```
bitwalker, line 35: **Suggestion:** It's better to use `env::var("OUT_DIR")` which Cargo sets during compilation, and ensures that the output is placed in the compiler's output directory. You can then do things like `include!(concat!(env!("OUT_DIR"), "foo.txt"))` in your Rust code if needed.

If you want to source control the outputs, by placing the, relative to the crate directory that contains `Cargo.toml`, use `env::var("CARGO_MANIFEST_DIR")`, which is guaranteed to point to the correct location. There are a number of useful Cargo env vars you may find handy described [here](https://doc.rust-lang.org/cargo/reference/environment-variables.html)

build.rs
JamesGuthrie: From https://doc.rust-lang.org/cargo/reference/build-scripts.html#rerun-if-changed:

> If the path points to a directory, it will scan the entire directory for any modifications.

And based on this:

> If the build script inherently does not need to re-run under any circumstance, then emitting cargo:rerun-if-changed=build.rs is a simple way to prevent it from being re-run (otherwise, the default if no rerun-if instructions are emitted is to scan the entire package directory for changes). Cargo automatically handles whether or not the script itself needs to be recompiled, and of course the script will be re-run after it has been recompiled. Otherwise, specifying build.rs is redundant and unnecessary.

I understand that we can either remove all `rerun-if-changed` entries and have the build script re-run on any change in the project dir, or specify only those things which should cause the build script to re-run. I would prefer to keep the `rerun-if-changed` entries, to reduce recompile time.

build.rs
JamesGuthrie: Yes, I am aware of `env::var("OUT_DIR")`. There are two considerations that led to the current implementation:

1. pgx doesn't (currently) support passing anything other than a string literal to `extension_sql_file`, so the input path has to be hard-coded. I have a PR open to enable support: https://github.com/zombodb/pgx/pull/407.

2. I don't really want to source control the output, but IMHO it would be nice to have the SQL which is output in this step in an "easy to find and view" location. `OUT_DIR` is a pain to find (there are multiple, with a hash appended, have to step through a wide tree of directories to find it). You're right, `CARGO_MANIFEST_DIR` is a better solution than `current_dir`, thanks for pointing that out!

templates/migration-wrapper.sql
JamesGuthrie: 👍

migration-table.sql
JamesGuthrie: 👍

migration-table.sql
JamesGuthrie: 👍

build.rs
bitwalker: Ah cool! Good to know that behavior is available for directories!

RE: build script. Since you _are_ emitting some rerun-if-changed entries, emitting one for build.rs is redundant/unnecessary as mentioned in the section you quoted. It's only beneficial to emit it if you don't otherwise have any of those instructions, and you don't want to rerun the build script when any file in the crate changes, and instead only want to rerun it when just the build script changes. If you have any `rerun-if-changed` instructions other than that, then Cargo reruns it automatically when the build script is changed/recompiled.

build.rs
```diff
@@ -0,0 +1,116 @@
...
+}
+
+#[derive(Template)]
+#[template(path = "idempotent-wrapper.sql", escape = "none")]
+struct IdempotentWrapper<'a> {
+    filename: &'a str,
+    body: String,
+}
+
+fn main() {
```
cevian, line 24: I'd add a comment above this to document what this does

build.rs
```diff
@@ -0,0 +1,116 @@
...
+    println!("cargo:rerun-if-changed=templates");
+    println!("cargo:rerun-if-changed=migration-table.sql");
+
+    let mut sql = include_str!("migration-table.sql").to_string();
+    let rendered_sql = render_sql().expect("unable to render migrations");
+    sql.push_str(rendered_sql.as_str());
+
+    let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let mut out_path = PathBuf::from(manifest_dir);
+    out_path.push("migration.sql");
```
cevian, line 37: nit: migration.sql seems too generic. Maybe generated-migration.sql or hand-written-migration.sql?

cevian (commented):
Mostly looks good. Some nits but this is really solid work

sql/promscale--0.0.0--0.5.0.sql
```diff
@@ -0,0 +1 @@
+promscale--0.5.0.sql
```
cevian, line 1: where in the code is this symlink created?

migration-table.sql
```diff
@@ -0,0 +1,7 @@
+-- migration-table.sql
+CREATE SCHEMA IF NOT EXISTS _ps_catalog;
```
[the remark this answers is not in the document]
JamesGuthrie: done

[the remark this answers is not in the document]
JamesGuthrie: Hmm... actually not done. I will discuss with you over Slack.

templates/idempotent-wrapper.sql
```diff
@@ -0,0 +1,8 @@
+
+-- {{filename}}
+DO
```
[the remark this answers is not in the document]
JamesGuthrie: Originally I figured it would make sense to have the migration and idempotent scripts structurally similar. I've since added a `RAISE LOG` line to this block, which does mandate the wrapping.

build.rs
JamesGuthrie: done

build.rs
JamesGuthrie: done

build.rs
JamesGuthrie: I like hand-written-migration. It's long, but it's explicit.

sql/promscale--0.0.0--0.5.0.sql
JamesGuthrie: It isn't... Yet. This has been the thorn in my side for the last couple of days.

JamesGuthrie:
Closing, all changes have been merged in separate PRs.

PR #69, opened 2022-01-27
Prune non-idempotent objects from pgx-generated SQL

bitwalker:
I've reworked this PR from the ground up. It now depends on changes to pgx, which you can review [here](https://github.com/zombodb/pgx/pull/410).

The changes here take advantage of the new `#[to_sql]` attribute I introduced, which provides us two options for customizing the SQL generator for specific entities:

1. We can disable SQL generation entirely for an entity, which I've done in this PR to the `GapfillDeltaTransition` type
2. We can handle generating the SQL ourselves, or redirect the generated SQL to an alternative output file. This is not demonstrated in this PR, though I've tested it; but we may want to consider that approach as a means for generating upgrades (there are a number of possibilities here, but the simplest would be generating a separate file for each non-idempotent entity, so that we can easily use those to check our handwritten SQL as well as write new migrations when things change).

As an unrelated aside: I noticed we have some `extension_sql!` calls still hanging around that do `CREATE OPERATOR`, we should remember to extract those to migrations.


JamesGuthrie:
I like how concise this is with `#[to_sql(false)]`

src/aggregates/prom_rate.rs
```diff
@@ -88,7 +88,7 @@ unsafe extern "C" fn gapfill_rate_transition(fcinfo: pg_sys::FunctionCallInfo) -
 // for proper behavior the input must be ORDER BY sample_time
 extension_sql!(
     r#"
-CREATE AGGREGATE @extschema@.prom_rate(
+CREATE OR REPLACE AGGREGATE @extschema@.prom_rate(
```
JamesGuthrie, line 91: for completeness you missed this change on `prom_delta` and `prom_increase`.

JamesGuthrie (approved):

cevian (approved):
LGTM

PR #70, opened 2022-01-27
Use pg_sys::TimestampTz in place of custom TimestampTz wrapper

JamesGuthrie:
pgx used to emit pg_sys::TimestampTz as BIGINT in SQL. This has been
fixed upstream to emit TIMESTAMP WITH TIME ZONE. With that upstream
change, we can jettison our custom TimestampTz wrapper.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -254,9 +254,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "3.0.9"
+version = "3.0.13"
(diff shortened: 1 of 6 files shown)


bitwalker (approved):
Awesome! Glad this got upstreamed 🎉 

/cc @JLockerman

jgpruitt (approved):

cevian (approved):

PR #76, opened 2022-02-07
Document vector_selector pushdown functionality

JamesGuthrie:


bitwalker (approved):
Looks great! Definitely something I had wished for when I started, good to see it documented so well :)

PR #79, opened 2022-02-08
Specify desired pgx revision in Dockerfile

JamesGuthrie:
If the revision is not specified, then docker can cache the build step
and get stuck on an old revision of the branch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,9 +322,9 @@ dependencies = [
 
 [[package]]
 name = "convert_case"
-version = "0.4.0"
+version = "0.5.0"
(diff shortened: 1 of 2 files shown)


jgpruitt (approved):

cevian (approved):
This is ok but is there any way to have this be defined only once in the Cargo.toml and derived from that?

JamesGuthrie:
> This is ok but is there any way to have this be defined only once in the Cargo.toml and derived from that?

I don't think so. The install of the `cargo-pgx` utility is independent of the `Cargo.toml`.

PR #80, opened 2022-02-08
Add back pgextwlist to built docker image

JamesGuthrie:
At some point in the future we will mark the extension as trusted, and
then we will be able to drop pgextwlist. In the meantime, the Promscale
test suite fails if pgextwlist is not present. By default, the test
suite runs against the `timescaledb-ha` docker image. This means that
until we have the trusted extension in `timescaledb-ha` we either need
to make Promscale able to handle either having pgextwlist, or trusted
extension. Adding pgextwlist back seems to be the simpler option.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,6 +64,22 @@ COPY templates/ /build/promscale/templates/
 RUN --mount=type=cache,uid=70,gid=70,target=/build/promscale/.cargo/registry \
     make package
 
+FROM timescale/timescaledb:${TIMESCALEDB_VERSION}-${PG_VERSION_TAG} as pgextwlist-builder
+
+RUN \
+    apk add --no-cache --virtual .build-deps \
+        gcc \
+        libc-dev \
+        make \
+        git \
+        clang \
+        llvm
+
+RUN \
+    git clone --branch v1.12 --depth 1 https://github.com/dimitri/pgextwlist.git /pgextwlist && \
+    cd /pgextwlist && \
+    make
+
 # COPY over the new files to the image. Done as a seperate stage so we don't
 # ship the build tools.
 FROM timescale/timescaledb:${TIMESCALEDB_VERSION}-${PG_VERSION_TAG}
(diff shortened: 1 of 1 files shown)


bitwalker (approved):

jgpruitt:
>The ALTER EXTENSION ... ADD|DROP command is intentionnaly not supported so as not to allow users to modify an already installed extension. That means that it's not currently possible to CREATE EXTENSION ... FROM 'unpackaged';.

Won't we need the functionality listed above to change the ownership of existing objects to the extension on day 1?

JamesGuthrie:
> Won't we need the functionality listed above to change the ownership of existing objects to the extension on day 1?

From pg13 release notes:

> Remove support for upgrading unpackaged (pre-9.1) extensions (Tom Lane)

> The FROM option of CREATE EXTENSION is no longer supported. Any installations still using unpackaged extensions should upgrade them to a packaged version before updating to PostgreSQL 13.

JamesGuthrie:
Actually ignore my previous comment, you're completely right. At some point (soon) we will need to switch over from pgextwlist. But until we've published a version of the extension which doesn't rely on it, we're going to need to keep it. If that makes sense?

cevian (approved):
Whatever makes the transition easier is fine with me. Question: what happens if we mark the extension trusted now?

bitwalker:
@cevian I don't think it results in any significant changes to the way things work today, it should just continue to work as it does, though we probably would need to modify the install instructions a bit.

The bigger issue is that PG13 does indeed prevent us from upgrading from the old-style extension to the new one using the `CREATE EXTENSION .. FROM ..` approach. I'll do some testing to see if it is possible to run ALTER EXTENSION commands from within the "normal" upgrade script, which I think should work (after all, my understanding is that the script run for `CREATE EXTENSION .. FROM ..` _is_ just a normal upgrade script). If that works, we could try and just handle the upgrade that way instead, and just detect if the version that we're upgrading from is old-style (or if the objects exist, but the extension doesn't, to handle the case where users hadn't previously used the extension).

Either way though, I'd make the extension trusted sooner rather than later.

JamesGuthrie:
I'm actually not sure if this is really the right way to go. I will convert this into a draft until I've decided what to do with it.

PR #82, opened 2022-02-10
Improve docker image build time

JamesGuthrie:
This is a common rust/cargo trick: create a new empty project with the
actual Cargo.{toml,lock} files and build it. This pre-builds all of the
dependencies, a lot of which can be reused in the following steps,
reducing the amount of code compiled when updating source files.

On my machine it reduces rebuild time from ~250s to 80s.


bitwalker (approved):

jgpruitt (approved):
Very cool!

PR #83, opened 2022-02-10
feat: set up unpackaged upgrade scripts

bitwalker:
This PR adds scripts which handle upgrading from pre-0.5.0 (either Promscale with the extension, or without) to 0.5.0. I've tested the prelude of the script (i.e. the part with all of the `ALTER EXTENSION .. ADD` statements) on a local install with the latest Promscale, and that part works as expected, resulting in all Promscale objects being removed if `DROP EXTENSION promscale CASCADE;` is run.

The upgrade path looks like so:

* On an existing Promscale installation without the extension installed, one needs to run `CREATE EXTENSION promscale VERSION '0.0.0'; ALTER EXTENSION promscale UPDATE TO '0.5.0';` to bring all of the Promscale objects under the extensions ownership, and then apply all of the extension-only SQL migrations on top of that base.
* On an existing Promscale installation with the pre-0.5.0 extension already installed, one only needs to run `ALTER EXTENSION promscale UPDATE to '0.5.0';` and resulting in the `promscale--0.3.0--0.5.0.sql` script being run, which is essentially the same script as `promscale--0.0.0--0.5.0.sql`.

The caveat here is that Promscale needs to orchestrate the upgrade, and should ensure that before doing so, that all of the Promscale-managed objects are up to date, as the upgrade scripts here assume that the schema matches that of the latest release (if there are changes made between now and then, these scripts will need to be modified accordingly).

Another catch is that this PR relies on weaving together hand-written SQL, the `migration-table.sql` contents, and the pgx-generated SQL in a way that differs from the default `promscale--0.5.0.sql` script. Specifically, we need these upgrade scripts to insert rows for all of the migrations, since the objects already exist and are already up to date (per the assumptions I laid out above). To do this, we first do all of the `ALTER EXTENSION .. ADD` statements, then we create the `schema_migrations` table, then we insert all of the rows in that able corresponding to the migrations we know have been applied, and finally we run the contents of the generated SQL script. Currently though, the generated script includes the contents of `migration-table.sql` (which we want on a clean install, but not on an upgrade from pre-0.5.0). In short, what is here probably should be split out into files that get woven together by our SQL generator to produce the correct set of install and upgrade scripts, rather than having the `promscale--0.0.0--0.5.0.sql` and `promscale--0.3.0--0.5.0.sql` files be hand-written - but I'll leave that up to you all.


JamesGuthrie:
We're continuing this work in spirit, so I will close this PR.

PR #84, opened 2022-02-11
Bump pgx to 0.3.1

JamesGuthrie:

PR #85, opened 2022-02-11
Fix broken build on develop branch

JamesGuthrie:
I missed the fact that this breaks the build :/


JamesGuthrie:
Ya, I rebased it against (and merged it into) develop, and the tests fail now because the GapFillDelta type is missing.

JamesGuthrie:
I'm not sure that that fix is the right solution either. With that the tests would pass, but the extension would be broken when built and released. We need to move this generated non-idempotent into a migration script. I have some changes in the sql migration branch which work towards that. I will continue with that there.

JamesGuthrie:
This won't build without the 0.3.1 bump. Closing in place of #84.

PR #99, opened 2022-03-03
Do not attempt to drop a metric which is not present

JamesGuthrie:
TODO: We have a code freeze on SQL, and there is no test coverage for
this behaviour.


JamesGuthrie:
This has been covered in a different PR.

PR #102, opened 2022-03-07
Move sql scripts from promscale repo

JamesGuthrie:
Move all SQL files, with history, from the promscale repo. The promscale
history was rewritten at commit 914c2dfe on the master branch with the
following command:

```
git filter-repo \
  --path pkg/migrations/sql/idempotent \
  --path pkg/migrations/sql/preinstall \
  --path-rename pkg/migrations/sql/idempotent:migration/idempotent \
  --path-rename pkg/migrations/sql/preinstall:migration/migration
```


cevian (approved):
This is amazing. I didn't know you could preserve history like this. The history can be super-useful. Would like some other opinions however. @jgpruitt ?

jgpruitt (approved):
This is pretty sweet

PR #103, opened 2022-03-07
Wrap migration SQL in migration do-block, emit with pgx

JamesGuthrie:
This change introduces a build script which wraps the contents of each
SQL file in the `migrations` directory in a DO block which conditionally
applies the migration if it has not yet been applied to the database.

The build script outputs its contents to `migration.sql`, which in turn
is included in the generated SQL which pgx outputs.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,122 @@
+use crate::SqlType::{Idempotent, Migration};
+use askama::Template;
+use std::fs::File;
+use std::io::{Error, Read, Write};
+use std::path::{Path, PathBuf};
+use std::string::String;
+use std::{env, fs};
+
+#[derive(Template)]
+#[template(path = "migration-wrapper.sql", escape = "none")]
+struct MigrationWrapper<'a> {
+    filename: &'a str,
+    version: &'a str,
+    body: String,
+}
+
+#[derive(Template)]
+#[template(path = "idempotent-wrapper.sql", escape = "none")]
+struct IdempotentWrapper<'a> {
+    filename: &'a str,
+    body: String,
+}
+
+const MIGRATION_FILE_NAME: &str = "hand-written-migration.sql";
+
+/// This build script reads the SQL files placed in the `migration/{idempotent,migration}`
+/// directories, wraps each file in its respective template (in the `templates` directory), and
+/// outputs the whole contents as one long SQL file (`MIGRATION_FILE_NAME`) in the root directory.
+fn main() {
+    println!("cargo:rerun-if-changed=Cargo.lock");
+    // Note: directories are not traversed, instead Cargo looks at the mtime of the directory.
+    println!("cargo:rerun-if-changed=migration");
+    println!("cargo:rerun-if-changed=migration/migration");
+    println!("cargo:rerun-if-changed=migration/idempotent");
+    println!("cargo:rerun-if-changed=templates");
+
+    let mut sql = include_str!("migration/migration-table.sql").to_string();
+    let rendered_sql = render_sql().expect("unable to render migrations");
+    sql.push_str(rendered_sql.as_str());
+
+    let manifest_dir = env::var("CARGO_MANIFEST_DIR").unwrap();
+    let mut out_path = PathBuf::from(manifest_dir);
+    out_path.push(MIGRATION_FILE_NAME);
+    let mut file = File::create(out_path).unwrap();
+    file.write_all(sql.as_bytes())
+        .expect("unable to write file contents");
+}
+
+fn render_sql() -> Result<String, Error> {
+    let mut migration_dir = env::current_dir()?;
+    migration_dir.push("migration");
+    let mut result = String::new();
+    for sql_type in &[Migration, Idempotent] {
+        let mut path = migration_dir.clone();
+        path.push(dir_for_type(sql_type));
+        result += render_file(path, sql_type)?.as_str();
+    }
+    Ok(result)
+}
+
+fn dir_for_type(sql_type: &SqlType) -> String {
+    match sql_type {
+        Migration => "migration".to_string(),
+        Idempotent => "idempotent".to_string(),
+    }
+}
+
+fn render_file(path: PathBuf, sql_type: &SqlType) -> Result<String, Error> {
+    let mut result = String::new();
+    let mut paths: Vec<PathBuf> = fs::read_dir(path)?.map(|f| f.unwrap().path()).collect();
+    paths.sort();
+    paths.into_iter().for_each(|path| {
+        result += wrap(&path, sql_type).as_str();
+    });
+    Ok(result)
+}
+
+#[derive(Copy, Clone, Debug)]
+enum SqlType {
+    Migration,
+    Idempotent,
+}
+
+fn wrap(path: &Path, sql_type: &SqlType) -> String {
+    let filename = path
+        .file_name()
+        .expect("invalid path")
+        .to_str()
+        .expect("unable to convert file path to str");
+    let version = env!("CARGO_PKG_VERSION");
+    let body = read_file(path);
+    match sql_type {
+        SqlType::Migration => wrap_migration_file(filename, version, body),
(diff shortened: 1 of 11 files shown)


migration/migration-table.sql
```diff
@@ -0,0 +1,21 @@
...
+    DECLARE
+        _schema_present bool = false;
+    BEGIN
+        SELECT count(*) FILTER (WHERE s.nspname = '_ps_catalog') > 0
+        INTO STRICT _schema_present
+        FROM pg_catalog.pg_namespace s;
+        IF _schema_present THEN
+            RAISE LOG 'Schema "_ps_catalog" is already present, skipping creation';
+            RETURN;
+        END IF;
```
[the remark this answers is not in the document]
cevian: See "Creation of _ps_catalog.migration table" in https://docs.google.com/document/d/1bTuoUh-Vuz2AdBBvrd7vXaAEWUBuy9BHNp1ilUOSqEk/edit#

cevian (changes requested):
Review of only last commit. Need security fix. See comments below.

templates/idempotent-wrapper.sql
```diff
@@ -0,0 +1,9 @@
+
+-- {{filename}}
+DO
+$outer_idempotent_block$
+    BEGIN
+        {{body|indent(8)}}
+        RAISE LOG 'Applied idempotent {{filename}}';
```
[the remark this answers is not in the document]
JamesGuthrie: No, that wasn't the intention. It does get logged to the server (at least with the config in our docker image). What log level would you have expected here?

build.rs
```diff
@@ -0,0 +1,120 @@
...
+const MIGRATION_FILE_NAME: &str = "hand-written-migration.sql";
+
+/// This build script reads the SQL files placed in the `migration/{idempotent,migration}`
+/// directories, wraps each file in its respective template (in the `templates` directory), and
+/// outputs the whole contents as one long SQL file (`MIGRATION_FILE_NAME`) in the root directory.
+fn main() {
+    println!("cargo:rerun-if-changed=Cargo.lock");
+    println!("cargo:rerun-if-changed=migration");
+    println!("cargo:rerun-if-changed=templates");
+    println!("cargo:rerun-if-changed=migration/migration-table.sql");
```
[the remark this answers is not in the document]
JamesGuthrie: Yes, you're right. Looking into it I did find some other changes which needed to be made because of how this actually works. I've also documented it a little better.

migration/migration-table.sql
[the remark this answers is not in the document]
JamesGuthrie: I have reworked this quite a bit. Please take a look.

migration/migration-table.sql
```diff
@@ -0,0 +1,49 @@
...
+        WHERE pg_user.usename = current_user;
+
+        SELECT pg_namespace.nspowner
+        INTO _ps_catalog_schema_owner_id
+        FROM pg_catalog.pg_namespace
+        WHERE pg_namespace.nspname = '_ps_catalog';
+
+        IF _ps_catalog_schema_owner_id IS NOT NULL THEN
+            IF _ps_catalog_schema_owner_id != _current_user_id  THEN
+                RAISE 'Schema "_ps_catalog" is owned by the wrong user';
```
sumerman, line 21: Here and below would be helpful to state what is considered "the right user".

JamesGuthrie: I've addressed this. Although in writing the comment I realised that I'm not sure that this method is entirely correct: especially with the non-trusted path (with pg12), because in that case I assume that it's possible that there could be two separate users with superuser, so the owner doesn't match and the check fails. @cevian WDYT?

cevian: in PG12 I think its fine requiring a consistent owning user to run this every time.

migration/migration-table.sql
```diff
@@ -0,0 +1,54 @@
...
+        FROM pg_catalog.pg_namespace
+        WHERE pg_namespace.nspname = '_ps_catalog';
+
+        IF _ps_catalog_schema_owner_id IS NOT NULL THEN
+            IF _ps_catalog_schema_owner_id != _current_user_id  THEN
+                -- We require that the superuser is the schema owner, otherwise
+                -- there is a privilege escalation path available. We also know
+                -- that the current user must be superuser, otherwise they
+                -- wouldn't be able to install the extension.
+                RAISE 'Schema "_ps_catalog" is owned by the wrong user';
```
sumerman, line 25:
```suggestion
                RAISE 'Only the owner of "_ps_catalog" schema can run this migration';
```

migration/migration-table.sql
```diff
@@ -0,0 +1,54 @@
...
+        SELECT pg_class.relowner
+        INTO _migration_table_owner_id
+        FROM pg_catalog.pg_class
+          JOIN pg_catalog.pg_namespace ON pg_namespace.oid = pg_class.relnamespace
+        WHERE pg_class.relname = 'migration' AND pg_namespace.nspname = '_ps_catalog';
+
+        IF _migration_table_owner_id IS NOT NULL THEN
+            IF _migration_table_owner_id != _current_user_id THEN
+                -- We require that the superuser owns this table, see above.
+                RAISE 'Table "_ps_catalog.migration" is owned by the wrong user';
```
sumerman, line 42:
```suggestion
                RAISE 'Only the owner of "_ps_catalog.migration" table can run this migration';
```

migration/migration-table.sql
JamesGuthrie: done

migration/migration-table.sql
JamesGuthrie: done

migration/migration-table.sql
```diff
@@ -0,0 +1,54 @@
...
+                -- We require that the superuser owns this table, see above.
+                RAISE 'Only the owner of the "_ps_catalog.migration" table can install/upgrade this extension';
+                RETURN;
+            END IF;
+            RAISE LOG 'Table "_ps_catalog.migration" is already present, skipping creation';
+        ELSE
+            CREATE TABLE _ps_catalog.migration(
+                  name TEXT NOT NULL PRIMARY KEY
+                , applied_at_version TEXT
+                , applied_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT now()
```
cevian, line 50: I agree with John, clock_timestamp would be better here
```suggestion
                , applied_at TIMESTAMP WITH TIME ZONE NOT NULL DEFAULT clock_timestamp()
```

templates/idempotent-wrapper.sql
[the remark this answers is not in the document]
cevian: really I get: 
```
postgres=# show log_min_messages ;
 log_min_messages
------------------
 warning
(1 row)
```
in the ha image and LOG is above warning in `log_min_messages`

I think LOG is the right level. Would prefer it NOT to be logged in server log by default

cevian (approved):
Approving, by do agree with @jgpruitt 's comments so put suggestions below.

The security stuff looks good now

migration/migration-table.sql
JamesGuthrie: done

migration/migration-table.sql
```diff
@@ -0,0 +1,54 @@
...
+
+        IF _migration_table_owner_id IS NOT NULL THEN
+            IF _migration_table_owner_id != _current_user_id THEN
+                -- We require that the superuser owns this table, see above.
+                RAISE 'Only the owner of the "_ps_catalog.migration" table can install/upgrade this extension';
+                RETURN;
+            END IF;
+            RAISE LOG 'Table "_ps_catalog.migration" is already present, skipping creation';
+        ELSE
+            CREATE TABLE _ps_catalog.migration(
```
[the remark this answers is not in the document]
JamesGuthrie: Okay, I would prefer not to add that in this change. I have opened https://github.com/timescale/promscale_extension/issues/111 to track.

PR #105, opened 2022-03-08
Fix search path in update script

cevian:


jgpruitt (approved):

sql/promscale--0.1.2--0.1.3-beta.sql
```diff
@@ -40,11 +40,11 @@ CREATE AGGREGATE @extschema@.vector_selector(
     sample_time timestamptz,
     sample_value DOUBLE PRECISION)
 (
-    sfunc = vector_selector_transition,
+    sfunc = @extschema@.vector_selector_transition,
```
JamesGuthrie, line 43: It looks like we have the same pattern of issue in `promscale--0.1.3.beta.sql`

JamesGuthrie (changes requested):

sql/promscale--0.1.2--0.1.3-beta.sql
sumerman: What about `extension_sql!` in `lib.rs`? Do we have to fix it there as well, or does it not affect the SQL that ships?

JamesGuthrie: That doesn't affect the SQL which ships, so I think that we can skip it for now.

sql/promscale--0.3.0--0.3.1.sql
```diff
@@ -0,0 +1,49 @@
+CREATE OR REPLACE FUNCTION @extschema@.label_find_key_equal(key_to_match prom_api.label_key, pat prom_api.pattern)
```
sumerman, line 1: 1. Shouldn't we assert or set ownership of these functions? 
2. Does it matter whether a type was fully specified or not when aggregates and operators were defined? If it does, they may require a similar treatment.

sumerman (changes requested):
I believe it's missing `sql/promscale--0.3.2.sql` symlink.

cevian:
@sumerman good catch. Fixed!

sumerman (approved):

JamesGuthrie (approved):

PR #106, opened 2022-03-08
Add e2e test framework and port golden SQL tests

JamesGuthrie:
This change adds a dedicated e2e workspace which can be used to run
end-to-end tests against the promscale extension.

This change also also ports the golden SQL tests from the promscale
repository.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,6 @@
+[workspace]
+members = ["e2e"]
+
 [package]
 name = "promscale"
 version = "0.5.0"
@@ -44,3 +47,6 @@ pgx-tests = "0.3.1"
 pgx = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
 pgx-macros = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
 pgx-tests = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
+
+# from e2e workspace
+test-generator = { git = "https://github.com/JamesGuthrie/test-generator" }
--- a/e2e/build.rs
+++ b/e2e/build.rs
@@ -1,0 +1,6 @@
+extern crate build_deps;
+
+fn main() {
+    build_deps::rerun_if_changed_paths("testdata/*.sql").unwrap();
+    build_deps::rerun_if_changed_paths("testdata").unwrap();
+}
--- a/e2e/tests/common/mod.rs
+++ b/e2e/tests/common/mod.rs
@@ -1,0 +1,66 @@
+use postgres::Client;
+use std::env;
+use std::process::{Command, Stdio};
+use std::str::from_utf8;
+use testcontainers::clients::Cli;
+use testcontainers::images::generic::{GenericImage, WaitFor};
+use testcontainers::{images, Container, Docker};
+
+pub const DB: &str = "postgres-db-test";
+pub const USER: &str = "postgres-user-test";
+pub const PASSWORD: &str = "postgres-password-test";
+
+pub fn run_postgres(client: &Cli) -> Container<Cli, GenericImage> {
+    let docker_image = env::var("TS_DOCKER_IMAGE").unwrap_or(String::from(
+        "ghcr.io/timescale/dev_promscale_extension:develop-ts2-pg14",
+    ));
+
+    let src = concat!(env!("CARGO_MANIFEST_DIR"), "/testdata");
+    let dest = "/testdata";
+
+    let generic_postgres = images::generic::GenericImage::new(docker_image)
+        .with_wait_for(WaitFor::message_on_stderr(
+            "database system is ready to accept connections",
+        ))
+        .with_env_var("POSTGRES_DB", DB)
+        .with_env_var("POSTGRES_USER", USER)
+        .with_env_var("POSTGRES_PASSWORD", PASSWORD)
+        .with_volume(src, dest);
+
+    client.run(generic_postgres)
+}
+
+#[allow(dead_code)]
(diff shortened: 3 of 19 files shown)


e2e/tests/golden-tests.rs
```diff
@@ -0,0 +1,80 @@
...
+        node.get_host_port(5432).unwrap(),
+        DB
+    );
+
+    let mut client = postgres::Client::connect(connection_string, postgres::NoTls).unwrap();
+
+    let path = Path::new(resource);
+    assert!(path.exists());
+    let query = read_to_string(path).unwrap();
+    let result = client.simple_query(&query).unwrap();
```
JamesGuthrie, line 45: It would be nicer to just run `docker exec psql -c ...` and take the output of that. At the moment testcontainers doesn't support that, so I've taken the (annoyingly hacky) approach of running the file as a query against the DB and then transforming the results to a string.

The formatting of the final string could be improved to increase readability.

sumerman:
Do you intend to use this to copy golden SQL tests from the promscale repo? If so, how do you plan to keep them up to date?

JamesGuthrie:
> Do you intend to use this to copy and run golden SQL tests from the promscale repo? If so, how do you plan to keep them up to date?

Do you mean how do I intend to keep the golden sql tests here up to date with those from the promscale repo?

sumerman:
> > Do you intend to use this to copy and run golden SQL tests from the promscale repo? If so, how do you plan to keep them up to date?
> 
> Do you mean how do I intend to keep the golden sql tests here up to date with those from the promscale repo?

I guess I was trying to ask two questions at once:
- do you intend to use this to run tests from the promscale repo?
- how do I intend to keep them to date with those from the promscale repo?

e2e/tests/golden-tests.rs
cevian: Why don't we just do `docker exec psql -c ...` using process::Command::new("docker")... that's what testcontainers does anyway then 
```
let res = process::Command::new("docker")
            .arg("exec")
            .arg("-d")
            .arg(id)
            .arg("sh")
            .arg("-c")
            .arg(cmd)
            .stdout(Stdio::piped())
             .output()
             .expect("could not exec");
String::from_utf8_lossy(&output.stdout)

```

JamesGuthrie: Thanks for the input, I've done it this way.

Cargo.toml
```diff
@@ -44,3 +47,6 @@ pgx-tests = "0.3.1"
 pgx = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
 pgx-macros = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
 pgx-tests = { git = "https://github.com/timescale/pgx", branch = "promscale-staging" }
+
+# from e2e workspace
+test-generator = { git = "https://github.com/JamesGuthrie/test-generator" }
```
sumerman, line 52: Why do we need it at the top level? Can't it live in the e2e workspace?

e2e/build.rs
```diff
@@ -0,0 +1,6 @@
+extern crate build_deps;
+
+fn main() {
+    build_deps::rerun_if_changed_paths("testdata/*.sql").unwrap();
```
sumerman, line 4: Why is this necessary? To make `cargo test` re-run tests when SQL changes? Does it get cached otherwise?

e2e/tests/common/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+        PASSWORD,
+        node.get_host_port(5432).unwrap(),
+        DB
+    );
+
+    postgres::Client::connect(connection_string, postgres::NoTls).unwrap()
+}
+
+#[allow(dead_code)]
+pub fn exec_query(node: &Container<Cli, GenericImage>, query_path: &str) -> String {
```
sumerman, line 47: A minor nit: it would be more clear to me if in the context of this function it was `sql_script` instead of `query`

e2e/tests/common/mod.rs
```diff
@@ -0,0 +1,66 @@
...
+    let id = node.id();
+    let abs_query_path = "/".to_owned() + query_path;
+    let output = Command::new("docker")
```
sumerman, line 50: It relies on `docker` command being available on the system, where the test runs, doesn't it?  I think this needs to be mentioned in the readme.

sumerman (approved):

Cargo.toml
JamesGuthrie: https://doc.rust-lang.org/cargo/reference/workspaces.html#key-features

> The key points of workspaces are:
> - All packages share a common Cargo.lock file which resides in the workspace root.
> - All packages share a common [output directory](https://doc.rust-lang.org/cargo/guide/build-cache.html), which defaults to a directory named target in the workspace root.
> - **The [[patch]](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-patch-section), [[replace]](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html#the-replace-section) and [[profile.*]](https://doc.rust-lang.org/cargo/reference/profiles.html) sections in Cargo.toml are only recognized in the root manifest, and ignored in member crates' manifests.**

e2e/build.rs
JamesGuthrie: This is necessary because the test cases are autogenerated from the `testdata` directory with a macro. If you make changes to the test code in the `src`, cargo knows that it needs to recompile. But without this `build.rs`, cargo doesn't know that it needs to recompile when you make changes to the `testdata` contents.

e2e/tests/common/mod.rs
JamesGuthrie: done

PR #107, opened 2022-03-08
Extension rust <-> SQL cleanup

JamesGuthrie:
Note, this change is a follow-on to #103. I will rebase against develop when that is merged. To see only the relevant changes, review this diff: https://github.com/timescale/promscale_extension/pull/107/files/bbf283db978a8512a24175d7c8ae80f3f7767d91..8ad3d8514f37e83a55fce57b31bf9d01a98d16dd


migration/migration/006-matcher-operators.sql
```diff
@@ -128,7 +131,33 @@ AS $func$
     FROM _prom_catalog.label l
     WHERE l.key = key_to_match and l.value ~ pat
 $func$
-LANGUAGE SQL STABLE PARALLEL SAFE;
+LANGUAGE SQL STABLE PARALLEL SAFE
+SUPPORT _prom_ext.rewrite_fn_call_to_subquery;
+
+-- TODO (james): should these operators be in the _prom_catalog schema, or a "public" schema?
```
jgpruitt, line 137: schemas starting with underscores are considered internal. Since these operators are meant to be used by end-users, I think they should go in one of our metrics-specific public schemas.

jgpruitt (changes requested):

migration/migration/006-matcher-operators.sql
```diff
@@ -128,7 +131,33 @@ AS $func$
...
+    LEFTARG = prom_api.label_key,
+    RIGHTARG = prom_api.pattern,
+    FUNCTION = _prom_catalog.label_find_key_regex
+);
+
+CREATE OPERATOR _prom_catalog.!=~ (
+    LEFTARG = prom_api.label_key,
+    RIGHTARG = prom_api.pattern,
+    FUNCTION = _prom_catalog.label_find_key_not_regex
+);
```
cevian, line 160: we don't need any of these operator definitions anymore since this is now done differently here https://github.com/timescale/promscale_extension/blob/develop/migration/idempotent/002-tag-operators.sql#L34

Delete all of this please.

src/schema/estimates.rs
```diff
@@ -1,34 +0,0 @@
...
-
-// wrapper around jsonb_each_text to give a better row_estimate
-// for labels (10 not 100)
-extension_sql!(
-    r#"
-CREATE OR REPLACE FUNCTION @extschema@.label_jsonb_each_text(js jsonb, OUT key text, OUT value text)
-    RETURNS SETOF record
-    LANGUAGE internal
-    IMMUTABLE PARALLEL SAFE STRICT ROWS 10
-AS $function$jsonb_each_text$function$;
```
cevian, line 15: The function definition in [base.sql](https://github.com/timescale/promscale_extension/blob/develop/migration/idempotent/001-base.sql#L641) differs from the definition here (that's language SQL this is language internal). That was done because you could only do language internal in extensions. But, language internal is much more efficient. So, we should replace the definition in base.sql with the definition from here as part of the same commit removing this.

migration/migration/002-users.sql
```diff
@@ -64,3 +64,11 @@ CALL _prom_catalog.execute_everywhere('grant_prom_reader_prom_writer',$ee$
     END
     $$;
 $ee$);
+
+GRANT EXECUTE ON FUNCTION _prom_ext.num_cpus() TO prom_reader;
```
cevian, line 68: can we colocate these grants with the function definitions somehow?

cevian (changes requested):
A few things need changing but I think this is the right direction

src/schema/estimates.rs
JamesGuthrie: done

migration/migration/002-users.sql
JamesGuthrie: Possibly, but it would require restructuring things quite a bit. Currently, the emitted sql schema file contains (in this order):

- pgx idempotent content
- hand-written migration content
- hand-written idempotent content

This means that at the point that the function is defined (pgx idempotent content) the users for the grants are not present: they are only added in this file.

We will probably have to rethink some of this structure anyway when addressing the security concerns, so perhaps we could punt on that until then? I'll add a todo to address later.

migration/migration/006-matcher-operators.sql
JamesGuthrie: done

migration/migration/006-matcher-operators.sql
JamesGuthrie: this has been removed, see Mat's comment

migration/migration/002-users.sql
cevian: Punting on this for now is fine. Just one other idea to consider - move this to an idempotent instead of migration script

cevian (approved):
LGTM

jgpruitt (approved):

PR #118, opened 2022-03-10
Provide docker-quick-ts2-{12,13,14} target for quick iteration on SQL

JamesGuthrie:
To use, first run: `make docker-image-ts2-14`. Then you can edit the SQL
and run `make docker-quick-ts2-14`.

### DIFF
--- a/Dockerfile.quick
+++ b/Dockerfile.quick
@@ -1,0 +1,10 @@
+# This dockerfile is a helper for quick iteration on the extension SQL
+# To use, first run: `make docker-image-ts2-14`, then edit SQL and run `make docker-quick-ts2-14`
+ARG PG_VERSION_TAG
+ARG TIMESCALEDB_VERSION
+ARG EXTENSION_VERSION
+FROM timescaledev/promscale-extension:${EXTENSION_VERSION}-${TIMESCALEDB_VERSION}-${PG_VERSION_TAG}
+
+ARG EXTENSION_VERSION
+COPY sql/promscale-${EXTENSION_VERSION}.sql /usr/local/share/postgresql/extension/promscale--${EXTENSION_VERSION}.sql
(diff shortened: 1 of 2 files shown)


Dockerfile.quick
```diff
@@ -0,0 +1,11 @@
+# This dockerfile is a helper for quick iteration on the extension SQL
+# To use, first run: `make docker-image-ts2-14`, then edit SQL and run `make docker-quick-ts2-14`
+ARG PG_VERSION_TAG
+ARG TIMESCALEDB_VERSION
+ARG EXTENSION_VERSION
+FROM timescaledev/promscale-extension:${EXTENSION_VERSION}-${TIMESCALEDB_VERSION}-${PG_VERSION_TAG}
+
+ARG EXTENSION_VERSION
+COPY sql/promscale-${EXTENSION_VERSION}.sql /usr/local/share/postgresql/extension/promscale--${EXTENSION_VERSION}.sql
+COPY sql/promscale-${EXTENSION_VERSION}.sql /usr/local/share/postgresql/extension/promscale--0.0.0--${EXTENSION_VERSION}.sql
```
sumerman, line 10: Not sure I get this line.

JamesGuthrie: I've removed this line because it shouldn't appear yet (we haven't added the `promscale--0.0.0.sql` file in the develop branch yet).

The idea behind this line is that we reuse the "main" version's SQL file as the upgrade file from "all previous" versions to that "main" version. e.g. `promscale--0.5.0.sql` has the same contents as the upgrade script `promscale--0.0.0--0.5.0.sql`. This line doesn't quite handle "upgrade from all previous versions", but it does allow the upgrade path to be tested.

sumerman (approved):

PR #123, opened 2022-03-10
add compresion fixes for tracing tables

JamesGuthrie:


jgpruitt (approved):

migration/migration/008-install-uda.sql
```diff
@@ -43,6 +51,24 @@ $$
...
+AS $func$
+DECLARE
+    is_distributed BOOLEAN = false;
+BEGIN
+    IF _prom_catalog.get_timescale_major_version() >= 2 THEN
+        SELECT count(*) > 0 FROM timescaledb_information.data_nodes
+            INTO is_distributed;
+    END IF;
+    RETURN is_distributed;
+EXCEPTION WHEN SQLSTATE '42P01' THEN -- Timescale 1.x, never distributed
```
sumerman, line 65: How can we end up here if we check Timescale version first?

migration/migration/015-tracing-compression.sql
```diff
@@ -0,0 +1,99 @@
...
+    END IF;
+
+    IF _is_timescaledb_installed THEN
+        IF _is_multinode THEN
+            --need to clear the search path while creating distributed
+            --hypertables because otherwise the datanodes don't find
+            --the right column types since type names are not schema
+            --qualified if in search path.
+            _saved_search_path := current_setting('search_path');
+            SET search_path = pg_temp;
```
cevian, line 41: can we just always do this and avoid the if _is_multinode logic?

migration/migration/015-tracing-compression.sql
```diff
@@ -0,0 +1,99 @@
...
+        IF _timescaledb_major_version >= 2 THEN
+            _is_timescaledb_oss = (current_setting('timescaledb.license') = 'apache');
+        ELSE
+            _is_timescaledb_oss = (SELECT edition = 'apache' FROM timescaledb_information.license);
+        END IF;
+    END IF;
+
+    IF _is_timescaledb_installed THEN
+        IF _is_multinode THEN
+            --need to clear the search path while creating distributed
```
sumerman, line 36: I am confused: where hypertables are "created"? I can only see them being altered.

migration/migration/015-tracing-compression.sql
```diff
@@ -0,0 +1,99 @@
...
+            PERFORM public.remove_compression_policy('_ps_trace.event', if_exists=>true);
+            PERFORM public.remove_compression_policy('_ps_trace.link', if_exists=>true);
+
+            ALTER TABLE _ps_trace.span SET (timescaledb.compress=false);
+            ALTER TABLE _ps_trace.event SET (timescaledb.compress=false);
+            ALTER TABLE _ps_trace.link SET (timescaledb.compress=false);
+
+            ALTER TABLE _ps_trace.span SET (timescaledb.compress, timescaledb.compress_segmentby='trace_id', timescaledb.compress_orderby='span_id,start_time');
+            ALTER TABLE _ps_trace.event SET (timescaledb.compress, timescaledb.compress_segmentby='trace_id', timescaledb.compress_orderby='span_id,time');
+            ALTER TABLE _ps_trace.link SET (timescaledb.compress, timescaledb.compress_segmentby='trace_id', timescaledb.compress_orderby='span_id,span_start_time');
```
cevian, line 90: I think these settings should change. The trace_id should go into the orderby. Is that right @niksajakovljevic ?

migration/migration/015-tracing-compression.sql
```diff
@@ -0,0 +1,99 @@
...
+        PERFORM public.set_chunk_time_interval(
+            '_ps_trace.link'::regclass,
+            '1 hour'::interval
+        );
+
+        IF _is_multinode THEN
+            execute format('SET search_path = %s', _saved_search_path);
+        END IF;
+
+        IF (NOT _is_timescaledb_oss) AND _is_compression_available THEN
```
cevian, line 61: does it make more sense to set _is_compression_available=false if using oss above and simplify these IFs.

cevian (changes requested):

migration/migration/015-tracing-compression.sql
niksajakovljevic: Yes but we should also remove `segment_by` . But I suggest we make that a separate PR

migration/migration/015-tracing-compression.sql
jgpruitt: yes

migration/migration/015-tracing-compression.sql
jgpruitt: yes

migration/migration/015-tracing-compression.sql
jgpruitt: they were created in the 012-tracing.sql migration, but with different compression settings. this migration is needed to alter those settings

migration/migration/008-install-uda.sql
jgpruitt: i don't know. i actually copied this code, from where it was already implemented, but I agree it shouldn't be possible, especially now that we don't support 1.x anymore

migration/migration/015-tracing-compression.sql
jgpruitt: is 1 hour chunk size still too big?

migration/migration/015-tracing-compression.sql
sumerman: If they are not created here do we need to clear the search path anyway?

migration/migration/008-install-uda.sql
```diff
@@ -5,6 +5,14 @@ AS $func$
 $func$
 LANGUAGE SQL STABLE PARALLEL SAFE;
 
+CREATE OR REPLACE FUNCTION _prom_catalog.get_timescale_minor_version()
```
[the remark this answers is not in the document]
jgpruitt: they were added in the idempotent section but since idempotent runs last in the migration sequence, you cannot always rely on the functions existing early enough in the fresh install to use them on new migrations.

migration/migration/015-tracing-compression.sql
jgpruitt: @niksajakovljevic do the compression settings look good now? I think since this migration hasn't been released, we can continue to add the trace_id changes to it when ready in another PR

migration/migration/008-install-uda.sql
cevian, line 8: Is this used anywhere anymore, if not I would delete this.

cevian (approved):
I am fine with this once @niksajakovljevic approves these settings. I think its best to get the settings right here. We don't need to combine it with the uuid stuff.

Issue #113, opened 2022-03-10
Old non-versioned promscale.so file needs to be in new promscale_extension image as well. 


cevian:
The old non-versioned promscale.so file needs to exist in the new promscale_extension image (maybe). This is to enable transitions from the old to the new version to run smoothly. Without this, the docker image will be unusable until an `alter extension promscale upgrade` occurs. 

This is also gonna be required for older promscale versions once promscale versioned .so are a thing. We'll need to keep X previous versioned .so files in the docker image. Toolkit needs [to do the same ](https://github.com/timescale/timescaledb-docker-ha/blob/master/build_scripts/install_timescaledb-toolkit.sh).

We can probably build upon previous docker images in the short term.

cevian:
@JamesGuthrie the issue is for non-docker images where you'll only have .so for versions that have been previously installed so if you are going from 0.12.0 to 0.14.0 and have never installed 0.13.0 you'll be missing that .so

cevian:
I think this issue is conflating 2 things.

cevian:
@sumerman can this be closed?

sumerman:
yeah.

PR #125, opened 2022-03-11
Create version upgrade symlinks

JamesGuthrie:
When a new extension version is published, we require both a SQL schema
file for the new version (e.g. promscale--0.8.0.sql), as well as an
upgrade schema file (e.g. promscale--0.7.0--0.8.0.sql).

Because we're versioning the shared object, we actually require an
upgrade schema file for every previously-published version to the new
version (otherwise function definitions reference a .so which is not
present).

The fact that our major-version SQL schema file has an internal
migration mechanism making it effectively idempotent means that we can
reuse the major-version SQL schema file as the upgrade file to that
version.

This commit adds machinery to automatically create symlinks pointing to
the major-version SQL schema.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -50,11 +50,11 @@ USER postgres
 RUN cd ../ && cargo pgx new promscale && cd promscale
 COPY Cargo.* Makefile /build/promscale/
 RUN --mount=type=cache,uid=70,gid=70,target=/build/promscale/.cargo/registry \
-    make package
+    make dependencies
 
 # Build extension
 COPY Cargo.* /build/promscale/
-COPY promscale.control Makefile build.rs /build/promscale/
+COPY promscale.control Makefile build.rs create-upgrade-symlinks.sh /build/promscale/
 COPY .cargo/ /build/promscale/.cargo/
 COPY src/ /build/promscale/src/
 COPY sql/*.sql /build/promscale/sql/
--- a/Dockerfile.quick
+++ b/Dockerfile.quick
@@ -7,4 +7,6 @@ FROM timescaledev/promscale-extension:${EXTENSION_VERSION}-${TIMESCALEDB_VERSION
 
 ARG EXTENSION_VERSION
 COPY sql/promscale-${EXTENSION_VERSION}.sql /usr/local/share/postgresql/extension/promscale--${EXTENSION_VERSION}.sql
+# TODO (james): This probably needs to be extended to be created for all `upgradeable_from` in promscale.control
+COPY sql/promscale-${EXTENSION_VERSION}.sql /usr/local/share/postgresql/extension/promscale--0.0.0--${EXTENSION_VERSION}.sql
 
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 .DEFAULT_GOAL := help
 
 ARCH ?= $(shell uname -m)
-EXT_VERSION ?= $(shell cargo pkgid | cut -d':' -f3)
+# This little workaround allows usage of the '#' character on both gnu make 3.8 and 4.3+
+H := \#
+EXT_VERSION ?= $(shell cargo pkgid | cut -d'$H' -f2 | cut -d':' -f2)
 IMAGE_NAME ?= timescaledev/promscale-extension
 OS_NAME ?= debian
 OS_VERSION ?= 11
@@ -62,14 +64,23 @@ build: ## Build the extension
 clean: ## Clean up latest build
 	cargo clean
 
+.PHONY: dependencies
+dependencies: ## Used in docker build to improve build caching
+	# both of these steps are also run in the `package` target, so we run them here to provide better caching
+	cargo pgx schema pg${PG_BUILD_VERSION} --out sql/promscale--${EXT_VERSION}.sql
+	cargo pgx package --pg-config ${PG_CONFIG}
+	rm sql/promscale--${EXT_VERSION}.sql
+
 .PHONY: package
 package: ## Generate extension artifacts for packaging
+	cargo pgx schema pg${PG_BUILD_VERSION} --out sql/promscale--${EXT_VERSION}.sql
+	bash create-upgrade-symlinks.sh
 	cargo pgx package --pg-config ${PG_CONFIG}
 
 .PHONY: install
 install: ## Install the extension in the Postgres found via pg_config
+	bash create-upgrade-symlinks.sh
 	cargo pgx install --pg-config ${PG_CONFIG}
-
 
 .PHONY: release
 release: release-builder ## Produces release artifacts based on OS_NAME, OS_VERSION, and PG_RELEASE_VERSION
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
 .DEFAULT_GOAL := help
 
-ARCH ?= `uname -m`
-EXT_VERSION ?= `cargo pkgid | cut -d':' -f3`
+ARCH ?= $(shell uname -m)
+EXT_VERSION ?= $(shell cargo pkgid | cut -d':' -f3)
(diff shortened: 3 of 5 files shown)


create-upgrade-symlinks.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+prev_versions=$(cat promscale.control | sed -n 's/# upgradeable_from = \(.*\)/\1/p' | sed "s/[[:space:]']//g" | tr ',' '\n')
```
sumerman, line 3: Does this mean we need to manually list all the `prev_versions` in that file?

Dockerfile.quick
```diff
@@ -7,4 +7,6 @@ FROM timescaledev/promscale-extension:${EXTENSION_VERSION}-${TIMESCALEDB_VERSION
 
 ARG EXTENSION_VERSION
 COPY sql/promscale-${EXTENSION_VERSION}.sql /usr/local/share/postgresql/extension/promscale--${EXTENSION_VERSION}.sql
+# TODO (james): This probably needs to be extended to be created for all `upgradeable_from` in promscale.control
```
sumerman, line 10: Is there an issue with using `create-upgrade-symlinks.sh` here?

sumerman (changes requested):
Requesting changes is about backticks in the Makefile. The rest is optional.

create-upgrade-symlinks.sh
JamesGuthrie: yes

JamesGuthrie: Note: this might change, depending on what exactly we do here: https://github.com/timescale/promscale_extension/issues/113

Makefile
```diff
@@ -1,14 +1,16 @@
 .DEFAULT_GOAL := help
 
 ARCH ?= `uname -m`
```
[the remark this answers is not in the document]
JamesGuthrie: I agree.

Dockerfile.quick
JamesGuthrie: Ya, it wasn't designed to be used here and there are enough differences between where it _was_ designed to work and here that I'm not sure if it makes sense to make it work here. As I mention below, I suspect that some of this stuff will change in the near future, so I'd prefer to punt on perfecting that for now.

Makefile
```diff
@@ -60,15 +62,24 @@ build: ## Build the extension
 clean: ## Clean up latest build
 	cargo clean
 
+.PHONY: prepackage
+prepackage: ## Used internally to run packaging
```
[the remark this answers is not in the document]
JamesGuthrie: done

Makefile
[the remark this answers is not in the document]
JamesGuthrie: I have done this. It took a little bit more work, because I realised that something had been broken a while back, and I needed to fix that. I have broken out those changes into two separate commits in this PR.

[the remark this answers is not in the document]
sumerman: I've looked at the change you are talking about and I can't say I follow it. I agree that the variables are used for two distinct purposes. Although this statement:

>Context 1. expects to look at the local environment at build time with
pg_config to determine what PG_VERSION is. It does not expect PG_VERSION
to be passed in from the outside.

was not accurate before `53a992f27f41e666331bbbd7fc87fbcd54085680`.

Furthermore, both sets of variables are altered when building artefacts: 
- `PG_BUILD_VERSION` and `PG_BUILD_VER` are used when building images. Although instead of passing them from the outside we set them in the Makefile itself.
- `PG_RELEASE_VERSION` and `PG_RELEASE_VER` are used for the packages and we pass them from the outside.

I suggest either better separate uses of build vs release variables (e.g. use build variables only with `cargo` and use release variables when building artefacts) or keep only one set of variables and make all of them overridable.

cevian (changes requested):
The symlink creation script looks good, but I am very confused by the docker caching/prepackage stuff

Dockerfile
```diff
@@ -48,9 +48,9 @@ USER postgres
 
 # Pre-build extension dependencies
 RUN cd ../ && cargo pgx new promscale && cd promscale
-COPY Cargo.* Makefile /build/promscale/
+COPY Cargo.* Makefile create-upgrade-symlinks.sh /build/promscale/
```
cevian, line 51: Why do we need to copy over or use the create-upgrade-symlinks.sh? Isn't this just about caching Rust dependencies? But create-upgrade-symlinks.sh is about the SQL not the Rust stuff and the SQL isn't even copied over from the container at this point. So I am confused.

JamesGuthrie: Ya you're right. It doesn't need to be copied over at this point yet. I've moved it to `package` only.

JamesGuthrie: I'm not sure if that answered your question adequately though, please follow up if something's still unclear...

Makefile
```diff
@@ -125,49 +137,49 @@ release-test: release-tester ## Test the currently selected release package
...
-		-t $(IMAGE_NAME):latest-ts$(TIMESCALEDB_MAJOR)-$(PG_VER) \
+		--build-arg PG_VERSION_TAG=$(PG_BUILD_VER) \
+		-t $(IMAGE_NAME):$(EXT_VERSION)-$(TIMESCALEDB_VER)-$(PG_BUILD_VER) \
+		-t $(IMAGE_NAME):$(EXT_VERSION)-ts$(TIMESCALEDB_MAJOR)-$(PG_BUILD_VER) \
+		-t $(IMAGE_NAME):latest-ts$(TIMESCALEDB_MAJOR)-$(PG_BUILD_VER) \
 		.
 
 .PHONY: docker-image-12
-docker-image-12: PG_VER=pg12
+docker-image-12: PG_BUILD_VER=pg12
```
sumerman, line 147: I find it somewhat confusing that `make release` is a single target and relies on passing `OS_NAME`, `OS_VERSION`, and `PG_RELEASE_VERSION` variables, yet `make docker-image` is a group of targets and they manipulate `PG_BUILD_VER` internally.

sumerman (approved):

Makefile
```diff
@@ -42,41 +46,49 @@ ifeq ($(OS_NAME),fedora)
...
+	cargo build --release --features pg${PG_BUILD_VERSION} $(EXTRA_RUST_ARGS)
+	cargo pgx schema -f pg${PG_BUILD_VERSION} --release
 
 .PHONY: clean
 clean: ## Clean up latest build
 	cargo clean
 
+.PHONY: prepackage
+prepackage: ## Used in docker build to improve the build caching
+	cargo pgx schema pg${PG_BUILD_VERSION} --out sql/promscale--${EXT_VERSION}.sql
```
cevian, line 69: does this line help cache any dependencies? can we get rid of line 69 and 71 and just keep `cargo pgx package`?

Makefile
```diff
@@ -42,41 +46,49 @@ ifeq ($(OS_NAME),fedora)
...
-	cargo pgx schema -f pg${PG_VERSION} --release
+	cargo build --release --features pg${PG_BUILD_VERSION} $(EXTRA_RUST_ARGS)
+	cargo pgx schema -f pg${PG_BUILD_VERSION} --release
 
 .PHONY: clean
 clean: ## Clean up latest build
 	cargo clean
 
+.PHONY: prepackage
+prepackage: ## Used in docker build to improve the build caching
```
cevian, line 68: optional nit
```suggestion
dependencies: ## Used in docker build to improve the build caching
```

cevian (approved):

Makefile
JamesGuthrie: It does help to cache dependencies, I've left it as-is but added a comment to clarify.

Makefile
JamesGuthrie: Thanks for the suggestion! done.

Issue #128, opened 2022-03-11
Unable to index large tag values


cevian:
This is a takeover of as it's now an extension issue https://github.com/timescale/promscale/issues/1091

PR #130, opened 2022-03-11
Fix for uniqueness constraint on large tag values.

sumerman:
This addresses #128 by replacing `value` attribute within the UNIQUE constraint with `sha512` hash of its textual representation.

I ran the following SQL as a microbenchmark:

```sql
SELECT put_tag('service.name', n::text::jsonb, resource_tag_type())
FROM generate_series(1, 20000) AS gs (n)
```

and

```sql
SELECT get_tag_map(('{"service.name": ' || n::text || '}')::jsonb)
FROM generate_series(1, 20000) AS gs (n)
```

~There doesn't seem to be a significant performance difference on the insertion path. Unmodified get_tag_map became roughly twice as slow, so I had to alter it as well. This hints at possible performance degradation elsewhere.~ 

I think I was able to chase down all functions that relied on an index covering `(key, value)`  and augmented these places with an additional check `jsonb_digest(a.value) = ...` to make use of the new index. Unfortunately, some of these functions might become marginally slower, as they'd need to access both the index (for `key, jsonb_digest(value)`) and heap (to fetch `value` for hits). In the end performance hit appears to be ~10%.

### DIFF
--- a/migration/idempotent/007-tracing-tags.sql
+++ b/migration/idempotent/007-tracing-tags.sql
@@ -186,6 +186,7 @@ AS $func$
     SELECT jsonb_build_object(a.key_id, a.id)
     FROM _ps_trace.tag a
     WHERE a.key = _op.tag_key
+    AND _prom_ext.jsonb_digest(a.value) = _prom_ext.jsonb_digest(_op.value)
     AND a.value = _op.value
     LIMIT 1
 $func$
--- a/migration/idempotent/008-tracing-functions.sql
+++ b/migration/idempotent/008-tracing-functions.sql
@@ -339,10 +339,14 @@ RETURNS BIGINT
 AS $func$
 DECLARE
     _tag _ps_trace.tag;
+    _value_digest bytea;
 BEGIN
+    SELECT _prom_ext.jsonb_digest(_value) INTO _value_digest;
+
     SELECT * INTO _tag
     FROM _ps_trace.tag
     WHERE key = _key
+    AND _prom_ext.jsonb_digest(value) = _value_digest
     AND value = _value
     FOR UPDATE;
 
@@ -355,19 +359,24 @@ BEGIN
             _value
         FROM _ps_trace.tag_key k
         WHERE k.key = _key
-        ON CONFLICT (key, value) DO
+        ON CONFLICT (key, _prom_ext.jsonb_digest(value)) DO
         UPDATE SET tag_type = t.tag_type | EXCLUDED.tag_type
         WHERE t.tag_type & EXCLUDED.tag_type = 0;
 
         SELECT * INTO STRICT _tag
         FROM _ps_trace.tag
         WHERE key = _key
+        AND _prom_ext.jsonb_digest(value) = _value_digest
         AND value = _value;
     ELSIF _tag.tag_type & _tag_type = 0 THEN
         UPDATE _ps_trace.tag as t
         SET tag_type = t.tag_type | _tag_type
         WHERE t.key = _key -- partition elimination
         AND t.id = _tag.id;
+    END IF;
+
+    IF _tag.value != _value THEN
+        RAISE EXCEPTION 'put_tag failed. Distinct values % and % for key % have identical sha512 digest.', _tag.value, _value, _key;
     END IF;
 
     RETURN _tag.id;
@@ -386,6 +395,7 @@ AS $func$
         SELECT a.key_id, a.id
         FROM _ps_trace.tag a
         WHERE x.key = a.key
+        AND _prom_ext.jsonb_digest(x.value) = _prom_ext.jsonb_digest(a.value)
         AND x.value = a.value
         LIMIT 1
     ) a on (true)
@@ -399,12 +409,18 @@ AS $func$
 DECLARE
     _service_name_id bigint;
     _operation_id bigint;
+    _service_name_json jsonb;
+    _service_name_digest bytea;
 BEGIN
+    SELECT to_jsonb(_service_name::text) INTO _service_name_json;
+    SELECT _prom_ext.jsonb_digest(_service_name_json) INTO _service_name_digest;
+
     SELECT id INTO _service_name_id
     FROM _ps_trace.tag
     WHERE key = 'service.name'
     AND key_id = 1
-    AND value = to_jsonb(_service_name::text)
+    AND _prom_ext.jsonb_digest(value) = _service_name_digest
+    AND value = _service_name_json
     ;
 
     IF NOT FOUND THEN
@@ -414,7 +430,7 @@ BEGIN
             ps_trace.resource_tag_type(),
             'service.name',
             1,
-            to_jsonb(_service_name::text)
+            _service_name_json
         )
         ON CONFLICT DO NOTHING
         RETURNING id INTO _service_name_id;
@@ -424,7 +440,8 @@ BEGIN
             FROM _ps_trace.tag
             WHERE key = 'service.name'
             AND key_id = 1
-            AND value = to_jsonb(_service_name::text);
+            AND _prom_ext.jsonb_digest(value) = _service_name_digest
+            AND value = _service_name_json;
         END IF;
     END IF;
 
--- a/migration/migration/015-tracing-large-tags-fix.sql
+++ b/migration/migration/015-tracing-large-tags-fix.sql
@@ -1,0 +1,2 @@
+ALTER TABLE _ps_trace.tag DROP CONSTRAINT tag_key_value_id_key_id_key;
+CREATE UNIQUE INDEX tag_key_value_id_key_id_key ON _ps_trace.tag (key, _prom_ext.jsonb_digest(value)) INCLUDE (id, key_id);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -13,6 +13,132 @@ mod util;
 
 pg_module_magic!();
 
+/// TODO: presently, we deliberate on how to structure SQL-only tests.
+/// As soon as we agree on something these tests should be moved over.
+
+#[cfg(any(test, feature = "pg_test"))]
+#[pg_schema]
+mod tests {
+    use pgx::*;
+    use serde_json::json;
+    use std::collections::HashSet;
+    use std::iter::FromIterator;
+
+    #[pg_test]
+    fn test_very_large_values_in_put_tag() {
+        let tag_id = Spi::get_one::<i64>(
+            r#"
+            SELECT put_tag('service.namespace', gen.kilobytes_of_garbage::jsonb, resource_tag_type())
+			FROM (
+				SELECT string_agg(n::text, '') AS kilobytes_of_garbage
+				FROM generate_series(1, 2000) AS gs (n)
+			) AS gen
+            "#,
+        )
+        .expect("SQL query failed");
+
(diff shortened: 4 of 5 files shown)


cevian:
Meant to start a discussion: I think we should use sha512 (512-bit) not md5 (128-bit). Hash collisions are potentially really bad here, so I'd like to make that as least-likely as possible.

migration/idempotent/008-tracing-functions.sql
```diff
@@ -355,7 +355,7 @@ BEGIN
             _value
         FROM _ps_trace.tag_key k
         WHERE k.key = _key
-        ON CONFLICT (key, value) DO
+        ON CONFLICT (key, pg_catalog.md5(value::TEXT)) DO
```
cevian, line 358: I think at the end of this function we should check if there was a collision (IF _tag.value != _value) and throw an error if that's the case. That will verify that in even in the highly-unlikely case of a hash collision, we don't get data corruption. Instead we get an error.

migration/idempotent/008-tracing-functions.sql
```diff
@@ -355,7 +355,7 @@ BEGIN
             _value
```
cevian, line 355: line 343-347 and 362-367 should be changed to include where clauses on the hash `pg_catalog.md5(value::TEXT) = pg_catalog.md5(_value::TEXT)` so that the new unique index `tag_key_value_id_key_id_key` is fully used, right (or is Postgres smart enough to figure this out)

cevian:
This addresses #128

migration/idempotent/008-tracing-functions.sql
sumerman: Yeah, as the description states: it's not smart enough and there might be hidden performance problems. The ones you pointed out here, equality operators, possibly more. 

First I want to run e2e and make sure nothing is broken, then I'll do an optimization sweep.

migration/idempotent/008-tracing-functions.sql
sumerman: Good call. Do you think it's enough to guard against collisions during ingestion only? Or should I apply similar logic on reads? 

E.g.:
```
AND pg_catalog.md5(x.value::TEXT) = pg_catalog.md5(a.value::TEXT)
AND x.value = a.value
```

A potential downside is a heap lookup even in queries where `value` isn't used for anything else.

migration/migration/012-tracing.sql
```diff
@@ -58,10 +58,10 @@ CREATE TABLE _ps_trace.tag
     tag_type ps_trace.tag_type NOT NULL,
     key_id bigint NOT NULL,
     key ps_trace.tag_k NOT NULL REFERENCES _ps_trace.tag_key (key) ON DELETE CASCADE,
-    value ps_trace.tag_v NOT NULL,
-    UNIQUE (key, value) INCLUDE (id, key_id)
+    value ps_trace.tag_v NOT NULL
 )
 PARTITION BY HASH (key);
+CREATE UNIQUE INDEX tag_key_value_id_key_id_key ON _ps_trace.tag (key, pg_catalog.sha512(value::TEXT::BYTEA)) INCLUDE (id, key_id);
```
jgpruitt, line 64: I believe we can cast jsonb straight to bytea and forgo the cast to text first, right?

migration/idempotent/008-tracing-functions.sql
```diff
@@ -343,6 +343,7 @@ BEGIN
     SELECT * INTO _tag
     FROM _ps_trace.tag
     WHERE key = _key
+    AND pg_catalog.sha512(value::TEXT::BYTEA) = pg_catalog.sha512(_value::TEXT::BYTEA)
```
jgpruitt, line 346: We have to put `pg_catalog.sha512(_value::TEXT::BYTEA)` on the right hand side in three places to get the index scan, right? That's why we cannot calc the sha once into a variable and reuse it?

migration/idempotent/008-tracing-functions.sql
```diff
@@ -386,6 +392,7 @@ AS $func$
         SELECT a.key_id, a.id
         FROM _ps_trace.tag a
         WHERE x.key = a.key
+        AND pg_catalog.sha512(x.value::TEXT::BYTEA) = pg_catalog.sha512(a.value::TEXT::BYTEA)
         AND x.value = a.value
```
jgpruitt, line 396: I think we should remove `AND x.value = a.value` so we can avoid the heap lookup

migration/idempotent/008-tracing-functions.sql
```diff
@@ -424,6 +432,7 @@ BEGIN
             FROM _ps_trace.tag
             WHERE key = 'service.name'
             AND key_id = 1
+            AND pg_catalog.sha512(value::TEXT::BYTEA) = pg_catalog.sha512(_service_name::TEXT::BYTEA)
             AND value = to_jsonb(_service_name::text);
```
jgpruitt, line 436: remove `AND value = to_jsonb(_service_name::text)` ?

migration/idempotent/007-tracing-tags.sql
```diff
@@ -186,6 +186,7 @@ AS $func$
     SELECT jsonb_build_object(a.key_id, a.id)
     FROM _ps_trace.tag a
     WHERE a.key = _op.tag_key
+    AND pg_catalog.sha512(a.value::TEXT::BYTEA) = pg_catalog.sha512(_op.value::TEXT::BYTEA)
     AND a.value = _op.value
```
jgpruitt, line 190: remove `AND a.value = _op.value`?

migration/migration/012-tracing.sql
```diff
@@ -58,10 +58,10 @@ CREATE TABLE _ps_trace.tag
     tag_type ps_trace.tag_type NOT NULL,
     key_id bigint NOT NULL,
     key ps_trace.tag_k NOT NULL REFERENCES _ps_trace.tag_key (key) ON DELETE CASCADE,
-    value ps_trace.tag_v NOT NULL,
-    UNIQUE (key, value) INCLUDE (id, key_id)
+    value ps_trace.tag_v NOT NULL
```
jgpruitt, line 61: As I understand it, we cannot modify this file `012-tracing.sql`, rather we have to add a new migration file which would assume the original version of the table exists and issue ALTERs etc. to make this change.

jgpruitt (changes requested):

migration/migration/012-tracing.sql
sumerman: Perhaps I'm missing something, but I get `ERROR: cannot cast type jsonb to bytea` when I run `select sha512('{"test": 1}'::jsonb::bytea);`

migration/idempotent/008-tracing-functions.sql
sumerman: This and other redundant comparisons are safeguards against hash collisions. AFAIK heap lookups would happen for index hits only, so I'd rather stay on the "correct but slightly slower" side.

migration/migration/012-tracing.sql
sumerman: But that's actually a very good point: I'm not so sure a textual representation of JSONB is deterministic across major PG versions. Maybe it's worth defining a custom conversion.

sumerman: Turns out `jsonb_send` and `::text` are practically the same https://doxygen.postgresql.org/jsonb_8c_source.html#l00148 And AFAICT there hasn't been a commitment to keep behaviour across pg versions.

sumerman:
E2E passes except for `TestSQLGoldenFiles`

cevian:
E2E tests were previously broken on master. Now fixed. Running tests here: https://github.com/timescale/promscale/actions/runs/2041259807

migration/migration/015-tracing-large-tags-fix.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE _ps_trace.tag DROP CONSTRAINT tag_key_value_id_key_id_key;
+CREATE UNIQUE INDEX tag_key_value_id_key_id_key ON _ps_trace.tag (key, _prom_ext.jsonb_digest(value)) INCLUDE (id, key_id);
```
cevian, line 2: can this not have value in the INCLUDE either if the value is too large?

jgpruitt: unfortunately, the value cannot be in the INCLUDE either:

```
create table foo (bar text, baz int);
create index on foo using btree (baz) include (bar);
insert into foo (bar, baz) select string_agg(x::text, ''), 42 from generate_series(1, 10000) x;
ERROR:  54000: index row requires 38912 bytes, maximum size is 8191
LOCATION:  index_form_tuple, indextuple.c:191
```

src/lib.rs
```diff
@@ -13,6 +13,74 @@ mod util;
...
 pg_module_magic!();
 
+/// TODO: presently, we deliberate on how to structure SQL-only tests.
+/// As soon as we agree on something these tests should be moved over.
+
+#[cfg(any(test, feature = "pg_test"))]
+#[pg_schema]
+mod tests {
+    use pgx::*;
+
```
jgpruitt, line 23: can we add a test using values of all the supported types to make sure the hash works for each

1. string
2. int
3. float
4. bool
5. etc...

src/lib.rs
```diff
@@ -13,6 +13,74 @@ mod util;
 
 pg_module_magic!();
 
+/// TODO: presently, we deliberate on how to structure SQL-only tests.
+/// As soon as we agree on something these tests should be moved over.
+
```
jgpruitt, line 18: can we add a test that inserts the same value with two different keys, and a test with two values for the same key?

jgpruitt (changes requested):

src/lib.rs
sumerman: Could you have a look at the tests I've implemented for the `jsonb_digest` itself and see if they suffice or you'd like me to add something? #149 includes a bunch of hand-coded cases as well as a property-based (randomized) test that checks 1000 cases on each run.

https://github.com/timescale/promscale_extension/pull/149/files#diff-07dec81586875043695312db239d17a9ea2ee59473fb436fe0234e4a4339df2dR72

src/lib.rs
sumerman: done

migration/migration/015-tracing-large-tags-fix.sql
sumerman: Yeah, keeping the value in the INCLUDE is not an option. As discussed on slack we have two options if we really want to claw back performance:

- Use a pair of partial indexes: one with `AND length(value::text) < 10000` the other with `AND length(value::text) >= 10000`, include `value` in the first one and make sure to transform every query into a union of the two. (@pgnickb suggestion)
-  Use a custom data type that can store the actual value if it's small enough instead or along with a hash. (@jgpruitt suggestion)

jgpruitt: it doesn't have to be a custom data type, but the concept is what's important

src/lib.rs
jgpruitt: lgtm

jgpruitt (approved):

PR #133, opened 2022-03-17
0.0.0 script to transition an existing installation into the new way

jgpruitt:
This is the 0.0.0 version script, a.k.a. the "takeover" script. When upgrading an existing installation from pre-0.5.0 to 0.5.0 and above, this script is run as a part of the process. The extension is dropped (if exists) and then version 0.0.0 of the extension is installed, which runs this script. Next, the extension is updated to 0.5.0 or higher.

The script adds existing database objects to the extension, changes privileges and ownership, and makes certain tables "config" tables such that they get pg_dumped.

Another PR will handle the "fresh install" path. Unfortunately, I don't expect the upgrade tests to pass with the two paths split like that. Furthermore, many of the e2e tests may also fail pending the idempotent changes which I assume will come with the "fresh install" PR. If this isn't the case, we may need to reevaluate.

### DIFF
--- a/sql/promscale--0.0.0.sql
+++ b/sql/promscale--0.0.0.sql
@@ -1,0 +1,460 @@
+DROP TABLE public.prom_schema_migrations;
+
+DO $block$
+DECLARE
+    _rec record;
+BEGIN
+    FOR _rec IN
+    (
+        select *
+        from (
+        values
+          ('SCHEMA', '_prom_catalog')
+        , ('SCHEMA', '_prom_ext')
+        , ('SCHEMA', '_ps_catalog')
+        , ('SCHEMA', '_ps_trace')
+        , ('SCHEMA', 'prom_api')
+        , ('SCHEMA', 'prom_data')
+        , ('SCHEMA', 'prom_data_exemplar')
+        , ('SCHEMA', 'prom_data_series')
+        , ('SCHEMA', 'prom_info')
+        , ('SCHEMA', 'prom_metric')
+        , ('SCHEMA', 'prom_series')
+        , ('SCHEMA', 'ps_tag')
+        , ('SCHEMA', 'ps_trace')
+        , ('TABLE', '_prom_catalog.default')
+        , ('TABLE', '_prom_catalog.exemplar')
+        , ('TABLE', '_prom_catalog.exemplar_label_key_position')
+        , ('TABLE', '_prom_catalog.ha_leases')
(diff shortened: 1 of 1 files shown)


sql/promscale--0.0.0.sql
```diff
@@ -0,0 +1,460 @@
...
+        , ('OPERATOR', 'ps_trace.?(ps_trace.tag_map, ps_tag.tag_op_regexp_matches)')
+        , ('OPERATOR', 'ps_trace.?(ps_trace.tag_map, ps_tag.tag_op_jsonb_path_exists)')
+        ) x(objtype, objname)
+    )
+    LOOP
+        -- extension is installed into _prom_ext schema. thus cannot add it to the extension
+        IF NOT (_rec.objtype = 'SCHEMA' AND _rec.objname = '_prom_ext') THEN
+            EXECUTE format('ALTER EXTENSION promscale ADD %s %s', _rec.objtype, _rec.objname);
+        END IF;
+        EXECUTE format('ALTER %s %s OWNER TO %I', _rec.objtype, _rec.objname, '@extowner@');
```
cevian, line 329: shouldn't this be to superuser not extowner? I also wonder why this wasn't caught by the tests.

sql/promscale--0.0.0.sql
```diff
@@ -0,0 +1,460 @@
...
+            END AS typ,
+            n.nspname,
+            k.relname
+        FROM pg_class k
+        INNER JOIN pg_namespace n ON (k.relnamespace = n.oid)
+        WHERE k.relkind in ('r', 'v')
+        AND n.nspname in ('prom_metric', 'prom_series', 'prom_data_series', 'prom_data')
+    )
+    LOOP
+        EXECUTE format($sql$ALTER %s %I.%I OWNER TO prom_modifier$sql$, _rec.typ, _rec.nspname, _rec.relname);
```
cevian, line 375: this will need to change to the superuser as in the other PR.

sql/promscale--0.0.0.sql
```diff
@@ -0,0 +1,460 @@
+
```
cevian, line 1: don't we need to drop the old schema migration table here?

cevian (changes requested):
A few issues. Biggest one is why are the object ownership changed to extowner and not superuser?

cevian (approved):

sumerman (approved):

PR #147, opened 2022-03-21
WIP: Remove docker image on PR close

JamesGuthrie:


sumerman (approved):

JamesGuthrie:
The remove step seems to be failing. Not quite sure why, but I suspect it's a permissions problem. Will probably have to debug with @cevian when he's onling.

jgpruitt (approved):

PR #148, opened 2022-03-21
Store migration body in migration table

JamesGuthrie:
We emit a warning when detecting that the checksum has changed.

### DIFF
--- a/templates/migration-wrapper.sql
+++ b/templates/migration-wrapper.sql
@@ -3,14 +3,20 @@
 DO
 $outer_migration_block$
 DECLARE
-    _already_applied bool = false;
-    _migration       _ps_catalog.migration = row ('{{filename}}', '{{version}}');
+    _migration_name TEXT = NULL;
+    _body_differs BOOL = false;
+    _migration _ps_catalog.migration = row ('{{filename}}', '{{version}}');
+    _body TEXT = $migrationbody${{body}}$migrationbody$;
 BEGIN
-    SELECT count(*) FILTER (WHERE name = _migration.name) > 0
-    INTO STRICT _already_applied
-    FROM _ps_catalog.migration;
-    IF _already_applied THEN
+    SELECT migration.name, migration.body <> _body
+    INTO _migration_name, _body_differs
+    FROM _ps_catalog.migration
+    WHERE name = _migration.name;
+    IF _migration_name IS NOT NULL THEN
         RAISE LOG 'Migration "{{filename}}" already applied, skipping';
+        IF _body_differs THEN
+            RAISE WARNING 'Checksum of migration "{{filename}}" has changed';
+        END IF;
         RETURN;
     END IF;
 
@@ -20,7 +26,7 @@ BEGIN
 END;
 $inner_migration_block$;
 
-    INSERT INTO _ps_catalog.migration (name, applied_at_version) VALUES (_migration.name, _migration.applied_at_version);
+    INSERT INTO _ps_catalog.migration (name, applied_at_version, body) VALUES (_migration.name, _migration.applied_at_version, _body);
     RAISE LOG 'Applied migration {{filename}}';
 END;
 $outer_migration_block$;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,9 +395,9 @@ dependencies = [
 
 [[package]]
 name = "crc32fast"
-version = "1.3.1"
+version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a2209c310e29876f7f0b2721e7e26b84aff178aa3da5d091f9bfbf47669e60e3"
+checksum = "b540bd8bc810d3885c6ea91e2018302f68baba2129ab3e88f32389ee9370880d"
 dependencies = [
  "cfg-if",
 ]
--- a/build.rs
+++ b/build.rs
@@ -11,14 +11,14 @@ use std::{env, fs};
 struct MigrationWrapper<'a> {
     filename: &'a str,
     version: &'a str,
-    body: String,
+    body: &'a str,
 }
 
 #[derive(Template)]
 #[template(path = "idempotent-wrapper.sql", escape = "none")]
 struct IdempotentWrapper<'a> {
     filename: &'a str,
-    body: String,
+    body: &'a str,
 }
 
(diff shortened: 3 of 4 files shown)


JamesGuthrie:
Implementation for #111.

cevian:
Why store checksum and not full body? What's the reasoning?

templates/migration-wrapper.sql
```diff
@@ -3,14 +3,19 @@
...
-    FROM _ps_catalog.migration;
-    IF _already_applied THEN
+    SELECT migration.name, migration.checksum <> _migration.checksum
+    INTO _migration_name, _checksum_differs
+    FROM _ps_catalog.migration
+    WHERE name = _migration.name;
+    IF _migration_name IS NOT NULL THEN
         RAISE LOG 'Migration "{{filename}}" already applied, skipping';
+        IF _checksum_differs THEN
+            RAISE WARNING 'Checksum of migration "{{filename}}" has changed';
```
JamesGuthrie, line 17: Open question: should we just warn, or fail when discovering a checksum difference?

For our development it would make sense to fail hard here, because changing migration file is in general undesired. But if we would hard fail here then we also need to have a mechanism to purposefully override/rewrite the migration body.

JamesGuthrie:
> Why store checksum and not full body? What's the reasoning?

I wondered if we need the full body? What is the use-case for having it?

cevian:
> I wondered if we need the full body? What is the use-case for having it?

cevian:
> I wondered if we need the full body? What is the use-case for having it?

Debugging. If a checksum isn't what we expect we may have a very poor understanding of what was actually run.
Also if we have the body we could compute a diff to see what changed.

It just seems like the body gives us strictly more context if things go wrong.

templates/migration-wrapper.sql
```diff
@@ -3,14 +3,20 @@
 DO
 $outer_migration_block$
 DECLARE
-    _already_applied bool = false;
-    _migration       _ps_catalog.migration = row ('{{filename}}', '{{version}}');
+    _migration_name TEXT = false;
```
sumerman, line 6: `TEXT = false` 🤔

jgpruitt (approved):

templates/migration-wrapper.sql
```diff
@@ -3,14 +3,20 @@
...
+    _body_differs BOOL = false;
+    _migration _ps_catalog.migration = row ('{{filename}}', '{{version}}');
+    _body             TEXT = $body${{body}}$body$;
```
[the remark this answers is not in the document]
JamesGuthrie: yes, I believe so:

```
promscale=# SELECT * FROM _ps_catalog.migration WHERE name = '011-tables-exemplar.sql';
          name           | applied_at_version |                                                  body                                                   |          applied_at
-------------------------+--------------------+---------------------------------------------------------------------------------------------------------+-------------------------------
 011-tables-exemplar.sql | 0.5.0              | CREATE TABLE IF NOT EXISTS _prom_catalog.exemplar_label_key_position (                                 +| 2022-03-22 16:09:02.452129+01
                         |                    |     metric_name TEXT NOT NULL,                                                                         +|
                         |                    |     key         TEXT NOT NULL,                                                                         +|
                         |                    |     pos         INTEGER NOT NULL,                                                                      +|
                         |                    |     PRIMARY KEY (metric_name, key) INCLUDE (pos)                                                       +|
                         |                    | );                                                                                                     +|
                         |                    | GRANT SELECT ON TABLE _prom_catalog.exemplar_label_key_position TO prom_reader;                        +|
                         |                    | GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _prom_catalog.exemplar_label_key_position TO prom_writer;+|
                         |                    |                                                                                                        +|
                         |                    | CREATE TABLE IF NOT EXISTS _prom_catalog.exemplar (                                                    +|
                         |                    |     id          SERIAL PRIMARY KEY,                                                                    +|
                         |                    |     metric_name TEXT NOT NULL,                                                                         +|
                         |                    |     table_name  TEXT NOT NULL,                                                                         +|
                         |                    |     UNIQUE (metric_name) INCLUDE (table_name, id)                                                      +|
                         |                    | );                                                                                                     +|
                         |                    | GRANT SELECT ON TABLE _prom_catalog.exemplar TO prom_reader;                                           +|
                         |                    | GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _prom_catalog.exemplar TO prom_writer;                   +|
                         |                    |                                                                                                        +|
                         |                    | GRANT USAGE, SELECT ON SEQUENCE _prom_catalog.exemplar_id_seq TO prom_writer;                          +|
                         |                    |                                                                                                         |
(1 row)
```

templates/migration-wrapper.sql
JamesGuthrie: Oh man, that was a 🤦

templates/migration-wrapper.sql
jgpruitt: What do we tell customers to do in the case it does error out?

JamesGuthrie: hmmm... good point. But how do we prevent ourselves from accidentally updating old migrations?

jgpruitt: Thinking about it more, I tend to agree with you @JamesGuthrie , but I think we need a plan for the case where it happens to a customer

JamesGuthrie: @cevian WDYT?

templates/migration-wrapper.sql
```diff
@@ -3,14 +3,20 @@
 DO
 $outer_migration_block$
 DECLARE
-    _already_applied bool = false;
-    _migration       _ps_catalog.migration = row ('{{filename}}', '{{version}}');
+    _migration_name TEXT = NULL;
+    _body_differs BOOL = false;
+    _migration _ps_catalog.migration = row ('{{filename}}', '{{version}}');
+    _body TEXT = $body${{body}}$body$;
```
cevian, line 9:
```suggestion
    _body TEXT = $migrationbody${{body}}$migrationbody$;
```

templates/migration-wrapper.sql
cevian: I think a warning is good enough here. There /will be/ cases when we need to change migration scripts for future migrations. Our upgrade tests will catch any errors here but let's just warn for now.

cevian (approved):

PR #153, opened 2022-03-21
Remove SECURITY DEFINER from _prom_catalog.execute_everywhere

jgpruitt:
The commands stored in the remote_commands table are executed by the extension/superuser. Making execute_everywhere security definer is a vulnerability. Remove the security definer.

Also, adjusted the "quick" targets in the Makefile so that they tag the image with all the same tags the "non-quick" targets do.


JamesGuthrie (commented):
Are there no other changes necessary to get this working? I'm slightly surprised.

cevian (approved):
I am happy with this change if the tests still pass. Although yes, I am also surprised if that's the case.

PR #157, opened 2022-03-25
Revoke execute on delete_all_traces from prom_writer

jgpruitt:


cevian (approved):
LGTM but how was this previously missed in upgrade tests?

PR #162, opened 2022-03-30
Adjust autovacuum settings

cevian:
On uncompressed tables
- set the threshold and scale_factors higher to make
  less aggressive

On compressed tables
- make freeze run on every vacuum
- make sure vacuum runs quickly after creation by decreasing
  threshold and scale_factor

Fix settings on PG13 and 14 by using the `_insert_` version of settings.

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -369,8 +369,28 @@ BEGIN
         RETURN NEW;
     END IF;
 
-    EXECUTE format('CREATE TABLE %I.%I(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL) WITH (autovacuum_vacuum_threshold = 50000, autovacuum_analyze_threshold = 50000)',
-                    NEW.table_schema, NEW.table_name);
+    --autovacuum notes: we need to vacuum here for the visibility map optimization
+    --but, we don't want to freeze anything since these tables will be truncated upon
+    --compression anyway (so keep freeze settings to the default).
+    --Make the vacuum less aggressive here by upping the threshold and scale factors.
+    IF current_setting('server_version_num')::integer >= 130000 THEN
+        EXECUTE format('CREATE TABLE %I.%I(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL) WITH
+                        (
+                            autovacuum_vacuum_insert_threshold=50000,
+                            autovacuum_vacuum_insert_scale_factor=2.0,
+                            autovacuum_analyze_threshold = 50000,
+                            autovacuum_analyze_scale_factor = 0.5
+                        )',
+        NEW.table_schema, NEW.table_name);
+    ELSE
+        --pg12 doesn't have autovacuum_vacuum_insert_threshold
+        EXECUTE format('CREATE TABLE %I.%I(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL) WITH
+                        (
+                            autovacuum_analyze_threshold = 50000,
+                            autovacuum_analyze_scale_factor=0.5
+                        )',
+        NEW.table_schema, NEW.table_name);
+    END IF;
     EXECUTE format('GRANT SELECT ON TABLE %I.%I TO prom_reader', NEW.table_schema, NEW.table_name);
     EXECUTE format('GRANT SELECT, INSERT ON TABLE %I.%I TO prom_writer', NEW.table_schema, NEW.table_name);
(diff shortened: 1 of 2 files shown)


migration/idempotent/001-base.sql
```diff
@@ -279,6 +279,30 @@ BEGIN
...
+   --but, we don't want to freeze anything since these tables will be truncated upon
+   --compression anyway (so keep freeze settings to the default).
+   --Make the vacuum less aggressive here by upping the threshold and scale factors. 
+   IF current_setting('server_version_num')::integer >= 130000 THEN
+            EXECUTE format('CREATE TABLE %I.%I(time TIMESTAMPTZ NOT NULL, value DOUBLE PRECISION NOT NULL, series_id BIGINT NOT NULL) WITH 
+                            ( 
+                              autovacuum_vacuum_threshold = 50000, autovacuum_vacuum_insert_threshold=50000, 
+                              autovacuum_vacuum_scale_factor=2.0, autovacuum_vacuum_insert_scale_factor=2.0,
+                              autovacuum_analyze_threshold = 50000,
+                              autovacuum_analyze_scale_factor = 1.0
```
pgnickb, line 292: This seems rather high. Analyze should generally be very cheap to perform and has positive impact on the performance, so 0.1 or 0.2 should be safe.

migration/idempotent/001-base.sql
```diff
@@ -1576,6 +1602,27 @@ BEGIN
...
+        INNER JOIN _timescaledb_catalog.hypertable c ON (h.compressed_hypertable_id= c.id) 
+        WHERE h.schema_name = 'prom_data' AND h.table_name = metric_table_name;
+
+        --Make the compressed tables freeze on every vacuum run. They won't be changed
+        --later anyway and there is no point requiring another vacuum before wraparound.
+        --Also, make sure autovacuum runs quickly after initial creation, by setting the
+        --threshold low.
+        IF current_setting('server_version_num')::integer >= 130000 THEN
+            EXECUTE FORMAT($$
+                ALTER TABLE %I,%I SET (autovacuum_freeze_min_age=0, autovacuum_freeze_table_age=0, autovacuum_vacuum_insert_threshold=5, autovacuum_vacuum_insert_scale_factor=0.0)
```
pgnickb, line 1617: I don't quite understand the 5 in the
```
autovacuum_vacuum_insert_threshold=5
```
But I imagine the whole table is populated in a single transaction so difference between 5 and 5M is negligible.

pgnickb (commented):
This ended up being longer than I expected, so I've added a tl;dr at the end.

Note:
The formula pg uses to make the decision to perform auto-vacuum reads:
```
threshold = vac_base_thresh + vac_scale_factor * reltuples
```
Where `threshold` is the number of inserted tuples needed to trigger auto-vacuum and `reltuples` is the estimate of number of rows in the table, which happens to also be updated by vacuum (exclusively). 

This effectively reads: for PostgreSQL 12 and earlier, unless the user runs a manual VACUUM on the table, `vac_scale_factor * reltuples` will be exactly 0.

In our case the `threshold` will never be met regardless of the setting, since the workload is insert-only. If these `relopts` aren't specified, then global VACUUM settings are used which might or might not be desirable.

The settings for analyze (counter-intuitively) act differently and in fact **do** take `inserts` into account, so those make sense.

With all above in mind, I'd say it is safe to remove the following `relopts` as they only apply to updates/deletes:
```
- autovacuum_vacuum_threshold
- autovacuum_vacuum_scale_factor
```

As for the `freeze` settings:
```
autovacuum_freeze_min_age=0, 
autovacuum_freeze_table_age=0
```
For PostgreSQL 13/14 it will work just fine, but for the earlier versions it can lead to some problems, as we don't expect auto-vacuum to ever trigger there under normal operation. This in turn means, that **if** it is triggered, it will end up scanning the entire table due to `autovacuum_freeze_table_age` being 0 (since there will be no frozen pages).

With the above in mind, I'd propose completely ditching any non-analyze vacuum settings for PostgreSQL 12 and earlier and focus on getting the most out of it for 13 and later.

If we want to have full support of these features on v12, then we shall consider running regular vacuums. One way to achieve this would be a background worker and a loop over:

```
SET statement_timeout = x;
VACUUM ANALYZE mytable;
```

tl;dr:
- Since the workload is insert-only, we should only keep changes to `analyze`, `vacuum_insert` and `freeze` for PostgreSQL v13 and later.
- v12 only needs `analyze` options, but those aren't very critical and can be omitted in favor of the defaults for the sake of simplicity.

migration/idempotent/001-base.sql
cevian: we've seen some issues with high writes to the stat directory so let's compromise on 0.5

migration/idempotent/001-base.sql
cevian: yeah I changed it to 1.

cevian:
@nick good points, addressed. One caveat on the compressed tables and PG12 I am still keeping the freeze settings because if an auto vac does happen, we still want it to freeze what it can

pgnickb (approved):

sumerman (approved):

Issue #134, opened 2022-03-17
Change trace_id from domain to new uuid-like type


jgpruitt:
Niksa:
>Good news on querying compressed chunks. With the new segment by approach (no segment by) we had a problem with TSDB query planner not pushing down filters that would lead to inefficient decompression of the whole chunk. As Mat pointed the planner gets confused when working with domain types (we wrapped UUID into domain type). I've just tried using plain UUIDs and all work well!

PR #163, opened 2022-03-31
Custom type for trace_id

niksajakovljevic:
Using domain type didn't play nicely with TimescaleDB planner: some pushdowns
were not working well on compressed data (https://github.com/timescale/timescaledb/issues/4179). 
Here we add custom type that wrapps around PG UUID type.
The idea is to later on customize this type - eg. string representation.
Since trace_id is used for data partitioning this change is not backwards
compatible.

Closes https://github.com/timescale/promscale_extension/issues/134

### DIFF
--- a/migration/idempotent/014-trace-id-functions.sql
+++ b/migration/idempotent/014-trace-id-functions.sql
@@ -1,0 +1,84 @@
+-- functions for trace_id custom type which is a wrapper around PG UUID type
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_in(cstring)
+RETURNS ps_trace.trace_id
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_in$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_in(cstring) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_out(ps_trace.trace_id)
+RETURNS cstring
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_out$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_out(ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_send(ps_trace.trace_id)
+RETURNS bytea
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_send$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_send(ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_recv(internal)
+RETURNS ps_trace.trace_id
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_recv$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_recv(internal) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_ne(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS bool
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_ne$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_ne(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_eq(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS bool
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_eq$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_eq(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_ge(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS bool
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_ge$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_ge(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_le(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS bool
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_le$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_le(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_gt(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS bool
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_gt$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_gt(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_lt(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS bool
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_lt$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_lt(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_cmp(ps_trace.trace_id, ps_trace.trace_id)
+RETURNS int
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_cmp$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_cmp(ps_trace.trace_id, ps_trace.trace_id) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_hash(ps_trace.trace_id)
+RETURNS int
+LANGUAGE internal
+IMMUTABLE PARALLEL SAFE STRICT
+AS $function$uuid_hash$function$;
+GRANT EXECUTE ON FUNCTION _ps_trace.trace_id_hash(ps_trace.trace_id) TO prom_reader;
--- a/migration/migration/015-tracing-compression.sql
+++ b/migration/migration/015-tracing-compression.sql
@@ -1,67 +1,339 @@
+-- Due to incompatible change of trace_id type we need to drop tables and delete data
+DROP TABLE _ps_trace.link CASCADE;
+DROP TABLE _ps_trace.event CASCADE;
+DROP TABLE _ps_trace.span CASCADE;
 
--- Fix compression on trace tables
+TRUNCATE TABLE _ps_trace.tag;
+DELETE FROM _ps_trace.tag_key WHERE id >= 1000;
+ANALYZE _ps_trace.tag_key;
+REINDEX TABLE _ps_trace.tag_key;
+PERFORM setval('_ps_trace.tag_key_id_seq', 1000);
+TRUNCATE TABLE _ps_trace.operation;
+ANALYZE _ps_trace.operation;
+TRUNCATE TABLE _ps_trace.instrumentation_lib CASCADE;
+ANALYZE _ps_trace.instrumentation_lib;
+TRUNCATE TABLE _ps_trace.schema_url CASCADE;
+ANALYZE _ps_trace.schema_url;
+
+-- We are replacing domain with custom type
+DROP DOMAIN ps_trace.trace_id CASCADE;
+
+CREATE TYPE ps_trace.trace_id;
+
+CREATE OR REPLACE FUNCTION _ps_trace.trace_id_in(cstring)
+RETURNS ps_trace.trace_id
+LANGUAGE internal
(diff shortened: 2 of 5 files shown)


niksajakovljevic:
Please hold on reviewing. It seems that TimescaleDB code doesn't play nicely with custom supplied hash function for partitioning.

niksajakovljevic:
Problem is gone after making hash operator class default (so custom type hash function is auto resolved by PG).

migration/migration/012-tracing.sql
```diff
@@ -1,7 +1,158 @@
...
+        RECEIVE = ps_trace.trace_id_recv,
+        alignment = double
+    );
+
+    CREATE OPERATOR ps_trace.= ( 
+        FUNCTION=ps_trace.trace_id_eq,
+        LEFTARG=ps_trace.trace_id,
+        RIGHTARG=ps_trace.trace_id,
+        COMMUTATOR= OPERATOR(ps_trace.=),
+        NEGATOR= OPERATOR(ps_trace.!=),
```
sumerman, line 91: Is there a meaningful difference between `<>` and `!=`?

sumerman (changes requested):
The change itself LGTM, but there's no upgrade path for beta users. From what I remember, 0.0.0 was suggested as a good place to perform it.

migration/migration/012-tracing.sql
niksajakovljevic: I believe not

niksajakovljevic:
@sumerman Like state in the comment this is a breaking change. So the only way existing users can move on is to drop all the existing data. Let's talk more offline.

jgpruitt:
@niksajakovljevic (assuming I ran them correctly) the `trace_operation_calls_test` and `TestCompareTraceQueryResponse` tests on the connector side fail against this branch. I think you'll need to modify those tests.

Also, have you run the upgrade tests against this branch?

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,325 @@
...
 
--- Fix compression on trace tables
+TRUNCATE TABLE _ps_trace.tag;
+DELETE FROM _ps_trace.tag_key WHERE id >= 1000;
+ANALYZE _ps_trace.tag_key;
+REINDEX TABLE _ps_trace.tag_key;
+PERFORM setval('_ps_trace.tag_key_id_seq', 1000);
+TRUNCATE TABLE _ps_trace.operation;
+TRUNCATE TABLE _ps_trace.instrumentation_lib CASCADE;
+
```
jgpruitt, line 13: `ANALYZE` all the truncated tables?

jgpruitt (changes requested):

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,325 @@
+-- Due to incompatible change of trace_id type we need to drop tables and delete data
+DROP TABLE _ps_trace.link;
+DROP TABLE _ps_trace.event;
+DROP TABLE _ps_trace.span;
 
--- Fix compression on trace tables
+TRUNCATE TABLE _ps_trace.tag;
+DELETE FROM _ps_trace.tag_key WHERE id >= 1000;
+ANALYZE _ps_trace.tag_key;
```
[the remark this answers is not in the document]
niksajakovljevic: I had that initially but `VACUUM` can't run inside a function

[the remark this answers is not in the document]
jgpruitt: oh yeah, i forgot about that. bummer

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,341 @@
...
+REINDEX TABLE _ps_trace.tag_key;
+PERFORM setval('_ps_trace.tag_key_id_seq', 1000);
+TRUNCATE TABLE _ps_trace.operation;
+ANALYZE _ps_trace.operation;
+TRUNCATE TABLE _ps_trace.instrumentation_lib CASCADE;
+ANALYZE _ps_trace.instrumentation_lib;
+TRUNCATE TABLE _ps_trace.schema_url CASCADE;
+ANALYZE _ps_trace.schema_url;
+
+CALL _prom_catalog.execute_everywhere('tracing_types', $ee$ DO $$ BEGIN
```
JamesGuthrie, line 18: As discussed on Slack, it's not clear to me that we should be adding this type in an `execute_everywhere` block. I think instead we should be running something like `CREATE EXTENSION promscale`.

JamesGuthrie (changes requested):

JamesGuthrie (changes requested):

migration/migration/015-tracing-compression.sql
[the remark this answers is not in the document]
niksajakovljevic: That is a separate design decision and I don't think we should be conflating that with a newly added custom type. At the moment we are not installing Promscale extension on data nodes.

niksajakovljevic: That is a separate design decision and I don't think we should be conflating that with a newly added custom type. At the moment we are not installing Promscale extension on data nodes.

niksajakovljevic: fyi `execute_everywhere` mechanism is already in use in many different places

cevian (changes requested):

migration/migration/015-tracing-compression.sql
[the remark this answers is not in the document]
niksajakovljevic: I believe that components on Cloud are provisioned using superuser so that should work without problems

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,341 @@
...
+    AS $function$uuid_hash$function$;
+    GRANT EXECUTE ON FUNCTION ps_trace.trace_id_hash(ps_trace.trace_id) TO prom_reader;
+
+    CREATE TYPE ps_trace.trace_id(
+        internallength = 16,
+        INPUT = ps_trace.trace_id_in,
+        OUTPUT = ps_trace.trace_id_out,
+        SEND = ps_trace.trace_id_send,
+        RECEIVE = ps_trace.trace_id_recv,
+        alignment = double
```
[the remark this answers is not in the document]
niksajakovljevic: i put double because I assumed it is stored as 16 bytes array, but seems it's actually a pointer

cevian (changes requested):
seeing some failures in here: https://github.com/timescale/promscale/runs/5854892743?check_suite_focus=true

cevian:
Ahh never mind about above -- investigating

cevian:
Ok even after rebase, there is a problem in the update path. See https://github.com/timescale/promscale/runs/5856026303?check_suite_focus=true

cevian:
" ERROR: cannot drop table _ps_trace.link because other objects depend on it (SQLSTATE 2BP01) . continuing without extension"

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,339 @@
+-- Due to incompatible change of trace_id type we need to drop tables and delete data
+DROP TABLE _ps_trace.link;
```
jgpruitt, line 2: probably need `cascade` on these drops

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,339 @@
...
+PERFORM setval('_ps_trace.tag_key_id_seq', 1000);
+TRUNCATE TABLE _ps_trace.operation;
+ANALYZE _ps_trace.operation;
+TRUNCATE TABLE _ps_trace.instrumentation_lib CASCADE;
+ANALYZE _ps_trace.instrumentation_lib;
+TRUNCATE TABLE _ps_trace.schema_url CASCADE;
+ANALYZE _ps_trace.schema_url;
+
+-- We are replacing domain with custom type
+DROP DOMAIN ps_trace.trace_id;
```
jgpruitt, line 19: probably need `cascade` on drop domain

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,339 @@
...
+
+CREATE OPERATOR CLASS ps_trace.hash_trace_id_ops
+DEFAULT FOR TYPE ps_trace.trace_id USING hash AS 
+    OPERATOR 1 ps_trace.=,
+    FUNCTION 1 ps_trace.trace_id_hash(ps_trace.trace_id);
+
+GRANT USAGE ON TYPE ps_trace.trace_id TO prom_reader;
+
+
+CREATE TABLE IF NOT EXISTS _ps_trace.span
```
jgpruitt, line 188: remove the `if not exists` as it is a security issue

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,339 @@
...
+);
+CREATE INDEX ON _ps_trace.span USING BTREE (trace_id, parent_span_id) INCLUDE (span_id); -- used for recursive CTEs for trace tree queries
+CREATE INDEX ON _ps_trace.span USING GIN (span_tags jsonb_path_ops); -- supports tag filters. faster ingest than json_ops
+CREATE INDEX ON _ps_trace.span USING BTREE (operation_id); -- supports filters/joins to operation table
+--CREATE INDEX ON _ps_trace.span USING GIN (jsonb_object_keys(span_tags) array_ops); -- possible way to index key exists
+CREATE INDEX ON _ps_trace.span USING GIN (resource_tags jsonb_path_ops); -- supports tag filters. faster ingest than json_ops
+GRANT SELECT ON TABLE _ps_trace.span TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.span TO prom_writer;
+
+CREATE TABLE IF NOT EXISTS _ps_trace.event
```
jgpruitt, line 220: remove `if not exists`

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,339 @@
...
+    name text NOT NULL,
+    tags ps_trace.tag_map NOT NULL,
+    dropped_tags_count int NOT NULL DEFAULT 0
+);
+CREATE INDEX ON _ps_trace.event USING GIN (tags jsonb_path_ops);
+CREATE INDEX ON _ps_trace.event USING BTREE (trace_id, span_id);
+GRANT SELECT ON TABLE _ps_trace.event TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.event TO prom_writer;
+
+CREATE TABLE IF NOT EXISTS _ps_trace.link
```
jgpruitt, line 235: remove `if not exists`

jgpruitt (changes requested):

migration/migration/015-tracing-compression.sql
jgpruitt: there are views referencing these tables. idempotent scripts will add the views back

sql/promscale--0.0.0.sql
```diff
@@ -448,7 +468,9 @@ BEGIN
         ('011-tables-exemplar.sql'        , '0.0.0'),
         ('012-tracing.sql'                , '0.0.0'),
         ('013-tracing-well-known-tags.sql', '0.0.0'),
-        ('014-telemetry.sql'              , '0.0.0')
+        ('014-telemetry.sql'              , '0.0.0'),
+        ('015-tracing-compression.sql'    , '0.0.0'),
+        ('015-tracing-large-tags-fix.sql' , '0.0.0')
```
jgpruitt, line 473: These need to be removed. Adding them here says that these migrations have already been applied, however, these migrations show up in `0.5.0`. So, they need to be applied both in the upgrade and fresh install paths. Removing them from this list will cause them to be applied in both cases.

migration/migration/015-tracing-compression.sql
```diff
@@ -1,67 +1,339 @@
...
+ANALYZE _ps_trace.instrumentation_lib;
+TRUNCATE TABLE _ps_trace.schema_url CASCADE;
+ANALYZE _ps_trace.schema_url;
+
+-- We are replacing domain with custom type
+DROP DOMAIN ps_trace.trace_id CASCADE;
+
+CREATE TYPE ps_trace.trace_id;
+
+CREATE OR REPLACE FUNCTION ps_trace.trace_id_in(cstring)
```
jgpruitt, line 23: The operators should be in the `ps_trace` schema but the functions backing them should go in the `_ps_trace` schema.

migration/idempotent/014-trace-id-functions.sql
```diff
@@ -0,0 +1,84 @@
+-- functions for trace_id custom type which is a wrapper around PG UUID type
+CREATE OR REPLACE FUNCTION ps_trace.trace_id_in(cstring)
```
jgpruitt, line 2: These functions should all go in the `_ps_trace` schema

jgpruitt (changes requested):

sql/promscale--0.0.0.sql
```diff
@@ -276,6 +276,18 @@ BEGIN
...
+        , ('FUNCTION', 'ps_trace.trace_id_send(ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_recv(internal)')
+        , ('FUNCTION', 'ps_trace.trace_id_ne(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_eq(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_ge(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_le(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_gt(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_lt(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_cmp(ps_trace.trace_id, ps_trace.trace_id)')
+        , ('FUNCTION', 'ps_trace.trace_id_hash(ps_trace.trace_id)')
```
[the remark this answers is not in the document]
niksajakovljevic: I think we should do a better work in having code comments explaining the process.

sql/promscale--0.0.0.sql
niksajakovljevic: Ok that was not clear. I hope we can make code comment more meaningful. 
eg. what we have now `-- Bring migrations table up to speed` doesn't look descriptive :)

sql/promscale--0.0.0.sql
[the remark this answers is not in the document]
niksajakovljevic: Also I am not sure I follow you logic. Let's discuss this more offline

cevian (approved):

JamesGuthrie (approved):

PR #166, opened 2022-04-01
Remove setting of search path during extension install

JamesGuthrie:


jgpruitt (approved):

cevian (approved):

JamesGuthrie:
I'm going to close this. We're taking a slightly different approach in #198.

PR #168, opened 2022-04-04
Autogenerate function and operator documentation

JamesGuthrie:


pgnickb:
This looks really cool and is a huge step forward imo.
I'd like to propose some minor changes though:
1. Break down into 3 sections: Functions, Operators, Procedures
2.Change format of first column to `|{}\n{}|` so that the name occupies less space, put schema in bold if needed
3. Add a Table of Contents pointing to any of those 3 sections

I can push some code in a separate branch to demonstrate what I mean, but I'd be happy to hear what others think first.

sumerman (approved):
Not sure why do we need gendoc inside the docker images, LGTM otherwise! ❤️

JamesGuthrie:
> Not sure why do we need gendoc inside the docker images, LGTM otherwise! ❤️

Ya, this is a little silly. It's because I've setup `gendoc` as a workspace project, so Cargo complains if the sources aren't present when compiling the extension. We could move this to a separate (non-workspace) subdir.

JamesGuthrie:
I've also reworked the output to be heading-per-function/operator as suggested by others. I know that the formatting isn't perfect yet, but let's improve the documentation before we improve the formatting 😅 .

sumerman:
> > Not sure why do we need gendoc inside the docker images, LGTM otherwise! ❤️
> 
> Ya, this is a little silly. It's because I've setup `gendoc` as a workspace project, so Cargo complains if the sources aren't present when compiling the extension. We could move this to a separate (non-workspace) subdir.

I don't a strong opinion on that, but was curious why it's there.

sumerman (approved):

Issue #109, opened 2022-03-08
Make trusted extension optional (for PG12)


JamesGuthrie:
We plan to support PG12 with the next extension release. This is not compatible with having the extension marked as trusted (which is only available since PG13). We would like to have the extension trusted because it reduces the privileges required to install it.

For pg12 it would be okay to either require superuser, or to fall back to something like pgextwlist.

PR #171, opened 2022-04-04
Make trusted extension optional (for PG12)

sumerman:
- `promscale.control` is generated via `build.rs` based on pg version
- CI tests and image builds are re-enabled on pg12
- Migrations, involving `@extowner@`, had to be altered to be compatible with pg12.
- Some tests had to be generalized to pass on both pg14 and pg12

Addresses #109 

Upgrade and e2e tests pass https://github.com/timescale/promscale/pull/1287

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -21,11 +21,15 @@ struct IdempotentWrapper<'a> {
     body: &'a str,
 }
 
+#[derive(Template)]
+#[template(path = "promscale.control", escape = "none")]
+struct ControlFile {
+    is_pg_12: bool,
+}
+
+const CONTROL_FILE_NAME: &str = "promscale.control";
 const MIGRATION_FILE_NAME: &str = "hand-written-migration.sql";
 
-/// This build script reads the SQL files placed in the `migration/{idempotent,migration}`
-/// directories, wraps each file in its respective template (in the `templates` directory), and
-/// outputs the whole contents as one long SQL file (`MIGRATION_FILE_NAME`) in the root directory.
 fn main() {
     println!("cargo:rerun-if-changed=Cargo.lock");
     // Note: directories are not traversed, instead Cargo looks at the mtime of the directory.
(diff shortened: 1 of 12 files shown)


sql/promscale--0.0.0.sql
```diff
@@ -454,19 +454,23 @@ END
...
+        GRANT prom_reader TO %1$s WITH ADMIN OPTION;
+        GRANT prom_writer TO %1$s WITH ADMIN OPTION;
+        GRANT prom_maintenance TO %1$s WITH ADMIN OPTION;
+        GRANT prom_modifier TO %1$s WITH ADMIN OPTION;
+        GRANT prom_admin TO %1$s WITH ADMIN OPTION;
     END
     $$;
-$ee$, '@extowner@'
+$ee$, CASE
+        WHEN version() LIKE 'PostgreSQL 12%' THEN quote_ident(current_user)
```
jgpruitt, line 471: I think this might be "safer". At least, it "feels safer" to me.

```
CASE WHEN current_setting('server_version') like '12.%' THEN ...
```

migration/migration/003-users.sql
```diff
@@ -65,14 +65,21 @@ CALL _prom_catalog.execute_everywhere('grant_prom_reader_prom_writer',$ee$
...
+        GRANT prom_reader TO %1$s WITH ADMIN OPTION;
+        GRANT prom_writer TO %1$s WITH ADMIN OPTION;
+        GRANT prom_maintenance TO %1$s WITH ADMIN OPTION;
+        GRANT prom_modifier TO %1$s WITH ADMIN OPTION;
+        GRANT prom_admin TO %1$s WITH ADMIN OPTION;
     END
     $$;
-$ee$);
+$ee$, CASE
+        WHEN version() LIKE 'PostgreSQL 12%' THEN quote_ident(current_user)
```
jgpruitt, line 81: I think this might be "safer". At least, it "feels safer" to me.

```
CASE WHEN current_setting('server_version') like '12.%' THEN ...
```

jgpruitt: Will be a long time before PostgreSQL 120 is released, but might at least consider `12.%` instead of `12%`. Did version `12.0` include the `.0` in this value?

jgpruitt (changes requested):

migration/migration/003-users.sql
sumerman: Great catch, thank you!

ha.Dockerfile
```diff
@@ -54,10 +54,11 @@ COPY --chown=postgres:postgres templates/ /build/promscale/templates/
 
 RUN make package
 
-FROM timescale/timescaledb-ha:pg${PG_VERSION}-ts${TIMESCALEDB_VERSION_MAJMIN}-latest
+FROM timescale/timescaledb-ha:pg14-ts${TIMESCALEDB_VERSION_MAJMIN}-latest
```
jgpruitt, line 57: is this supposed to be `pg14`?

jgpruitt (changes requested):

jgpruitt (approved):

ha.Dockerfile
cevian, line 57: Given that this docker file will only be used for testing, I think this is OK because this image also contains 12,13,14. Let's just add a comment

src/aggregates/vector_selector.rs
```diff
@@ -814,55 +813,21 @@ mod tests {
                 .expect("SQL query failed");
 
         // Assert that we're running in parallel mode.
-        // Note: this query plan is specific to PG14.
+        let top_level_plan = parallel_plan.0[0]["Plan"].clone();
```
cevian, line 816: This is great!

cevian (approved):
LGTM

ha.Dockerfile
sumerman: Sure! We can't do much about it anyway. It's a cloud's decision to have a universal image that contains all currently supported versions.

sumerman: Yes. Will add a comment.

sumerman: I guess we could take their `Dockerfile` and build with our own set of ARGs but I'm not sure we really wanna do this.

build.rs
```diff
@@ -21,29 +21,61 @@ struct IdempotentWrapper<'a> {
...
-/// directories, wraps each file in its respective template (in the `templates` directory), and
-/// outputs the whole contents as one long SQL file (`MIGRATION_FILE_NAME`) in the root directory.
 fn main() {
     println!("cargo:rerun-if-changed=Cargo.lock");
     // Note: directories are not traversed, instead Cargo looks at the mtime of the directory.
     println!("cargo:rerun-if-changed=migration");
     println!("cargo:rerun-if-changed=migration/migration");
     println!("cargo:rerun-if-changed=migration/idempotent");
     println!("cargo:rerun-if-changed=templates");
+    println!("cargo:rerun-if-changed={}", CONTROL_FILE_NAME);
```
JamesGuthrie, line 40: This causes `build.rs` to always re-compile, because as specified, this is an output which `build.rs` is producing. Did you mean:

```
println!("cargo:rerun-if-changed=templates/{}", CONTROL_FILE_NAME);
```

src/aggregates/vector_selector.rs
JamesGuthrie: Yes! This is so much better than the hacky thing I had implemented!

JamesGuthrie (changes requested):
The only thing that bothers me about generating the control file like this is that we add another step in the development process. `cargo pgx run` will fail unless one first runs `make promscale.control`. The reason is that `cargo-pgx` wants to inspect the control file before calling the build process (which produces the control file).

I guess one potential solution could be to add a `run` make target, which would look like this:

```make
.PHONY: run
run: promscale.control ## Custom wrapper around cargo pgx run
	cargo pgx run
```

build.rs
JamesGuthrie: I suspect that something else weird is going on here. I'm looking into it.

JamesGuthrie (approved):

PR #173, opened 2022-04-05
Fixes to ownership model

cevian:


sql/promscale--0.0.0.sql
```diff
@@ -372,40 +381,55 @@ END;
...
+     _rec record;
+ BEGIN
+     FOR _rec IN
+     (
+         SELECT e.table_name
+         FROM _prom_catalog.exemplar e
+         ORDER BY e.table_name
+     )
+     LOOP
+         EXECUTE format($sql$ALTER TABLE prom_data_exemplar.%I OWNER TO %I$sql$, _rec.table_name, current_user);
```
jgpruitt, line 417: should this also be prom_admin?

jgpruitt (changes requested):

sql/promscale--0.0.0.sql
cevian: Yep changed

jgpruitt (approved):

migration/idempotent/001-base.sql
```diff
@@ -283,9 +282,16 @@ BEGIN
...
+   EXECUTE format('GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE %I.%I TO prom_modifier', NEW.table_schema, NEW.table_name);
    EXECUTE format('CREATE UNIQUE INDEX data_series_id_time_%s ON %I.%I (series_id, time) INCLUDE (value)',
                     NEW.id, NEW.table_schema, NEW.table_name);
+  
+   --dynamically created tables are owned by our highest-privileged role, prom_admin
+   --these cannot be made part of the extension since we cannot make them config
+   --tables outside of extension upgrade/install scripts. They cannot be superuser
+   --owned without being part extension since that would prevent dump/restore from
+   --working on any environment without SU privileges (such as cloud).
+   EXECUTE format('ALTER TABLE %I.%I OWNER TO prom_admin', NEW.table_schema, NEW.table_name);
```
sumerman, line 294: From the comment, I understand why to pick `prom_admin` over a superuser. Yet, It's not quite clear to me why the `prom_admin` role works while `prom_modifier` doesn't. From what I remember `ALTER TABLE ... OWNER TO` construct itself was a problem on multi-node.

cevian: we switched prom_modifier->prom_admin not because of multinode but because prom-admin is our most-privileged user and thus it is the user that should own these things. The multinode fix is moving the ownership change to before hypertable creation

sumerman (approved):

migration/idempotent/001-base.sql
```diff
@@ -3110,3 +3119,85 @@ END;
 $$
 LANGUAGE PLPGSQL;
 GRANT EXECUTE ON FUNCTION _prom_catalog.insert_metric_row(NAME, TIMESTAMPTZ[], DOUBLE PRECISION[], BIGINT[]) TO prom_writer;
+
+CREATE OR REPLACE FUNCTION _prom_catalog.has_permission_on_table(_table regclass)
```
JamesGuthrie, line 3123: It seems as though this function is now unused, or am I missing something?

JamesGuthrie (changes requested):

migration/idempotent/001-base.sql
```diff
@@ -3076,7 +3085,7 @@ BEGIN
     END IF;
 END
 $func$ LANGUAGE PLPGSQL;
-GRANT EXECUTE ON PROCEDURE prom_api.add_prom_node(TEXT, BOOLEAN) TO prom_writer;
+-- add_prom_node is a superuser only function
```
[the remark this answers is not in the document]
cevian: yes. we are gonna have to figure this out later, but for now let's lock down the permissions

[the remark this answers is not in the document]
JamesGuthrie: 👍

JamesGuthrie (approved):

PR #175, opened 2022-04-06
Optimize builds

cevian:
The sql generator will by default use the develop profile but
since pgx package uses the release profile it is better to use
the release profile for the sql generation too. Otherwise,
everything needs to be compiled two times.


sumerman (approved):

JamesGuthrie (approved):
🎉

jgpruitt (approved):

PR #176, opened 2022-04-07
Align trace tables columns to save on space

antekresic:
Aligment of columns with respect to their sizes and alignment allows us to
save some storage space per row. From my simple test, I was able to reduce
the size from 140 to 129.95 bytes on average for the span table alone.


JamesGuthrie (changes requested):
I like the general idea behind this change, but a change like this can't be applied in this manner.

We can't modify existing migration files, this results in older installs having a different data layout than newer installs, and is bound to cause a problem somewhere.

Niksa's [custom type for trace_id change](https://github.com/timescale/promscale_extension/pull/163/files) illustrates how this would be done correctly. Considering that he's already dropping and re-adding the tables that you want to modify in his change, perhaps you could apply these suggestions to that change?

JamesGuthrie:
On another note, it would be amazing to have a linter for this.

antekresic:
Closing this in favor collaborating with Niksa to incorporate in his PR:
https://github.com/timescale/promscale_extension/pull/163/files

PR #178, opened 2022-04-08
Use correct approximate row count function

JamesGuthrie:
`hypertable_approximate_row_count` was dropped in Timescale 2.0 [1].

[1]: https://github.com/timescale/timescaledb/blob/332dffeebc72cf16afbde802e2f19aead9bd7836/sql/updates/1.7.5--2.0.0-rc1.sql#L13


JamesGuthrie:
Oh man, I missed something obvious 🤦

PR #179, opened 2022-04-08
Fix most security vulnerabilities

JamesGuthrie:
This change applies a number of changes to the extension's install SQL
script to make it not vulnerable to unsafe object creation and unsafe
search_path attacks (for more details, see migration/README.md).

This change does not address unsafe object creation for the `_prom_ext`
schema, which will be addressed in a follow-on change.

Using SET LOCAL additionally restricts the search_path within the
procedure. Note that this _only_ works when used with a procedure (not a
function), and when that procedure does transaction control. These two
preconditions ensure that the search_path configured in SET LOCAL does
not persist past the function call.

The relevant pieces of Postgres documentation are:

1. From the documentation on SET [1]:

> The effects of SET LOCAL last only till the end of the current
> transaction, whether committed or not.

2. From the documentation on CALL [2]:

> If CALL is executed in a transaction block, then the called procedure
> cannot execute transaction control statements. Transaction control
> statements are only allowed if CALL is executed in its own transaction

[1]: https://www.postgresql.org/docs/14/sql-set.html
[2]: https://www.postgresql.org/docs/14/sql-call.html


migration/idempotent/001-base.sql
```diff
@@ -1995,7 +2062,8 @@ LANGUAGE PLPGSQL STABLE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.get_cagg_info(text, text) TO prom_reader;
 
 CREATE OR REPLACE FUNCTION prom_api.is_stale_marker(value double precision)
-RETURNS BOOLEAN
+    RETURNS BOOLEAN
+    SET search_path = pg_catalog
```
cevian, line 2066: this should def by inlinable so let's remove this and just schema qualify the pg_catalog.float8send

migration/idempotent/001-base.sql
```diff
@@ -2017,6 +2086,7 @@ GRANT EXECUTE ON FUNCTION prom_api.is_normal_nan(double precision) TO prom_reade
 CREATE OR REPLACE FUNCTION prom_api.val(
         label_id INT)
     RETURNS TEXT
+    SET search_path = pg_catalog
```
cevian, line 2089: we should probably remove this and make it inlinable

migration/idempotent/001-base.sql
```diff
@@ -3009,6 +3073,7 @@ GRANT EXECUTE ON FUNCTION _prom_catalog.get_metrics_that_need_compression() TO p
 
 --only for timescaledb 2.0 in 1.x we use compression policies
 CREATE OR REPLACE PROCEDURE _prom_catalog.execute_compression_policy(log_verbose boolean = false)
+-- Note: Cannot SET search_path because we do transaction control
```
cevian, line 3076: txn control same question about set local

migration/idempotent/007-tracing-tags.sql
```diff
@@ -38,17 +40,18 @@ LANGUAGE SQL STABLE PARALLEL SAFE;
...
+    SELECT _tag_map OPERATOR(pg_catalog.@>) ANY(_ps_trace.eval_tags_by_key(_key))
 $func$
 LANGUAGE SQL STABLE PARALLEL SAFE;
 GRANT EXECUTE ON FUNCTION _ps_trace.has_tag(ps_trace.tag_map, ps_trace.tag_k) TO prom_reader;
 COMMENT ON FUNCTION _ps_trace.has_tag IS $$This function supports the #? operator.$$;
 
 DO $do$
 BEGIN
-    CREATE OPERATOR ps_trace.#? (
+    CREATE OPERATOR ps_trace.# (
```
cevian, line 54: why was the name of the operator changed?

migration/idempotent/007-tracing-tags.sql
```diff
@@ -99,25 +104,27 @@ $do$;
...
     FROM _ps_trace.tag a
-    WHERE a.key = _op.tag_key
+    WHERE a.key OPERATOR(pg_catalog.=) _op.tag_key
     -- if the jsonb value is a string, apply the regex directly
     -- otherwise, convert the value to a text representation, back to a jsonb string, and then apply
     AND CASE jsonb_typeof(a.value)
-        WHEN 'string' THEN jsonb_path_exists(a.value, format('$?(@ like_regex "%s")', _op.value)::jsonpath)
-        ELSE jsonb_path_exists(to_jsonb(a.value#>>'{}'), format('$?(@ like_regex "%s")', _op.value)::jsonpath)
+        WHEN 'string' THEN pg_catalog.jsonb_path_exists(a.value, pg_catalog.format('$?(@ like_regex "%s")', _op.value)::jsonpath)
+        ELSE pg_catalog.jsonb_path_exists(to_jsonb(a.value #>> '{}'), pg_catalog.format('$?(@ like_regex "%s")', _op.value)::jsonpath)
```
cevian, line 117: the #>> operator needs to be schema_qualified

migration/idempotent/007-tracing-tags.sql
```diff
@@ -140,25 +147,27 @@ $do$;
...
     FROM _ps_trace.tag a
-    WHERE a.key = _op.tag_key
+    WHERE a.key OPERATOR(pg_catalog.=) _op.tag_key
     -- if the jsonb value is a string, apply the regex directly
     -- otherwise, convert the value to a text representation, back to a jsonb string, and then apply
     AND CASE jsonb_typeof(a.value)
-        WHEN 'string' THEN jsonb_path_exists(a.value, format('$?(!(@ like_regex "%s"))', _op.value)::jsonpath)
-        ELSE jsonb_path_exists(to_jsonb(a.value#>>'{}'), format('$?(!(@ like_regex "%s"))', _op.value)::jsonpath)
+        WHEN 'string' THEN pg_catalog.jsonb_path_exists(a.value, pg_catalog.format('$?(!(@ like_regex "%s"))', _op.value)::jsonpath)
+        ELSE pg_catalog.jsonb_path_exists(to_jsonb(a.value #>> '{}'), pg_catalog.format('$?(!(@ like_regex "%s"))', _op.value)::jsonpath)
```
cevian, line 160: the #>> operator needs to be schema qualified

migration/idempotent/010-telemetry.sql
```diff
@@ -130,7 +134,22 @@ $$
...
+CREATE OR REPLACE FUNCTION _prom_ext.update_tsprom_metadata(meta_key text, meta_value text, send_telemetry BOOLEAN)
+RETURNS VOID
+SECURITY DEFINER
+SET search_path = pg_catalog
+AS $func$
+    INSERT INTO _timescaledb_catalog.metadata(key, value, include_in_telemetry)
+    VALUES ('promscale_' || meta_key,meta_value, send_telemetry)
+    ON CONFLICT (key) DO UPDATE SET value = EXCLUDED.value, include_in_telemetry = EXCLUDED.include_in_telemetry
+$func$
+LANGUAGE SQL VOLATILE;
```
cevian, line 147: since this is security definer should we revoke all from public like we do with other sec definers?

migration/idempotent/011-maintenance.sql
```diff
@@ -73,19 +79,17 @@ BEGIN
...
 --redundant given schema settings but extra caution for security definers
 REVOKE ALL ON FUNCTION _prom_catalog.drop_metric_chunk_data(text, text, timestamptz) FROM PUBLIC;
 GRANT EXECUTE ON FUNCTION _prom_catalog.drop_metric_chunk_data(text, text, timestamptz) TO prom_maintenance;
 
 --drop chunks from metrics tables and delete the appropriate series.
 CREATE OR REPLACE PROCEDURE _prom_catalog.drop_metric_chunks(
     schema_name TEXT, metric_name TEXT, older_than TIMESTAMPTZ, ran_at TIMESTAMPTZ = now(), log_verbose BOOLEAN = FALSE
-) AS $func$
+)
+-- Note: Cannot SET search_path because we do transaction control
```
cevian, line 91: another txn control question

migration/migration-table.sql
```diff
@@ -5,7 +5,11 @@ $migration_table$
         _current_user_id oid = NULL;
         _ps_catalog_schema_owner_id oid = NULL;
         _migration_table_owner_id oid = NULL;
+        _old_search_path text;
     BEGIN
+        EXECUTE 'SHOW search_path' INTO STRICT _old_search_path;
+        SET search_path TO pg_catalog;
```
cevian, line 11: why don't we do SET local here instead? That way we don't need to restore

migration/migration/004-schemas.sql
```diff
@@ -1,54 +1,39 @@
...
-DECLARE
-   new_path text;
-BEGIN
-   new_path := current_setting('search_path') || format(',%L,%L,%L,%L,%L,%L', 'ps_tag', '_prom_ext', 'prom_api', 'prom_metric', '_prom_catalog', 'ps_trace');
-   execute format('ALTER DATABASE %I SET search_path = %s', current_database(), new_path);
-   execute format('SET search_path = %s', new_path);
-END
-$$;
+-- _ps_catalog is created before
+GRANT USAGE ON SCHEMA _ps_catalog TO prom_reader;
```
cevian, line 39: this seem like a duplicate of line 2?

migration/migration/004-schemas.sql
```diff
@@ -1,54 +1,39 @@
-CALL _prom_catalog.execute_everywhere('create_schemas', $ee$ DO $$ BEGIN
-    CREATE SCHEMA IF NOT EXISTS _prom_catalog; -- this will be limited to metric and probably renamed in future
-    GRANT USAGE ON SCHEMA _prom_catalog TO prom_reader;
+-- _prom_catalog is created before
```
cevian, line 1: why are we removing the execute_everywhere?

templates/idempotent-wrapper.sql
```diff
@@ -1,8 +1,17 @@
 
 -- {{filename}}
 DO $outer_idempotent_block$
+DECLARE
+    _old_search_path text;
 BEGIN
+    EXECUTE 'SHOW search_path' INTO STRICT _old_search_path;
+    SET search_path TO pg_catalog;
```
cevian, line 8: we should just be able to use SET local here and avoid having to restore

templates/migration-wrapper.sql
```diff
@@ -7,7 +7,11 @@ DECLARE
     _body_differs BOOL = false;
     _migration _ps_catalog.migration = row ('{{filename}}', '{{version}}');
     _body TEXT = $migrationbody${{body}}$migrationbody$;
+    _old_search_path text;
 BEGIN
+    EXECUTE 'SHOW search_path' INTO STRICT _old_search_path;
+    SET search_path TO pg_catalog;
```
cevian, line 13: we should just be able to use SET local here and avoid having to restore

cevian (changes requested):
Looks good overall. Some pretty minor comments. Since this PR is so big lets make sure to add commits instead of squashing while we are reviewing

migration/idempotent/001-base.sql
```diff
@@ -658,25 +676,29 @@ GRANT EXECUTE ON FUNCTION _prom_catalog.get_new_label_id(text, text) to prom_wri
 --wrapper around jsonb_each_text to give a better row_estimate
 --for labels (10 not 100)
 CREATE OR REPLACE FUNCTION _prom_catalog.label_jsonb_each_text(js jsonb, OUT key text, OUT value text)
- RETURNS SETOF record
- LANGUAGE INTERNAL
- IMMUTABLE PARALLEL SAFE STRICT ROWS 10
+    RETURNS SETOF record
+    LANGUAGE INTERNAL
+    SET search_path = pg_catalog
```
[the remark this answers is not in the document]
svenklemm: inlining only applies to sql functions where the body of the sql function gets injected into the actual query. language internal is a C function so it will never get inlined. I dont see the harm in setting search_path explicitly but i'm not sure how much value it actually has for C functions.

migration/idempotent/010-telemetry.sql
JamesGuthrie: Yes, thanks for the catch.

migration/idempotent/001-base.sql
```diff
@@ -342,10 +360,7 @@ BEGIN
    RETURN NEW;
 END
 $func$
-LANGUAGE PLPGSQL VOLATILE
-SECURITY DEFINER
---search path must be set for security definer
-SET search_path = pg_temp;
+LANGUAGE PLPGSQL VOLATILE;
```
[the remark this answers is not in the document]
JamesGuthrie: done

migration/idempotent/001-base.sql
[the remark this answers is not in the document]
JamesGuthrie: done

migration/idempotent/001-base.sql
```diff
@@ -658,25 +676,29 @@ GRANT EXECUTE ON FUNCTION _prom_catalog.get_new_label_id(text, text) to prom_wri
...
 AS $function$jsonb_each_text$function$;
 GRANT EXECUTE ON FUNCTION _prom_catalog.label_jsonb_each_text(jsonb) to prom_reader;
 
 --wrapper around unnest to give better row estimate (10 not 100)
 CREATE OR REPLACE FUNCTION _prom_catalog.label_unnest(label_array anyarray)
- RETURNS SETOF anyelement
- LANGUAGE INTERNAL
- IMMUTABLE PARALLEL SAFE STRICT ROWS 10
+    RETURNS SETOF anyelement
+    LANGUAGE INTERNAL
```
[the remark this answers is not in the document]
JamesGuthrie: done

migration/idempotent/001-base.sql
```diff
@@ -1036,7 +1077,9 @@ GRANT EXECUTE ON FUNCTION prom_api.jsonb(prom_api.label_array) TO prom_reader;
 
 --Returns the label_array given a series_id
 CREATE OR REPLACE FUNCTION prom_api.labels(series_id BIGINT)
-RETURNS prom_api.label_array AS $$
+    RETURNS prom_api.label_array
+    SET search_path = pg_catalog
```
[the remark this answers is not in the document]
JamesGuthrie: I looked into this doc on inlining when I was trying to figure out what could/should be inlined: https://wiki.postgresql.org/wiki/Inlining_of_SQL_functions.

According to that, for functions used in the scalar position (e.g. `SELECT labels(some_id);`), inlining will not occur if the body contains a `FROM`:
> the body contains no aggregate or window function calls, no subqueries, no CTEs, no FROM clause or references to any table or table-like object, none of GROUP BY, HAVING, ORDER BY, DISTINCT, LIMIT, OFFSET, UNION, INTERSECT, EXCEPT

It seems as though for functions used in the table position (e.g. `SELECT * FROM labels(some_id);`), that restriction is lifted, but there is the following requirement:
> the function is declared RETURNS SETOF or RETURNS TABLE

Honestly I don't know enough to be able to say definitively, but it doesn't seem like this can be inlined.

migration/idempotent/001-base.sql
JamesGuthrie: done

migration/idempotent/001-base.sql
JamesGuthrie: Same as above, I'm not sure that this is inlinable in its current form.

migration/idempotent/007-tracing-tags.sql
JamesGuthrie: oh man, that was an error. good catch.

migration/idempotent/007-tracing-tags.sql
JamesGuthrie: Actually I first schema-qualified everything in this fn, then decided to set the search path instead, so the other schema qualifications are superfluous.

migration/idempotent/007-tracing-tags.sql
JamesGuthrie: same as above

migration/idempotent/001-base.sql
```diff
@@ -207,6 +222,7 @@ GRANT EXECUTE ON FUNCTION _prom_catalog.attach_series_partition(_prom_catalog.me
 --lock-order: metric table, data_table, series parent, series partition
 
 CREATE OR REPLACE PROCEDURE _prom_catalog.finalize_metric_creation()
+-- Note: Cannot SET search_path because we do transaction control
```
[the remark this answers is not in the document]
JamesGuthrie: So there were a couple of reasons that I left off `SET LOCAL`. The first was that I wasn't sure how it works. The second was that it seems to be unnecessary, as long as we are schema-qualifying things in the function. On that note, I hope to be able to add a pgspot check step soon, which gives us more safety w.r.t. schema-qualifying.

In the meantime, I have gathered more knowledge on how `SET LOCAL` works, and I'm not sure that it really helps in this scenario. `SET LOCAL` doesn't have any "function" magic, and relies on postgres' transaction control to work. From the docs:

> The effects of SET LOCAL last only till the end of the current transaction, whether committed or not.

This means that we would still need to do the "save and restore old search_path" dance. We can do it, but I'd prefer not to, if we don't have to.

migration/idempotent/001-base.sql
JamesGuthrie: same answer as above

migration/idempotent/011-maintenance.sql
JamesGuthrie: same as above

migration/migration-table.sql
JamesGuthrie: Same answer as for the transaction control. I don't think that `SET LOCAL` works that way.

migration/migration/004-schemas.sql
JamesGuthrie: Good question. I'm actually not sure that we should be removing it for all schemas, but the basic problem that I'm trying to work around here is:

- `execute_everywhere` runs both on the local node and on all data nodes
- some of the schemas created in the execute everywhere block have already been created on the local node during extension install (e.g. `_prom_ext`, and `_ps_catalog`
- removing `IF NOT EXISTS` breaks the extension install

We talked about how we might install the extension on data nodes as well, in which case we wouldn't want to have these schemas created in an execute everywhere, so I thought that I would remove the execute everywhere for now.

migration/migration/004-schemas.sql
JamesGuthrie: line 2 is `_prom_catalog`, this is `_ps_catalog`. We need fewer schemas.

templates/idempotent-wrapper.sql
JamesGuthrie: as above

templates/migration-wrapper.sql
JamesGuthrie: as above

migration/migration/004-schemas.sql
JamesGuthrie: done

migration/idempotent/008-tracing-functions.sql
```diff
@@ -492,12 +495,13 @@ BEGIN
     RETURN _operation_id;
 END;
 $func$
-LANGUAGE plpgsql VOLATILE STRICT;
+LANGUAGE plpgsql STRICT;
```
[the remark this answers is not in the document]
JamesGuthrie: done

jgpruitt:
was `CREATE OR REPLACE PROCEDURE _prom_catalog.compress_old_chunks(metric_table TEXT, compress_before TIMESTAMPTZ)` skipped on purpose? No search_path changes in that one.

migration/idempotent/001-base.sql
```diff
@@ -3026,45 +3142,49 @@ BEGIN
...
         PERFORM _prom_catalog.unlock_metric_for_maintenance(r.id);
 
         COMMIT;
+        -- reset search path after transaction end
+        SET LOCAL search_path = pg_catalog;
     END LOOP;
 
     FOR r IN
         SELECT *
         FROM unnest(remaining_metrics)
```
jgpruitt, line 3166: schema qualify unnest?

migration/migration/004-schemas.sql
```diff
@@ -47,7 +51,7 @@ DO $$
 DECLARE
    new_path text;
 BEGIN
-   new_path := current_setting('search_path') || format(',%L,%L,%L,%L,%L,%L', 'ps_tag', '_prom_ext', 'prom_api', 'prom_metric', '_prom_catalog', 'ps_trace');
+   new_path := format('%L,%L,%L,%L,%L,%L,%L,%L', '$user','public', 'ps_tag', '_prom_ext', 'prom_api', 'prom_metric', '_prom_catalog', 'ps_trace');
```
jgpruitt, line 54: `format` is not strictly necessary here anymore, no? It's just one big literal.

migration/migration/007-matcher-operators.sql
```diff
@@ -20,20 +19,22 @@ AS $function$jsonb_each_text$function$;
...
 LANGUAGE SQL IMMUTABLE PARALLEL SAFE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.count_jsonb_keys(jsonb) TO prom_reader;
 
 CREATE OR REPLACE FUNCTION prom_api.matcher(labels jsonb)
 RETURNS prom_api.matcher_positive
+-- Note: no explicit search_path because we want inlining
 AS $func$
     SELECT ARRAY(
            SELECT coalesce(l.id, -1) -- -1 indicates no such label
            FROM _prom_catalog.label_jsonb_each_text(labels-'__name__') e
```
jgpruitt, line 35: schema qualify the jsonb `-` operator

migration/migration/012-tracing.sql
```diff
@@ -253,12 +252,6 @@ BEGIN
 
     IF _is_timescaledb_installed THEN
         IF _is_multinode THEN
-            --need to clear the search path while creating distributed
-            --hypertables because otherwise the datanodes don't find
-            --the right column types since type names are not schema
-            --qualified if in search path.
-            _saved_search_path := current_setting('search_path');
```
jgpruitt, line 260: why was this removed?

migration/idempotent/013-apply-config.sql
```diff
@@ -1,13 +1,8 @@
 
 DO $block$
 DECLARE
-    _old_search_path text;
     _rec record;
 BEGIN
-    -- set the search_path such that pg_get_function_identity_arguments
-    -- should return fully schema qualified arguments
-    EXECUTE 'SHOW search_path' INTO STRICT _old_search_path;
-    SET search_path TO pg_temp;
```
jgpruitt, line 10: why was this removed?

migration/migration/015-tracing-compression.sql
```diff
@@ -277,12 +277,6 @@ BEGIN
 
     IF _is_timescaledb_installed THEN
         IF _is_multinode THEN
-            --need to clear the search path while creating distributed
-            --hypertables because otherwise the datanodes don't find
-            --the right column types since type names are not schema
-            --qualified if in search path.
-            _saved_search_path := current_setting('search_path');
-            SET search_path = pg_temp;
```
jgpruitt, line 285: why was this removed?

jgpruitt (changes requested):

migration/idempotent/011-maintenance.sql
```diff
@@ -458,6 +490,10 @@ AS $$
 DECLARE
    startT TIMESTAMPTZ;
 BEGIN
+    -- Note: We cannot use SET in the procedure declaration because we do transaction control
```
JamesGuthrie, line 493: I don't understand this one. Where are we doing transaction control in the body of this procedure? Does it apply transitively to other procedures called from this one?

migration/idempotent/001-base.sql
JamesGuthrie: As discussed on slack, it's not necessary to set `search_path` and schema qualify. For readability I've explicitly done either one or the other.

migration/idempotent/003-matcher-functions.sql
```diff
@@ -1,28 +1,31 @@
...
 LANGUAGE SQL STABLE PARALLEL SAFE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.label_value_contains(prom_api.label_value_array, TEXT) TO prom_reader;
 
 CREATE OR REPLACE FUNCTION prom_api.matcher(labels jsonb)
 RETURNS prom_api.matcher_positive
+-- Note: no explicit `SET SCHEMA` because we want this function to be inlined
 AS $func$
     SELECT ARRAY(
            SELECT coalesce(l.id, -1) -- -1 indicates no such label
            FROM _prom_catalog.label_jsonb_each_text(labels-'__name__') e
```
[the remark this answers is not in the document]
JamesGuthrie: Awesome catch. Looks like a bug in pgspot.

migration/idempotent/013-apply-config.sql
JamesGuthrie: I should have added the justification behind this to the readme. In the migration and idempotent wrappers we wrap the script with:

```SQL
    EXECUTE 'SHOW search_path' INTO STRICT _old_search_path;
    SET search_path TO pg_catalog;

-- Note: this weird indentation is important. We compare SQL across upgrade paths,
-- and the comparison is indentation-sensitive.
{{body}}

    EXECUTE format('SET search_path TO %s', _old_search_path);
```

Which sets the search_path for the whole block, in the same manner that this mechanism does.

migration/migration/015-tracing-compression.sql
JamesGuthrie: same as above

migration/migration/012-tracing.sql
JamesGuthrie: same as above

JamesGuthrie (commented):
I have a couple of questions, hopefully GitHub hasn't lost them 😅

migration/idempotent/011-maintenance.sql
cevian: yes it applies transitively

migration/migration/004-schemas.sql
cevian: I think format is more safe and future-proof so I'd rather leave. This code will go away in 166 anyway

migration/idempotent/001-base.sql
```diff
@@ -2950,10 +2950,11 @@ BEGIN
...
     BEGIN
+        --note search_path cannot be set here because this is a proc that may be executed transactionally so schema qualify everything
         SELECT h.* INTO STRICT hypertable_row FROM _timescaledb_catalog.hypertable h
-        WHERE table_name = metric_table AND schema_name = 'prom_data';
+        WHERE h.table_name OPERATOR(pg_catalog.=) metric_table AND h.schema_name OPERATOR(pg_catalog.=) 'prom_data';
 
-        SELECT d.* INTO STRICT dimension_row FROM _timescaledb_catalog.dimension d WHERE hypertable_id = hypertable_row.id ORDER BY id LIMIT 1;
+        SELECT d.* INTO STRICT dimension_row FROM _timescaledb_catalog.dimension d WHERE d.hypertable_id OPERATOR(pg_catalog.=) hypertable_row.id ORDER BY d.id LIMIT 1;
 
         IF min_time = timestamptz '-Infinity' THEN
```
jgpruitt, line 2959: schema qualify `timestamptz` in the cast

jgpruitt (changes requested):

jgpruitt (approved):

cevian (approved):

cevian:
Connector side tests passing https://github.com/timescale/promscale/actions/runs/2169200395

PR #181, opened 2022-04-12
Dump/Restore Test

jgpruitt:


e2e/tests/dump-restore-test.rs
```diff
@@ -0,0 +1,304 @@
...
+        exit.success(),
+        "copying the dump into the container failed: {}",
+        exit
+    );
+    let exit = Command::new("docker")
+        .arg("exec")
+        .arg(container_id)
+        .arg("psql")
+        .arg("-U")
+        .arg(username)
```
cevian, line 233: Nit: not sure if rustfmt allows this but would be nice to group args like this.
```suggestion
        .arg("-U").arg(username)
```

e2e/tests/dump-restore-test.rs
```diff
@@ -0,0 +1,304 @@
...
+    let dump = dump_db(container_id, "db0", "postgres"); // todo: dump using jack user
+    container.stop();
+    container.rm();
+
+    println!("dump file at {}", dump.display());
+
+    let container = run_postgres(&docker, "db1", "postgres");
+    let container_id = container.id();
+    psql_cmd(
+        container_id,
```
cevian, line 280: nit: are we sure it's better to pass in container ids rather than containers into these functions?

cevian (approved):
LGTM. Some nits, nothing major

e2e/tests/dump-restore-test.rs
jgpruitt: I had the same thought and tried it, but it complained :(

e2e/tests/dump-restore-test.rs
jgpruitt: no, not sure. the functions only need the id, but maybe the typing associated with passing the container itself would be nice? I'll give it a try...

jgpruitt: okay, that works well enough. changed it.

e2e/tests/dump-restore-test.rs
jgpruitt: oh, but  I can do this: `.args(["-v", "ON_ERROR_STOP=1"])` :fire:

e2e/tests/dump-restore-test.rs
```diff
@@ -0,0 +1,304 @@
...
+        .spawn()
+        .unwrap()
+        .wait()
+        .unwrap();
+    assert!(
+        exit.success(),
+        "copying the dump out of the container failed: {}",
+        exit
+    );
+    path
```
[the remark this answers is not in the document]
jgpruitt: yup. on it.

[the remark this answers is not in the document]
jgpruitt: done.

PR #182, opened 2022-04-12
Ensure that _prom_ext schema created safely

JamesGuthrie:
Because the extension was declared as non-relocatable, and in the schema
`_prom_ext`, it was possible to have that schema (and objects in it)
created before the extension install script is run. This opened the
extension up to precreation attacks.

This change takes a number of measures to mitigate that.

- We configure the schema to `public`, which allows the extension to
  control the creation of the `_prom_ext` schema.
- We add a `bootstrap` migration script which is run before the pgx
  generated SQL, which creates the `_prom_ext` schema.
- We tell pgx to put all generated objects in the `_prom_ext` schema.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -29,6 +29,7 @@ struct ControlFile {
 
 const CONTROL_FILE_NAME: &str = "promscale.control";
 const MIGRATION_FILE_NAME: &str = "hand-written-migration.sql";
+const BOOTSTRAP_FILE_NAME: &str = "bootstrap.sql";
 
 fn main() {
     println!("cargo:rerun-if-changed=Cargo.lock");
@@ -37,6 +38,7 @@ fn main() {
     println!("cargo:rerun-if-changed=migration/migration");
     println!("cargo:rerun-if-changed=migration/idempotent");
     println!("cargo:rerun-if-changed=templates");
+    println!("cargo:rerun-if-changed=migration/bootstrap");
     // According to the cargo documentation we are horribly misusing
     // the build script mechanism, hence this workaround:
     // "Build scripts may save any output files or intermediate artifacts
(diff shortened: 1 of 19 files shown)


sumerman (approved):
Almost LGTM with a couple of caveats:
- I only visually inspected SQL changes trying to assess whether changes fall into one of the few common patterns. But I didn't read through every single change.
- I assume that Rust code changes outside of `src/schema` are limited to adding `#[pg_schema] mod _prom_ext`, fixing tests and reformatting.

build.rs
```diff
@@ -33,17 +33,32 @@ fn main() {
...
+    let mut out_path = PathBuf::from(manifest_dir.clone());
     out_path.push(MIGRATION_FILE_NAME);
     let mut file = File::create(out_path).unwrap();
-    file.write_all(sql.as_bytes())
+    file.write_all(rendered_sql.as_bytes())
         .expect("unable to write file contents");
+
+    let mut bootstrap_path = PathBuf::from(manifest_dir);
+    bootstrap_path.push("bootstrap.sql");
+    let mut migration_sql = include_str!("migration/migration-table.sql").to_string();
```
cevian, line 50: should we move these files into
- migration/bootstrap/migration-table.sql
- migration/bootstrap/schema.sql
?

cevian (approved):
This review is only of the last commit. LGTM, some nits below

cevian (approved):

sql/promscale--0.0.0.sql
```diff
@@ -435,6 +435,7 @@ BEGIN
     -- Bring migrations table up to speed
     INSERT INTO _ps_catalog.migration (name, applied_at_version)
     VALUES
+        ('create_ext_schema.sql'          , '0.0.0'),
```
jgpruitt, line 438: the fact that this isn't 000_ makes my eye twitch, but I'll get over it :)

build.rs
```diff
@@ -76,15 +78,33 @@ fn generate_control_file(manifest_dir: &str) {
...
     out_path.push(MIGRATION_FILE_NAME);
     let mut file = File::create(out_path).unwrap();
-    file.write_all(sql.as_bytes())
-        .expect("unable to write migration file contents");
+    file.write_all(rendered_sql.as_bytes())
+        .expect("unable to write file contents");
+
+    let mut bootstrap_out_path = PathBuf::from(manifest_dir);
+    bootstrap_out_path.push(BOOTSTRAP_FILE_NAME);
+    let mut bootstrap_out_file = File::create(bootstrap_out_path).unwrap();
```
jgpruitt, line 91: Add this path to .gitignore, please?

jgpruitt (changes requested):
only 1 nit

build.rs
cevian: done

jgpruitt (approved):

migration/bootstrap.sql
```diff
@@ -0,0 +1 @@
+CREATE SCHEMA _prom_ext;
```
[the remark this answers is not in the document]
JamesGuthrie: done

build.rs
JamesGuthrie: done

PR #183, opened 2022-04-12
Bump timescaledb version to 2.6.1

JamesGuthrie:


jgpruitt (approved):
Do we need to bump the version in the tests on the connector side or is that handled automatically?

cevian (approved):

PR #185, opened 2022-04-13
Dump/restore fixes

jgpruitt:
# Test the dump/restore process

Added a test to ensure that the dump/restore process works.

To test this PR, run `make docker-image-14`, then run this command:

```
TS_DOCKER_IMAGE=local/dev_promscale_extension:head-ts2-pg14 cargo test --package e2e --test dump-restore-test dump_restore_test -- --exact
```

There is [a corresponding PR](https://github.com/timescale/promscale/pull/1311) on the promscale connector repo.

The rough outline of the test:

1. create a Postgres docker container with timescaledb
2. create the promscale extension
3. add metrics, exemplars, and traces to the database
4. snapshot the database
5. use pg_dump to create a logical SQL dump
6. destroy the docker container and create a new one
7. restore into the new database
8. snapshot the database
9. create more data to make sure things are still wired up correctly
10. compare the two snapshots. they should be equal

This uncovered several issues that were addressed:

The _ps_catalog.migration and _ps_catalog.promscale_instance_information tables should not be extension config tables. This means the structure of the tables will be dumped by pg_dump, but the data in them will not. The extension code will populate these tables appropriately. Because we are handling the _ps_catalog.migration table this way, you must use the same version of the extension on the restore side that was present on the dump side.

The extension inserts an initial row into the _prom_catalog.ids_epoch table on installation. We need to skip this if a restore is in progress so that the dumped value is what is inserted.

The public.prom_installation_info table was unused and caused problems for restores, so it was removed.

Both the _prom_catalog.remote_commands and _ps_trace.tag_key tables have rows managed by the extension and potentially user-supplied rows. For these tables, we manipulate the sequence such that ids < 1000 are managed by the extension and NOT dumped by pg_dump. ids >= 1000 are dumped and restored.

The `_prom_catalog.default` table contains default configuration values. End-users can update these values. The extension inserts the initial defaults during `create extension` and subsequent migrations. We want pg_dump to dump any user-modified values so that they are preserved when restored. We introduced a new view to provide the initial values, and the table's semantics have changed such that they only contain user-modified values. All access to the values now goes through getter/setter functions to handle the merging of the two.

The cascade option is required for dropping the extension. Our new custom type for trace_id is present in the span, link, and event hypertables. This means the custom type is present in the associated compressed tables and chunks. Dropping the extension drops the custom type and the span, link, and event tables, but does not attempt to drop the compressed tables and chunks. Thus, the drop fails because we cannot drop the trace_id type when the compressed tables and chunk definitions depend on it. The cascade option fixes this. This is probably an area for improvement in timescaledb.

The span, link, and event tables are hypertables if timescaledb is installed. The promscale extension creates these tables and various functions and views that depend on these tables. We need the associated backing compressed tables and chunks to be dumped and restored by pg_dump to preserve the data, but we need the tables themselves to be defined as a part of `create extension` so that the dependent objects can be created. Therefore, the span, link, and event tables belong to the extension, but the backing compressed tables and chunks do not. During a restore, the span, link, and event tables are created, but we do not call create_hypertable on them (we can't). We added a promscale_post_restore function which wires up the triggers to span, link, and event tables so that subsequent writes are redirected to the appropriate backing tables.

By default, triggers are added to both the parent table AND child tables. We don't want the child tables (chunk tables) to have the trigger, because that would block the insertion of data. So, in `promscale_post_restore` we add a step to drop the triggers from the chunk tables.


migration/migration/008-install-uda.sql
```diff
@@ -56,10 +56,22 @@ $func$
 LANGUAGE sql STABLE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.is_multinode() TO prom_reader;
 
+CREATE OR REPLACE FUNCTION _prom_catalog.is_restore_in_progress()
```
JamesGuthrie, line 59: If you create this here, then you will also need to create it in the 0.0.0 script. We need to be careful about the changes that we make the existing `migration/migration` scripts.

e2e/tests/dump-restore-test.rs
```diff
@@ -1,38 +1,65 @@
...
+    }
 
     docker.run(generic_postgres)
 }
 
-fn psql_script(container: &Container<Cli, GenericImage>, db: &str, username: &str, path: &str) {
-    println!("executing psql script {}...", path);
+/// Runs a SQL script file in the docker container
+///
+/// The script file is copied into the container, and executed with
```
JamesGuthrie, line 51: This says that the script file is copied into the container, but I don't see that actually happening in the implementation of `psql_file`.

e2e/tests/dump-restore-test.rs
```diff
@@ -140,13 +230,21 @@ fn snapshot_db(container: &Container<Cli, GenericImage>, db: &str, username: &st
         exit
     );
     let src = PathBuf::from("/tmp/snapshot.txt");
-    let dest = env::temp_dir().join(format!("snapshot-{}.txt", db));
-    copy_out(container, &src, &dest);
-    println!("database snapshot at: {}", dest.display());
-    fs::read_to_string(dest.as_path()).unwrap()
+    copy_out(container, &src, path);
```
JamesGuthrie, line 233: Remove the snapshot file before returning?

e2e/tests/dump-restore-test.rs
```diff
@@ -155,21 +253,21 @@ fn dump_db(container: &Container<Cli, GenericImage>, db: &str, username: &str) -
...
-        .args(["-T", "public.prom_installation_info"]) // todo: ???
-        .args(["-N", "_timescaledb_internal"])
-        .args(["-N", "_timescaledb_cache"])
-        .args(["-N", "_timescaledb_catalog"])
-        .args(["-N", "_timescaledb_config"])
-        .args(["-N", "_timescaledb_internal"])
-        .args(["-N", "timescaledb_experimental"])
-        .args(["-N", "timescaledb_information"])
-        .args(["-f", "/tmp/dump.sql"])
+        //.args(["-T", "_prom_catalog.\"default\""]) // todo: wrap in functions and only insert if not default value
```
JamesGuthrie, line 256: Do we need to keep these commented-out things around?

e2e/tests/dump-restore-test.rs
```diff
@@ -181,61 +279,116 @@ fn dump_db(container: &Container<Cli, GenericImage>, db: &str, username: &str) -
...
+/// 2. create metrics, series, samples, exemplars, traces in the database
+/// 3. snapshot the database
+/// 4. use pg_dump to create a logical sql dump
+/// 5. destroy the docker container and create a new one
+/// 6. restore into the new database
+/// 7. snapshot the database
+/// 8. compare the two snapshots. they should be equal
+/// 9. compare one of the snapshots to a known good snapshot. they should be equal
 #[test]
 fn dump_restore_test() {
```
JamesGuthrie, line 312: This function is not very readable. Can be improved later.

JamesGuthrie (changes requested):

JamesGuthrie:
This change can be rebased against the most recent develop, the changes that it's based on were merged.

migration/idempotent/001-base.sql
```diff
@@ -1,16 +1,80 @@
 --NOTES
 --This code assumes that table names can only be 63 chars long
 
+CREATE OR REPLACE VIEW _prom_catalog.default_default AS
```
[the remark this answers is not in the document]
jgpruitt: thank you

e2e/tests/dump-restore-test.rs
jgpruitt: I was leaving the comments here to remind me what I did. I'll fix up the commit messages accordingly and then nuke these comments.

e2e/tests/dump-restore-test.rs
jgpruitt: I'm not deleting the files on purpose for the sake of debugging. You can open the two snapshot files in a proper diff tool if the test fails, which is super helpful

e2e/tests/test.rs
```diff
@@ -14,7 +14,9 @@ fn create_drop_promscale_extension() {
     let result = client.simple_query("CREATE EXTENSION promscale;").unwrap();
     assert_eq!(result.len(), 1);
 
-    let result = client.simple_query("DROP EXTENSION promscale;").unwrap();
+    let result = client
+        .simple_query("DROP EXTENSION promscale CASCADE;")
```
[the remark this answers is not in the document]
jgpruitt: Added an explanation in the PR description. It has to do with our custom type for trace_id and the backing compressed tables and chunks for the span, link, and event hypertables.

migration/idempotent/001-base.sql
```diff
@@ -4,13 +4,22 @@
...
 BEGIN
     IF command_key IS NOT NULL THEN
        INSERT INTO _prom_catalog.remote_commands(key, command, transactional) VALUES(command_key, command, transactional)
        ON CONFLICT (key) DO UPDATE SET command = excluded.command, transactional = excluded.transactional;
     END IF;
 
     EXECUTE command;
+
+    -- do not call distributed_exec if we are in the middle of restoring from backup
+    _is_restore_in_progress = coalesce((SELECT setting = 'on' from pg_catalog.pg_settings where name = 'timescaledb.restoring'), false);
```
cevian, line 73: is cast to bool better?
`_is_restore_in_progress = coalesce((SELECT setting::boolean from pg_catalog.pg_settings where name = 'timescaledb.restoring'), false);`

migration/idempotent/011-maintenance.sql
```diff
@@ -604,4 +601,32 @@ LANGUAGE PLPGSQL;
 REVOKE ALL ON FUNCTION prom_api.config_maintenance_jobs(int, interval, jsonb) FROM PUBLIC;
 GRANT EXECUTE ON FUNCTION prom_api.config_maintenance_jobs(int, interval, jsonb) TO prom_admin;
 COMMENT ON FUNCTION prom_api.config_maintenance_jobs(int, interval, jsonb)
-IS 'Configure the number of maintence jobs run by the job scheduler, as well as their scheduled interval';
+IS 'Configure the number of maintenance jobs run by the job scheduler, as well as their scheduled interval';
+
+CREATE OR REPLACE FUNCTION public.promscale_post_restore()
+RETURNS void
+--security definer to add jobs as the logged-in user
```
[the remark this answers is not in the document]
cevian: no it is not...

migration/idempotent/013-apply-config.sql
```diff
@@ -81,6 +88,12 @@ BEGIN
         AND d.deptype = 'e'
         AND e.extname = 'promscale'
         AND k.relkind IN ('r', 'p')
+        AND (n.nspname, k.relname) NOT IN
+        (
+            -- these should NOT be config tables
+            ('_ps_catalog'::name, 'migration'::name),
```
cevian, line 94: add 
```
--migration table will be populated by the installation of extension
```
and
```
--we want the telemetry data reset with restores
```

migration/migration/002-utils.sql
```diff
@@ -11,14 +11,24 @@ CREATE TABLE _prom_catalog.remote_commands(
...
 BEGIN
     IF command_key IS NOT NULL THEN
        INSERT INTO _prom_catalog.remote_commands(key, command, transactional) VALUES(command_key, command, transactional)
        ON CONFLICT (key) DO UPDATE SET command = excluded.command, transactional = excluded.transactional;
     END IF;
 
     EXECUTE command;
+
+    -- do not call distributed_exec if we are in the middle of restoring from backup
+    _is_restore_in_progress = coalesce((SELECT setting = 'on' from pg_catalog.pg_settings where name = 'timescaledb.restoring'), false);
```
cevian, line 27: actually this query appears for the second time, should we make a function?

migration/migration/015-tracing-compression.sql
```diff
@@ -261,82 +261,135 @@ GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.link TO prom_writer;
...
+        '1 hour'::interval
+    );
+    PERFORM public.set_chunk_time_interval(
+        '_ps_trace.link'::regclass,
+        '1 hour'::interval
+    );
+
+    _is_compression_available = NOT (current_setting('timescaledb.license') = 'apache');
+    IF _is_compression_available THEN
+        -- drop any compressed chunks for the three tables we are modifying
```
cevian, line 342: I think this will go away after rebase; It should be around, we are truncating the tables on top now.

migration/migration/015-tracing-compression.sql
```diff
@@ -261,82 +261,135 @@ GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.link TO prom_writer;
...
-            -- turn on compression
-            ALTER TABLE _ps_trace.span SET (timescaledb.compress, timescaledb.compress_orderby='trace_id,span_id,start_time');
-            ALTER TABLE _ps_trace.event SET (timescaledb.compress, timescaledb.compress_orderby='trace_id,span_id,time');
-            ALTER TABLE _ps_trace.link SET (timescaledb.compress, timescaledb.compress_orderby='trace_id,span_id,span_start_time');
-
-            PERFORM public.add_compression_policy('_ps_trace.span', INTERVAL '1 hours');
-            PERFORM public.add_compression_policy('_ps_trace.event', INTERVAL '1 hours');
-            PERFORM public.add_compression_policy('_ps_trace.link', INTERVAL '1 hours');
-        END IF;
+    PERFORM public.set_chunk_time_interval(
```
cevian, line 327: I'm not sure why this was added?

cevian (changes requested):
Partial review. Will do another full review after rebase

e2e/scripts/snapshot.sql
```diff
@@ -1,4 +1,132 @@
...
+\do prom_api.*
+\do prom_data.*
+\do prom_data_exemplar.*
+\do prom_data_series.*
+\do prom_info.*
+\do prom_metric.*
+\do prom_series.*
+\do ps_tag.*
+\do ps_trace.*
+
```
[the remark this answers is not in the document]
jgpruitt: good idea. done

e2e/scripts/snapshot.sql
```diff
@@ -1,4 +1,132 @@
...
+\drds
+\du
 \dx
-\dx+ promscale
 \dn
--- todo: copy the snapshotting code from the connector
+\dx+ promscale
+
+-- tables
+\d+ public.*
```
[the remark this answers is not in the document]
jgpruitt: is this good enough? 1cdbfed0cda3e559ec32d3792149b2c507567b47

e2e/scripts/snapshot.sql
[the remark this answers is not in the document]
jgpruitt: done in 1cdbfed0cda3e559ec32d3792149b2c507567b47

e2e/tests/dump-restore-test.rs
jgpruitt: removed now

e2e/tests/snapshots/golden_tests__golden_test_testdata_defaults_sql.snap
```diff
@@ -0,0 +1,93 @@
...
+expression: query_result
+---
+\set ON_ERROR_STOP 1
+create extension promscale;
+CREATE EXTENSION
+\echo make sure the view returns what we expect it to
+make sure the view returns what we expect it to
+select key, value
+from _prom_catalog.default_default dd
+union
```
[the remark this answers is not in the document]
jgpruitt: fixed in 59b575c7a49fed94240ce92545fc7309f69e9cbe

migration/idempotent/013-apply-config.sql
jgpruitt: added in db5c7579dd6c0d87d48cf6fd888ede3cd7ce2a67

e2e/tests/dump-restore-test.rs
jgpruitt: fixed in 834df2de0c02767260f924e0766a29384e996f39

migration/idempotent/011-maintenance.sql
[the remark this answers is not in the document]
jgpruitt: removed in 5dfb5cd8052c928c0a0dff0876e1f234f42f8312

migration/idempotent/012-remote-commands.sql
```diff
@@ -31,8 +31,8 @@ FROM
     SELECT key, seq
     FROM x
     UNION
-    -- any other remote commands get listed afterwards
-    SELECT key, (SELECT max(seq) FROM x) + row_number() OVER (ORDER BY seq)
+    -- any other remote commands get listed afterwards starting with 1000
```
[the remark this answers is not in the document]
jgpruitt: I don't think we have to since this is an idempotent script. As soon as you upgrade from 0.0.0 to 0.5.0 it will be run and do the reshuffle.

migration/migration/008-install-uda.sql
jgpruitt: good catch

migration/migration/015-tracing-compression.sql
jgpruitt: I think I got confused during a rebase 🤦🏻‍♂️

migration/migration/008-install-uda.sql
jgpruitt: removed it from here and left it only in idempotent script

migration/idempotent/001-base.sql
jgpruitt: huh. I didn't realize 'on' would cast to true. wow

migration/migration/002-utils.sql
jgpruitt: i made it a function in idempotent/001-base.sql however unfortunately, the migration scripts run BEFORE the idempotent scripts so we cannot guarantee that it will exist at this point in the code

e2e/tests/dump-restore-test.rs
jgpruitt: agreed. I have iterated on it a bit, but I'm not sure how much better it is

migration/idempotent/012-remote-commands.sql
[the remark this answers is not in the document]
JamesGuthrie: Ah, I see. I misunderstood how this works when I read it.

migration/idempotent/001-base.sql
```diff
@@ -1,6 +1,61 @@
...
+-- the "default defaults". We don't need to keep values in the default table
+-- if they are the same as the default defaults
+DELETE FROM _prom_catalog.default x
+WHERE x.key IN
+(
+    SELECT d.key
+    FROM _prom_catalog.default d
+    LEFT OUTER JOIN _prom_catalog.default_default dd ON (d.key = dd.key)
+    WHERE d.value is not distinct from dd.value
+);
```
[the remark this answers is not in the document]
jgpruitt: my thinking was that as we change the initial defaults in the view, we'd always remove any "duplicate" entries in the table. But I think I like your thoughts better.

migration/idempotent/001-base.sql
```diff
@@ -1,6 +1,61 @@
...
+CREATE OR REPLACE FUNCTION _prom_catalog.set_default_value(_key text, _value text)
+    RETURNS VOID
+    SET search_path = pg_catalog
+AS $func$
+    -- remove the existing user-supplied default value if one exists
+    DELETE FROM _prom_catalog.default d WHERE d.key = _key;
+    -- insert the user-supplied value into the default table, unless the value is the same as the default default value
+    INSERT INTO _prom_catalog.default (key, value)
+    SELECT _key, _value
+    WHERE (_key, _value) NOT IN (SELECT dd.key, dd.value FROM _prom_catalog.default_default dd)
```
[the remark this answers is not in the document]
jgpruitt: good point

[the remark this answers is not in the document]
jgpruitt: resolved in https://github.com/timescale/promscale_extension/pull/185/commits/71faee03b73f1490aeb79f5eddce766a03c16285

migration/idempotent/001-base.sql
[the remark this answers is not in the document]
jgpruitt: resolved in https://github.com/timescale/promscale_extension/pull/185/commits/71faee03b73f1490aeb79f5eddce766a03c16285

e2e/scripts/snapshot.sql
```diff
@@ -1,4 +1,77 @@
...
+\g (tuples_only=on format=csv) describe.sql
+\i describe.sql
+
+-- snapshot the data from all the tables
+select
+    format($$select '%I.%I' as table_snapshot;$$, n.nspname, k.relname),
+    case (n.nspname, k.relname)
+        -- we don't care about comparing the applied_at_version and applied_at columns of the migration table
+        when ('_ps_catalog'::name, 'migration'::name) then 'select name, body from _ps_catalog.migration;'
+        else format('select * from %I.%I;', n.nspname, k.relname)
```
cevian, line 55: I suggest fixing the order
```suggestion
        else format('select * from %I.%I tbl order by tbl;', n.nspname, k.relname)
```

e2e/scripts/snapshot.sql
```diff
@@ -1,4 +1,77 @@
...
+order by c.c, s.s
+\g (tuples_only=on format=csv) describe.sql
+\i describe.sql
+
+-- snapshot the data from all the tables
+select
+    format($$select '%I.%I' as table_snapshot;$$, n.nspname, k.relname),
+    case (n.nspname, k.relname)
+        -- we don't care about comparing the applied_at_version and applied_at columns of the migration table
+        when ('_ps_catalog'::name, 'migration'::name) then 'select name, body from _ps_catalog.migration;'
```
cevian, line 54:
```suggestion
        when ('_ps_catalog'::name, 'migration'::name) then 'select name, body from _ps_catalog.migration order by name, body;'
```

e2e/tests/dump-restore-test.rs
```diff
@@ -113,7 +190,35 @@ fn copy_out(container: &Container<Cli, GenericImage>, src: &PathBuf, dest: &Path
     );
 }
 
-fn snapshot_db(container: &Container<Cli, GenericImage>, db: &str, username: &str) -> String {
+/// Edits a snapshot file to account for acceptable differences
+fn edit_snapshot(path: &Path) -> String {
+    let snapshot = fs::read_to_string(path).expect("failed to read snapshot file");
+
+    let re = Regex::new(r"Partition constraint: satisfies_hash_partition\('\d+'::oid").unwrap();
```
cevian, line 197:
```suggestion
    // partition constraints are printed with the OID of the table, but OIDs are allowed to change
    let re = Regex::new(r"Partition constraint: satisfies_hash_partition\('\d+'::oid").unwrap();
```

e2e/tests/dump-restore-test.rs
```diff
@@ -113,7 +190,35 @@ fn copy_out(container: &Container<Cli, GenericImage>, src: &PathBuf, dest: &Path
     );
 }
 
-fn snapshot_db(container: &Container<Cli, GenericImage>, db: &str, username: &str) -> String {
+/// Edits a snapshot file to account for acceptable differences
+fn edit_snapshot(path: &Path) -> String {
```
cevian, line 194: nit: better name?
```suggestion
fn normalize_snapshot(path: &Path) -> String {
```

e2e/tests/dump-restore-test.rs
```diff
@@ -181,61 +277,151 @@ fn dump_db(container: &Container<Cli, GenericImage>, db: &str, username: &str) -
...
+    let snapshot0 = {
+        let container = run_postgres(
+            &docker,
+            postgres_image.as_str(),
+            "db",
+            "postgres",
+            Some(&volumes),
+        );
+        pre_dump(&container);
+        let snapshot0 = snapshot_db(
```
cevian, line 372: not sure if this is a rust thing but I usually don't like to shadow variables like this. Snapshot0 appears in 2 places and are different vars.

migration/idempotent/001-base.sql
```diff
@@ -1,16 +1,76 @@
...
 BEGIN
     IF command_key IS NOT NULL THEN
        INSERT INTO _prom_catalog.remote_commands(key, command, transactional) VALUES(command_key, command, transactional)
        ON CONFLICT (key) DO UPDATE SET command = excluded.command, transactional = excluded.transactional;
     END IF;
 
     EXECUTE command;
+
+    -- do not call distributed_exec if we are in the middle of restoring from backup
+    _is_restore_in_progress = coalesce((SELECT setting::boolean from pg_catalog.pg_settings where name = 'timescaledb.restoring'), false);
```
cevian, line 69: can't we use _prom_catalog.is_restore_in_progress() here?

migration/idempotent/001-base.sql
```diff
@@ -364,12 +427,15 @@ BEGIN
...
         ) WITH (autovacuum_vacuum_threshold = 100, autovacuum_analyze_threshold = 100)
     $$, NEW.table_name, label_id, NEW.id);
-   EXECUTE format('GRANT SELECT ON TABLE prom_data_series.%I TO prom_reader', NEW.table_name);
-   EXECUTE format('GRANT SELECT, INSERT ON TABLE prom_data_series.%I TO prom_writer', NEW.table_name);
-   EXECUTE format('GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE prom_data_series.%I TO prom_modifier', NEW.table_name);
+    EXECUTE format('GRANT SELECT ON TABLE prom_data_series.%I TO prom_reader', NEW.table_name);
+    EXECUTE format('GRANT SELECT, INSERT ON TABLE prom_data_series.%I TO prom_writer', NEW.table_name);
+    EXECUTE format('GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE prom_data_series.%I TO prom_modifier', NEW.table_name);
+
+    EXECUTE format('CREATE INDEX ON prom_data_series.%I USING GIN (labels)', NEW.table_name);
```
cevian, line 434:
```suggestion
    --these indexes are logically on all series tables but they cannot be defined on the parent due to
    --dump/restore issues.
    EXECUTE format('CREATE INDEX ON prom_data_series.%I USING GIN (labels)', NEW.table_name);
```

migration/incremental/006-tables.sql
```diff
@@ -39,11 +39,6 @@ GRANT SELECT ON TABLE _prom_catalog.series TO prom_reader;
 GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _prom_catalog.series TO prom_writer;
```
cevian, line 41: I'd leave a comment here about the removed parent indexes.

migration/incremental/015-tracing-compression.sql
```diff
@@ -261,82 +261,101 @@ GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.link TO prom_writer;
...
+        (
+            SELECT c.schema_name, c.table_name
+            FROM _timescaledb_catalog.hypertable h
+            INNER JOIN _timescaledb_catalog.hypertable c
+            ON (h.compressed_hypertable_id = c.id)
+            WHERE h.schema_name = '_ps_trace'
+            AND h.table_name IN ('span', 'link', 'event')
+        )
+        LOOP
+            RAISE NOTICE 'removing %.% from extension', _rec.schema_name, _rec.table_name;
```
cevian, line 355: do we want this notice? why?

migration/incremental/021-post-restore.sql
```diff
@@ -0,0 +1,72 @@
+
+CREATE FUNCTION public.promscale_post_restore()
```
cevian, line 2: shouldn't we put this in prom_api? Then we can move this definition to idempotent scripts.

migration/incremental/021-post-restore.sql
```diff
@@ -0,0 +1,72 @@
...
+        WHERE h.schema_name = '_ps_trace'
+        AND h.table_name IN ('span', 'link', 'event')
+        ORDER BY h.table_name, k.table_name
+    )
+    LOOP
+        EXECUTE _sql;
+    END LOOP
+    ;
+
+    SELECT format('ALTER TABLE _ps_trace.tag_key ALTER COLUMN id RESTART WITH %s', max(id) + 1)
```
cevian, line 36: why do we need all these?

migration/incremental/021-post-restore.sql
```diff
@@ -0,0 +1,72 @@
...
+
+    CREATE TRIGGER ts_insert_blocker
+    BEFORE INSERT ON _ps_trace.event
+    FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker();
+
+    CREATE TRIGGER ts_insert_blocker
+    BEFORE INSERT ON _ps_trace.link
+    FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker();
+
+    FOR _sql IN
```
cevian, line 22: let's add a comment explaining why here

migration/incremental/021-post-restore.sql
cevian: also do these queries use indexes correctly to find the max?

sql/promscale--0.0.0.sql
```diff
@@ -356,28 +362,6 @@ END
 $block$
 ;
```
cevian, line 364: why is it safe to delete this code for tracing hypertables?

cevian (changes requested):
Looks good overall, just some minor-ish fixes

migration/incremental/021-post-restore.sql
jgpruitt: sure. I put it in public because that's where `timescaledb_pre_restore` and `timescale_post_restore` are, but I'd rather move it so it can live in idempotent

migration/incremental/021-post-restore.sql
jgpruitt: pg_dump doesn't appear to try to update sequences for extension config tables

jgpruitt: ohhhh wait!

```
The pg_extension_config_dump() function can also register sequences, so 
that the current properties of the registered sequences are saved by 
pg_dump and later restored. The function can support sequences either 
explicitely created with an ALTER SEQUENCE statement or implicitely 
created when a table contains SERIAL or BIGSERIAL columns. Note that the 
sequences associated to SERIAL or BIGSERIAL columns of a configuration 
table need to be registered using the pg_extension_config_dump() 
function if one wants to restore the properties they had at pg_dump time
```
https://www.postgresql.org/message-id/568C27BD.8030402%40free.fr

let me try that

sql/promscale--0.0.0.sql
jgpruitt: We want pg_dump to manage the dump/restore process for the chunks.

migration/incremental/021-post-restore.sql
jgpruitt: resolved in [486d8b5](https://github.com/timescale/promscale_extension/pull/185/commits/486d8b5e16f713ac139bbee875c8c801f1d9868c)

migration/incremental/021-post-restore.sql
jgpruitt: jeez that's so much easier. why did that not occur to me earlier?!?

resolved in [486d8b5](https://github.com/timescale/promscale_extension/pull/185/commits/486d8b5e16f713ac139bbee875c8c801f1d9868c)

migration/incremental/021-post-restore.sql
jgpruitt: resolved in [486d8b5](https://github.com/timescale/promscale_extension/pull/185/commits/486d8b5e16f713ac139bbee875c8c801f1d9868c)

e2e/tests/dump-restore-test.rs
jgpruitt: does it make you like it more or less that this is an anonymous block, and the result of the block is `snapshot0`? Yes, they are 2 different variables, but they contain the same values.

I'll try to break these out to separate functions. Maybe my rust-fu is strong enough now to deal with that

migration/incremental/015-tracing-compression.sql
jgpruitt: nope. leftover debugging cruft. removed in [2088920](https://github.com/timescale/promscale_extension/pull/185/commits/2088920c124ee87330170ec036092ace51fd4090)

migration/incremental/006-tables.sql
jgpruitt: resolved in [356b1a6](https://github.com/timescale/promscale_extension/pull/185/commits/356b1a63a3987e60c00f6f0816beef2941224c03)

migration/idempotent/001-base.sql
jgpruitt: yep. missed this in refactoring. thanks
resolved in [eda2cee](https://github.com/timescale/promscale_extension/pull/185/commits/eda2ceeea4e1d309986c3034a200707a9f5d661d)

sql/promscale--0.0.0.sql
jgpruitt: if we make them config tables, then our `create extension` code has to `create table` the chunks, but since chunks are created dynamically, and I don't think the timescaledb schemas are loaded yet, we don't have enough information to know what chunks we need to create. Have to leave it to pg_dump to do.

e2e/tests/dump-restore-test.rs
jgpruitt: resolved in [9263761](https://github.com/timescale/promscale_extension/pull/185/commits/92637615120da4f7db03e4c6c6adb2e08e5f66ca)

e2e/tests/dump-restore-test.rs
jgpruitt: resolved in [9263761](https://github.com/timescale/promscale_extension/pull/185/commits/92637615120da4f7db03e4c6c6adb2e08e5f66ca)

e2e/scripts/pre-dump.sql
```diff
@@ -0,0 +1,147 @@
+\set ECHO all
+\set ON_ERROR_STOP 1
+
+CREATE EXTENSION promscale;
+
+SELECT _prom_catalog.set_default_value('ha_lease_timeout'::text, '200 hours'::text);
+
+SELECT _prom_catalog.get_or_create_metric_table_name('cpu_usage');
```
cevian, line 8: I'd like to see a metric with a MixedCaseNameThatIsLongerThan63Chars as well in these tests

cevian: I'd also like a test with continuous aggs and  our cagg views

cevian (approved):
Looks great! two ideas for additional tests but probably in a followup PR. Let's make an issue?

cevian:
Lets squash to one or a few commits before merge

e2e/scripts/pre-dump.sql
jgpruitt: made issues for these

migration/idempotent/014-apply-config.sql
```diff
@@ -70,17 +70,32 @@ DECLARE
 BEGIN
     FOR _sql IN
     (
-        -- find tables which belong to the extension so we can mark them to have
+        -- find tables and sequences which belong to the extension so we can mark them to have
```
JamesGuthrie, line 73: I don't think that I understand why this is in an idempotent file. Doesn't it only need to run once?

migration/idempotent/011-maintenance.sql
```diff
@@ -604,4 +601,59 @@ LANGUAGE PLPGSQL;
...
+
+        CREATE TRIGGER ts_insert_blocker
+        BEFORE INSERT ON _ps_trace.event
+        FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker();
+
+        CREATE TRIGGER ts_insert_blocker
+        BEFORE INSERT ON _ps_trace.link
+        FOR EACH ROW EXECUTE FUNCTION _timescaledb_internal.insert_blocker();
+
+        -- Unfortunately, when adding a trigger to a parent table in an inheritance
```
JamesGuthrie, line 634: This is fun.

e2e/scripts/post-snapshot.sql
```diff
@@ -0,0 +1,139 @@
+\set VERBOSITY verbose
```
JamesGuthrie, line 1: Can you add a comment to this file explaining what it does? I _assume_ that the intention is to try and insert a bunch of data, and see that it's possible to do so.

migration/idempotent/014-apply-config.sql
jgpruitt: Ideally, it would only need to run once, however since we are dealing with both the security of the functions and whether or not data gets included in pg_dumps, I felt it was important enough to run after every migration in case something newly added slipped through the cracks.

e2e/scripts/post-snapshot.sql
jgpruitt: yes i will, your assumption is correct

jgpruitt: done in [090040d](https://github.com/timescale/promscale_extension/pull/185/commits/090040d10bc921ac7dfde28f06e3f210ccfa0095)

migration/idempotent/014-apply-config.sql
jgpruitt: WDYT?

JamesGuthrie (approved):

migration/incremental/002-utils.sql
```diff
@@ -11,14 +11,24 @@ CREATE TABLE _prom_catalog.remote_commands(
...
+    _is_restore_in_progress boolean = false;
 BEGIN
     IF command_key IS NOT NULL THEN
        INSERT INTO _prom_catalog.remote_commands(key, command, transactional) VALUES(command_key, command, transactional)
        ON CONFLICT (key) DO UPDATE SET command = excluded.command, transactional = excluded.transactional;
     END IF;
 
     EXECUTE command;
+
+    -- do not call distributed_exec if we are in the middle of restoring from backup
```
sumerman, line 26: But what happens if we are? Do we expect people to dump/restore each node individually?

sql/promscale--0.0.0.sql
sumerman: I'm not sure it answers Mat's question. My understanding is that the same code appears in `incremental/015-tracing-compression.sql`, so we don't need it in `0.0.0`

migration/incremental/002-utils.sql
jgpruitt: As I understand it, we are going to work on multinode support as a fast follow up. Right @cevian ?

cevian: Yes nodes are dump/restored individually

PR #188, opened 2022-04-15
Drop trace check contraints

niksajakovljevic:
It seems that CHECK constrains come with a huge CPU
cost for the database. We can instead move checks to Promscale connector code.
Benchmarks show around 35% improvement in CPU usage.


sumerman (commented):
Would be great if you'd leave constraints as comments as a reference.

jgpruitt (approved):

sumerman (approved):

PR #189, opened 2022-04-15
New support function for tag_map operations

sumerman:
A Rust rewrite of a tag_map support function initially written by @pgnickb (https://github.com/pgnickb/tag_map_supportfunc)

In addition, this implementation strips intermediate type coercions if they are present, making custom `->` op optional.
Plus, I've managed to enable InitPlan optimization by relying on `rewrite_fn_call_to_subquery`.

Another significant difference is polymorphism on supported operators:
Instead of rewriting every operator using `_ps_trace.find_label_ids` that is only suitable for equality, the Rust version of the support function derives a name of a suitable helper function from the operator's underlying function.

Porting TODO list:
- [x] Rewrite the C support function in Rust
- [x] Enable InitPlan optimization for constant arguments
- [x] Handle cases when operator's helper function returns `jsonb[]`
- [x] Port SQL and split it into migrations and idempotent parts
- [x] Add e2e tests for the 👆🏼


migration/incremental/015-tracing-compression.sql
```diff
@@ -181,43 +188,258 @@ DEFAULT FOR TYPE ps_trace.trace_id USING btree AS
...
+    $do$;
+
+    GRANT USAGE ON TYPE ps_trace.tag_map TO prom_reader;
+    GRANT USAGE ON TYPE ps_trace.tag_v TO prom_reader;
+
+    /* XXX
+    * _ps_trace.span table is now used indirectly, thus its name has changed
+    */
+
+    CREATE TABLE _ps_trace._span
```
sumerman, line 365: Why rename it?

migration/incremental/015-tracing-compression.sql
```diff
@@ -181,43 +188,258 @@ DEFAULT FOR TYPE ps_trace.trace_id USING btree AS
...
+/* Move the old tag_map type out of the way */
+ALTER DOMAIN ps_trace.tag_map
+    RENAME TO tag_map_old;
+ALTER DOMAIN ps_trace.tag_map_old
+    SET SCHEMA _ps_trace;
+
+/* Define custom tag_map and tag_v json aliases
+    * NOTE: We need to keep an eye on upstream jsonb type
+    * and keep these up-to-date.
+    */
```
sumerman, line 206: What's up with the whitespace? 🤔

migration/incremental/015-tracing-compression.sql
```diff
@@ -181,43 +188,258 @@ DEFAULT FOR TYPE ps_trace.trace_id USING btree AS
...
+            instrumentation_lib_id,
+            resource_schema_url_id,
+            event_time,
+            dropped_tags_count,
+            dropped_events_count,
+            dropped_link_count,
+            resource_dropped_tags_count,
+            status_code,
+            trace_state,
+            ps_trace.tag_map_denormalize(span_tags) AS span_tags,
```
sumerman, line 429: I don't remember the previous version of `span` to have any denormalization.

src/support.rs
```diff
@@ -116,6 +105,240 @@ mod _prom_ext {
...
+    /// ```
+    ///
+    /// Furthermore, if both `key` and `value` are constant expressions,
+    /// [`rewrite_fn_call_to_subquery`] will be used on the rhs to further optimize it:
+    /// ```sql
+    /// SELECT * FROM some_table
+    /// WHERE map_attribute @> (SELECT _ps_trace.OP_rewrite_helper(key, value))
+    /// ```
+    #[pg_extern(immutable, strict)]
+    pub unsafe fn tag_map_rewrite(input: Internal) -> Internal {
```
cevian, line 142: really great how general this is!

cevian (changes requested):
Overall looks good. Some nits/questions/suggestions below.

migration/idempotent/007-tracing-tags.sql
```diff
@@ -502,3 +502,139 @@ AS $func$
...
+    STYPE = pg_catalog.jsonb
+);
+GRANT EXECUTE ON FUNCTION _ps_trace.jsonb_cat(pg_catalog.jsonb) TO prom_reader;
+
+
+CREATE FUNCTION _ps_trace.tag_map_denormalize(_map ps_trace.tag_map)
+    RETURNS ps_trace.tag_map
+    LANGUAGE sql STABLE
+    PARALLEL SAFE AS
+/* NOTE: This function cannot be inlined since it's used in a scalar
```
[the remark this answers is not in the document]
pgnickb: My concern is the performance penalty that comes with the `set` clause here. I think it's better to write a search_path-agnostic code than pay the cost of the set-clause.

migration/idempotent/007-tracing-tags.sql
```diff
@@ -502,3 +502,139 @@ AS $func$
...
+ * context and uses an aggregate
+ */
+$fnc$
+    SELECT _ps_trace.jsonb_cat(pg_catalog.jsonb_build_object(t.key, t.value))
+        FROM pg_catalog.jsonb_each(_map) f(k,v)
+            JOIN _ps_trace.tag t ON
+                    f.k::int8 OPERATOR(pg_catalog.=) t.key_id
+                AND f.v::int8 OPERATOR(pg_catalog.=) t.id;
+$fnc$;
```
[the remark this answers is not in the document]
pgnickb: Yeah this works.

migration/idempotent/007-tracing-tags.sql
```diff
@@ -502,3 +502,139 @@ AS $func$
...
+DO $do$
+BEGIN
+	CREATE OPERATOR ps_trace.->	(
+	    FUNCTION = ps_trace.tag_map_object_field,
+	    LEFTARG  = ps_trace.tag_map,
+	    RIGHTARG = pg_catalog.text
+	);
+EXCEPTION
+    WHEN SQLSTATE '42723' THEN -- operator already exists
+        null;
```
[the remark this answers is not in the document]
pgnickb: I agree in principle, but I want to point out that for these operators to exist there should also be corresponding types defined, which makes testing this code path rather challenging.
That being said, we can do it, just confirm that we want to go through these lengths.

JamesGuthrie (changes requested):

migration/incremental/015-tracing-compression.sql
```diff
@@ -181,40 +181,203 @@ DEFAULT FOR TYPE ps_trace.trace_id USING btree AS
...
     OPERATOR 1 ps_trace.=,
     FUNCTION 1 _ps_trace.trace_id_hash(ps_trace.trace_id);
 
 GRANT USAGE ON TYPE ps_trace.trace_id TO prom_reader;
 
+/* Move the old tag_map type out of the way */
+ALTER DOMAIN ps_trace.tag_map
+    RENAME TO tag_map_old;
+ALTER DOMAIN ps_trace.tag_map_old
+    SET SCHEMA _ps_trace;
```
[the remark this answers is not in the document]
sumerman: This will require dropping everything that depends on them. I believe we agreed to have a follow-up PR for that.

src/support.rs
```diff
@@ -116,6 +105,240 @@ mod _prom_ext {
...
+        // Wrapping core logic into a funciton, that returns a Result,
```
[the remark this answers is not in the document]
sumerman: we'll lose `?` shorthand, wont we?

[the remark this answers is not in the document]
JamesGuthrie: I don't think so.

migration/incremental/015-tracing-compression.sql
```diff
@@ -181,40 +181,203 @@ DEFAULT FOR TYPE ps_trace.trace_id USING btree AS
...
+    /* Create subscript_handler function for pg v14+ and the type accordingly.
+    * For pg v13 jsonb type doesn't have a subscript_handler function so it shall
+    * be omitted.
+    */
+    DECLARE
+        _pg_version int4 := current_setting('server_version_num')::int4;
+    BEGIN
+        IF (_pg_version >= 140000) THEN
+            EXECUTE
+                'CREATE OR REPLACE FUNCTION ps_trace.tag_map_subscript_handler(internal) ' ||
```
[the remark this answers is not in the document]
pgnickb: Good point. The usage like this makes total sense:
```SQL
select (jsonb'{"a":1, "b":2}')['a'];
```
We decided to address this in a future PR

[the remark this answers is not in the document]
pgnickb: Apparently the return type of the `subscript` function is `internal`:
```SQL
CREATE OR REPLACE FUNCTION ps_trace.tag_map_subscript_handler(internal)
    RETURNS internal
    LANGUAGE internal
    IMMUTABLE PARALLEL SAFE STRICT
    AS $f$jsonb_subscript_handler$f$;
```

So nothing to do here for now. I'll dive deeper into it later

pgnickb:
I've addressed some of the minor changes Mat pointed out. Some of the changes either need to be done by Valeriy or at least with his involvement.

One more thing to address here are the E2E tests which need to be generalized. And extended.

src/pg_imports.rs
```diff
@@ -0,0 +1,110 @@
...
+#[cfg(any(feature = "pg12", feature = "pg13"))]
+#[inline]
+pub fn set_sa_hashfuncid(
+    _scalar_array_op: &mut pg_sys::ScalarArrayOpExpr,
+    _hash_func_oid: pg_sys::Oid,
+) {
+    // the field didn't exist prior to pg14
+}
+
+// pg_guard doesn't compile, so we have to do without it for now.
```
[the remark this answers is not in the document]
sumerman: > Hmm how bad is this? Can it crash or does this behaviour become undefined?

I think it's benign for our use case. Our extension doesn't have any global state. IIRC when C-functions crash they bring down the entire PostgreSQL backend. So, I don't expect to run into undefined behaviour because of it.



> Do we know why it doesn't compile?

I've tried asking in the pgx discord but didn't get a definitive answer.

https://discord.com/channels/561648697805504526/749454413407584258/964516348711542824

src/support.rs
```diff
@@ -116,6 +105,240 @@ mod _prom_ext {
...
+
+            // Make an attempt to move the function call into an InitPlan
+            // by calling [`rewrite_fn_call_to_subquery`]
+            let mut sub_simplify_req = pg_sys::SupportRequestSimplify {
+                type_: pg_sys::NodeTag_T_SupportRequestSimplify,
+                root: (*req).root,
+                fcall: helper_func_expr,
+            };
+            let helper_expr_init_plan = fcinfo::direct_pg_extern_function_call_as_datum(
+                rewrite_fn_call_to_subquery_wrapper,
```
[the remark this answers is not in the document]
sumerman: I didn't think it was easily possible with the `Internal` type, but it is. Thanks.

src/lib.rs
```diff
@@ -13,6 +14,18 @@ mod util;
...
 pg_module_magic!();
 
+pub fn build_pg_list_of_strings<'a, I>(parts: I) -> PgList<pg_sys::Value>
+where
+    I: IntoIterator<Item = &'a str>,
+{
+    let mut res = PgList::new();
+    for p in parts {
+        let cstr = ::std::ffi::CString::new(p).unwrap().into_raw();
+        res.push(unsafe { pg_sys::makeString(cstr) });
```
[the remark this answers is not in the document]
sumerman: I didn't pick up on the `into_` vs `as_` convention here and with `into_pg`. Thank you! 

> The documentation for `makeString` says:
> 
> > Caller is responsible for passing a palloc'd string.

AFAICT from usage inside  PostgreSQL itself: it doesn't have to be palloc'd per se, but the string needs to be in the same or a longer-lived scope.

migration/idempotent/007-tracing-tags.sql
```diff
@@ -502,3 +502,139 @@ AS $func$
...
+CREATE OR REPLACE FUNCTION ps_trace.tag_v_eq(_ps_trace.tag_v, pg_catalog.jsonb)
+    RETURNS pg_catalog.bool
+    LANGUAGE internal
+        IMMUTABLE
+        PARALLEL SAFE
+        SUPPORT _prom_ext.tag_map_rewrite
+    AS 'jsonb_eq';
+GRANT EXECUTE ON FUNCTION ps_trace.tag_v_eq(_ps_trace.tag_v, pg_catalog.jsonb) TO prom_reader;
+
+CREATE OR REPLACE FUNCTION _ps_trace.tag_v_eq_rewrite_helper(_tag_key pg_catalog.text, _value pg_catalog.jsonb)
```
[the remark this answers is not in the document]
sumerman: I like it. It describes what a helper func. does

src/palloc.rs
```diff
@@ -13,6 +15,28 @@ pub unsafe fn in_memory_context<T, F: FnOnce() -> T>(mctx: pg_sys::MemoryContext
...
+            None
+        } else {
+            Some(PallocdString {
+                pg_box: unsafe { PgBox::<_, AllocatedByRust>::from_rust(ptr) },
+            })
+        }
+    }
+
+    pub fn as_c_str(&self) -> &CStr {
+        unsafe { CStr::from_ptr(self.pg_box.as_ptr()) }
```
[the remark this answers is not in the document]
sumerman: I don't think we can make an assertion that would keep this safe: one can't do much with a non-NULL-terminated string without knowing its length. But I will make `from_ptr` unsafe and add a safety contract.

sumerman:
I believe I've addressed all review comments concerning the Rust part. Plus rebased on fresh `develop` to clear merge conflicts.

e2e/tests/snapshots/golden_tests__golden_test_testdata_tag_map_sql.snap
```diff
@@ -0,0 +1,143 @@
...
+               Join Filter: (s.instrumentation_lib_id = il.id)
+               ->  Nested Loop Left Join
+                     ->  Nested Loop Left Join
+                           Join Filter: (s.operation_id = o.id)
+                           ->  Custom Scan (ChunkAppend) on span s
+                                 ->  Bitmap Heap Scan on _hyper_7_1_chunk s_1
+                                       Recheck Cond: (resource_tags @> $1)
+                                       Filter: (span_tags @> ANY ($0))
+                                       ->  Bitmap Index Scan on _hyper_7_1_chunk_span_resource_tags_idx
+                                             Index Cond: (resource_tags @> $1)
```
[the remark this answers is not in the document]
sumerman: TBH I don't think we should be testing index condition optimization here. At least not right now, perhaps as a follow-up. The goal was to make sure operators, helpers and support functions work in tandem and apply rewrites.

migration/idempotent/007-tracing-tags.sql
```diff
@@ -497,8 +497,136 @@ AS $func$
...
+	    FUNCTION       = ps_trace.tag_v_ne,
+	    LEFTARG        = _ps_trace.tag_v,
+	    RIGHTARG       = pg_catalog.jsonb,
+	    NEGATOR        = OPERATOR(ps_trace.=),
+	    RESTRICT       = neqsel,
+	    JOIN           = neqjoinsel
+	);
+EXCEPTION
+    WHEN SQLSTATE '42723' THEN -- operator already exists
+        null;
```
cevian, line 630: needs an ownership set just like the operators above.

migration/incremental/015-tracing-compression.sql
```diff
@@ -190,34 +190,197 @@ GRANT USAGE ON TYPE ps_trace.trace_id TO prom_reader;
...
```
cevian, line 246: Lets create an issue to look at subscripts

migration/incremental/015-tracing-compression.sql
```diff
@@ -190,34 +190,197 @@ GRANT USAGE ON TYPE ps_trace.trace_id TO prom_reader;
...
+
+/* Define custom tag_map and tag_v json aliases
+    * NOTE: We need to keep an eye on upstream jsonb type
+    * and keep these up-to-date.
+    */
+
+    CREATE TYPE ps_trace.tag_map;
+    CREATE TYPE _ps_trace.tag_v;
+
+    CREATE OR REPLACE FUNCTION ps_trace.tag_map_in(cstring)
```
cevian, line 207: all these functions should also be defined in an idempotent script because all our functions should appear in an idempotent script

src/lib.rs
```diff
@@ -148,6 +160,6 @@ pub mod pg_test {
 
     pub fn postgresql_conf_options() -> Vec<&'static str> {
         // return any postgresql.conf settings that are required for your tests
-        vec!["search_path = 'public, _prom_ext, ps_trace'"]
+        vec!["search_path = 'public, _prom_ext, ps_trace, _ps_trace'"]
```
cevian, line 163: hmm actually why is the private schema needed in the tests? Does that mean users will also need it in their search_path for a good query experience (that shouldn't be the case).

cevian (approved):

src/lib.rs
pgnickb: There isn't a really good reason for this, to be honest. Users shouldn't need this.
The reason was plan output contained schema name since it wasn't in the `search_path`. We decided that it'd be much nicer to change search path when we merge the commit where we make new schema structure/naming. The alternative would be to update all tests (which was deemed not as nice).

sumerman: Yeah, it's just convenience for unit tests. E2E doesn't rely on this.

pgnickb:
This query crashes pg for me:
```SQL
select (select count(*) from span where span_tags -> 'key-55' = '"val-20"');
```
Will look into it after having lunch.

src/lib.rs
```diff
@@ -13,6 +14,17 @@ mod util;
 
 pg_module_magic!();
 
+pub fn build_pg_list_of_cstrings<'a, I>(parts: I) -> PgList<pg_sys::Value>
```
JamesGuthrie, line 17: I'm still not quite happy with this. The way we're using it right now is okay, but it's kind of a time bomb.

The following example will compile, but should not:

```rust
    let cstr1 = CString::new("Hello").unwrap();
    let cstr2 = CString::new("World").unwrap();

    let pg_list = crate::build_pg_list_of_cstrings([cstr1.as_bytes(), cstr2.as_bytes()]);

    // This is totally unsafe, `pg_list` is still holding a reference to this.
    drop(cstr1);
```

Ideally we would solve this with lifetimes, but I think that would require reworking (or wrapping) `PgList`. That would be too much work (for now).

As a short-term solution, I would suggest that we mark this function as unsafe, and specify in the function's documentation that the _caller_ must ensure that the parameters which are passed into this function live for longer than the returned `PgList`.

src/palloc.rs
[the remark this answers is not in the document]
JamesGuthrie: 👍

migration/idempotent/007-tracing-tags.sql
```diff
@@ -497,8 +497,136 @@ AS $func$
...
+    PARALLEL SAFE AS
+/* NOTE: This function cannot be inlined since it's used in a scalar
+ * context and uses an aggregate. Yet we don't want to use `SET` clause
+ * for the search_path, thus it is important to make it completely
+ * search_path-agnostic.
+ */
+$fnc$
+    SELECT pg_catalog.jsonb_object_agg(t.key, t.value)
+        FROM pg_catalog.jsonb_each(_map) f(k,v)
+            JOIN _ps_trace.tag t ON f.v::pg_catalog.int8 OPERATOR(pg_catalog.=) t.id;
```
jgpruitt, line 522: just a note: this is fine now that we removed partitioning on tag, but if we reintroduce it, we'll want to be sure the implementation here invokes partition elimination

JamesGuthrie (changes requested):

src/lib.rs
sumerman: How about this:
```rust
pub fn with_pg_list_of_cstrings<'a, I, R, F>(parts: I, f: F) -> R
where
    I: IntoIterator<Item = &'a [u8]>,
    F: Fn(&'a PgList<pg_sys::Value>) -> R,
{
    let mut list = PgList::new();
    for p in parts {
        list.push(unsafe { pg_sys::makeString(p.as_ptr() as _) });
    }
    f(&list)
}
```

sumerman: Usage example:

```rust
            let jsonb_contains_op_oid =
                with_pg_list_of_cstrings(CONTAINS_OP_PATH, |jsonb_contains_fully_qualified_name| {
                    LookupOperName(
                        std::ptr::null_mut(),
                        jsonb_contains_fully_qualified_name.as_ptr(),
                        pg_sys::JSONBOID,
                        pg_sys::JSONBOID,
                        false, // Raises an error if the operator is not found
                        -1,
                    )
                });
```

sumerman: Nevermind, `PgList`'s API is just unsound:

```rust
let mut lst = PgList::<pg_sys::Value>::new();
with_pg_list_of_cstrings(CONTAINS_OP_PATH, |foo| {
    lst.push(foo.pop().unwrap());
});
```

JamesGuthrie (approved):

migration/idempotent/009-tracing-views.sql
```diff
@@ -13,7 +12,7 @@ SELECT
     s.end_time,
     tstzrange(s.start_time, s.end_time, '[]') as time_range,
     s.duration_ms,
-    s.span_tags,
+    _ps_trace.tag_map_denormalize(s.span_tags) as span_tags,
```
jgpruitt, line 15: In the previous implementation, this column contained the tag_map rather than the actual tag_keys->tag_values. It was left to the user to decide whether to denormalize the tags. Have we verified that postgres will omit the call to tag_map_denormalize if this column is not selected?

migration/idempotent/009-tracing-views.sql
```diff
@@ -23,7 +22,7 @@ SELECT
     il.name as instrumentation_lib_name,
     il.version as instrumentation_lib_version,
     u1.url as instrumentation_lib_schema_url,
-    s.resource_tags,
+    _ps_trace.tag_map_denormalize(s.resource_tags) as resource_tags,
```
jgpruitt, line 25: same as above

migration/idempotent/009-tracing-views.sql
jgpruitt: we are denormalizing the tags in this view, but not in the event or link views. Is there a reason to do it in one and not the others?

migration/incremental/015-tracing-compression.sql
```diff
@@ -234,6 +241,167 @@ GRANT USAGE ON TYPE ps_trace.trace_id TO prom_reader;
...
+/* Move the old tag_map type out of the way */
+ALTER DOMAIN ps_trace.tag_map
+    RENAME TO tag_map_old;
+ALTER DOMAIN ps_trace.tag_map_old
+    SET SCHEMA _ps_trace;
+
+/* Define custom tag_map and tag_v json aliases
+ * NOTE: We need to keep an eye on upstream jsonb type
+ * and keep these up-to-date.
+ */
```
jgpruitt, line 253: why don't we just drop the domain?

migration/incremental/015-tracing-compression.sql
```diff
@@ -234,6 +241,167 @@ GRANT USAGE ON TYPE ps_trace.trace_id TO prom_reader;
...
+DECLARE
+    _pg_version int4 := current_setting('server_version_num')::int4;
+BEGIN
+    IF (_pg_version >= 140000) THEN
+        EXECUTE
+            'CREATE OR REPLACE FUNCTION ps_trace.tag_map_subscript_handler(internal) ' ||
+                'RETURNS internal '                 ||
+                'LANGUAGE internal '                ||
+                'IMMUTABLE PARALLEL SAFE STRICT '   ||
```
jgpruitt, line 301: does this mean that, if I install promscale on v13, and then upgrade to v14, I will never get the subscript functionality?

migration/idempotent/009-tracing-views.sql
sumerman: > In the previous implementation, this column contained the tag_map rather than the actual tag_keys->tag_values. It was left to the user to decide whether to denormalize the tags. Have we verified that postgres will omit the call to tag_map_denormalize if this column is not selected?

No, we didn't


> we are denormalizing the tags in this view, but not in the event or link views. Is there a reason to do it in one and not the others?

We probably didn't discuss what views require update hence others fell off the radar. Will add them in a follow-up PR.

migration/incremental/015-tracing-compression.sql
sumerman: We agreed to do that in a follow-up PR

migration/incremental/015-tracing-compression.sql
pgnickb: Yes, which shall be fixed if we address Mat's comment on adding all these to idempotent branch.

PR #190, opened 2022-04-16
Prevent duplicate or missing numeric prefixes

jgpruitt:
Migration files and idempotent filenames are prefixed with a number which determines the order in which they are combined during build. This commit checks these prefixes to ensure that there are no duplicate numbers present nor gaps in the numbering.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -110,10 +110,30 @@ fn render_file(path: PathBuf, sql_type: &SqlType) -> Result<String, Error> {
     let mut result = String::new();
     let mut paths: Vec<PathBuf> = fs::read_dir(path)?.map(|f| f.unwrap().path()).collect();
     paths.sort();
+    let mut prev = 0;
     paths.into_iter().for_each(|path| {
+        let curr = numeric_prefix(&path);
+        // assert that the numeric prefixes are in order with no gaps or duplicates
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -110,12 +110,24 @@ fn render_file(path: PathBuf, sql_type: &SqlType) -> Result<String, Error> {
...
         result += wrap(&path, sql_type).as_str();
     });
     Ok(result)
 }
 
+/// Parses and returns the numeric prefix in the filename of each idempotent and migration file
+fn numeric_prefix(path: &Path) -> i32 {
+    let file_name = path.file_name().unwrap().to_str().unwrap();
+    let (prefix, _) = file_name.split_once('-').unwrap();
+    prefix.parse::<i32>().unwrap()
```
sumerman, line 128: Since someone could conceivably forget to put a number in front of their migration name I suggest failing with an explicit message instead of `unwrap()`.

sumerman (commented):
Mostly LGTM, just one nit.

build.rs
jgpruitt: will do

jgpruitt: fixed!

sumerman (approved):

PR #193, opened 2022-04-19
Simultaneous Promscale versions

sumerman:
Addresses #113 

Verified following a procedure, similar to the one described [here](https://github.com/timescale/timescaledb-docker-ha/pull/227#issuecomment-1101568154)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -94,6 +94,7 @@ package: promscale.control ## Generate extension artifacts for packaging
 
 .PHONY: install
 install: promscale.control ## Install the extension in the Postgres found via pg_config
+	cargo pgx schema pg${PG_BUILD_VERSION} --out sql/promscale--${EXT_VERSION}.sql --release
 	bash create-upgrade-symlinks.sh
 	cargo pgx install --pg-config ${PG_CONFIG}
 
--- a/README.md
+++ b/README.md
@@ -72,6 +72,13 @@ See the Postgres [documentation](https://www.postgresql.org/download/linux/redha
     yum localinstall -y promscale_extension-0.3.0.pg14.x86_64.rpm
     ```
 
+### Docker images
+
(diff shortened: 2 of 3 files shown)


JamesGuthrie (changes requested):
Do we not also need to make changes to the github workflow to make it build the latest alpine image with the correct versions in it?

sumerman:
> Do we not also need to make changes to the github workflow to make it build the latest alpine image with the correct versions in it?

AFAICT `timescaledev/promscale-extension:latest-ts2-pg${PG_VERSION}` was published by hand in the past. I'm not sure if I should be changing that. Assuming the `latest-ts2-pg${PG_VERSION}` tag is correctly maintained in the future the current change should suffice as every subsequent image would include everything the previous version had.

JamesGuthrie:
> > Do we not also need to make changes to the github workflow to make it build the latest alpine image with the correct versions in it?
> 
> AFAICT `timescaledev/promscale-extension:latest-ts2-pg${PG_VERSION}` was published by hand in the past. I'm not sure if I should be changing that. Assuming the `latest-ts2-pg${PG_VERSION}` tag is correctly maintained in the future the current change should suffice as every subsequent image would include everything the previous version had.

I think that it was published by hand in the past, but I thought that we changed that since the most recent release.

sumerman:
> > > Do we not also need to make changes to the github workflow to make it build the latest alpine image with the correct versions in it?
> > 
> > 
> > AFAICT `timescaledev/promscale-extension:latest-ts2-pg${PG_VERSION}` was published by hand in the past. I'm not sure if I should be changing that. Assuming the `latest-ts2-pg${PG_VERSION}` tag is correctly maintained in the future the current change should suffice as every subsequent image would include everything the previous version had.
> 
> I think that it was published by hand in the past, but I thought that we changed that since the most recent release.

I can only see automated publishing to `ghcr.io/timescale/dev_promscale_extension:` that we build as a part of CI workflow. AFAICT release workflow builds and publishes packages, but not the image.

JamesGuthrie (approved):

PR #195, opened 2022-04-21
Secure the takeover script

cevian:
Make sure that all objects made part of the extension are owned by
the object owner. Objects owned by someone else can easily be exploited
later. Do this automatically by traversing the depend tree.

### DIFF
--- a/sql/promscale--0.0.0.sql
+++ b/sql/promscale--0.0.0.sql
@@ -1,3 +1,4 @@
+SET LOCAL search_path = pg_catalog;
 DROP TABLE public.prom_schema_migrations;
 
 REVOKE EXECUTE ON FUNCTION ps_trace.delete_all_traces() FROM prom_writer;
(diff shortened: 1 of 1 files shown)


sql/promscale--0.0.0.sql
```diff
@@ -505,3 +506,361 @@ CREATE TYPE _prom_ext.GapfillDeltaTransition (
...
+                'Object ' || pg_describe_object(classid, objid, objsubid) || 'is owned by ' || owner::text
+        )
+        into errors
+        from with_owner
+        where (
+            --either the owner is the object owner
+            (owner = obj_owner_oid)
+            OR
+            --or prom_admin (these are for metric and series tables)
+            (owner = (SELECT oid FROM pg_roles WHERE rolname = 'prom_admin') AND classid IN ('pg_class'::regclass, 'pg_type'::regclass))
```
jgpruitt, line 608: why is pg_type considered here? metric and series tables would both be pg_class, right?

jgpruitt (approved):
just one question

sql/promscale--0.0.0.sql
cevian: each pg_class creates a corresponding pg_type internally, always. It's for the record type.

JamesGuthrie (changes requested):
In general it looks good to me. I tested it with a few simple scenarios to validate that it does what's expected:

- change owner of existing function to other role
- change owner of existing view to other role
- add function with other owner
- add operator with other owner
- add table with other owner

In case anybody else would like to try the test script, it's available in a gist here: https://gist.github.com/JamesGuthrie/c16a18defb28b40141ff904dce9a37b3

JamesGuthrie (approved):

cevian:
passed in https://github.com/timescale/promscale/actions/runs/2229919658

PR #196, opened 2022-04-21
Add pgspot lint check

JamesGuthrie:
The pgspot [1] tool is used to spot vulnerabilities in PostgreSQL
extension scripts.

[1]: github.com/timescale/pgspot

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,8 +2,6 @@ name: ci
 on:
   pull_request:
     paths-ignore:
-    - '.github/workflows/docker.yaml'
-    - '.github/workflows/lint.yaml'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yaml
```diff
@@ -2,8 +2,6 @@ name: ci
 on:
   pull_request:
     paths-ignore:
-    - '.github/workflows/docker.yaml'
```
cevian, line 5: why are we removing the first two workflow files but not `'.github/workflows/release.yaml'`?

sumerman: The release isn't required when running PR checks.

sumerman (approved):

PR #197, opened 2022-04-21
Fix `safe_approximate_row_count` when timescaledb not present

JamesGuthrie:
Postgres doesn't short-circuit evaluate `IF cond_a AND cond_b` when
`cond_a` is false, so it to evaluates `cond_b`. In this case the
evaluation of `cond_b` will fail when `cond_a` is false.


cevian (approved):
what, really there is no short circuting?

PR #199, opened 2022-04-25
Re-creating tag table without partitions.

sumerman:
As a follow-up to https://github.com/timescale/promscale/issues/748

In the future, should we decide to adopt a list or range partitioning scheme, the current tag table can [be converted into a default partition](https://github.com/timescale/promscale/issues/748#issuecomment-1109041670).

### DIFF
--- a/migration/incremental/015-tracing-compression.sql
+++ b/migration/incremental/015-tracing-compression.sql
@@ -2,8 +2,8 @@
 DROP TABLE _ps_trace.link CASCADE;
 DROP TABLE _ps_trace.event CASCADE;
 DROP TABLE _ps_trace.span CASCADE;
+DROP TABLE _ps_trace.tag CASCADE;
 
-TRUNCATE TABLE _ps_trace.tag;
 DELETE FROM _ps_trace.tag_key WHERE id >= 1000;
 ANALYZE _ps_trace.tag_key;
 REINDEX TABLE _ps_trace.tag_key;
@@ -15,8 +15,16 @@ ANALYZE _ps_trace.instrumentation_lib;
 TRUNCATE TABLE _ps_trace.schema_url CASCADE;
 ANALYZE _ps_trace.schema_url;
 
-ALTER TABLE _ps_trace.tag DROP CONSTRAINT tag_key_value_id_key_id_key;
-CREATE UNIQUE INDEX tag_key_value_id_key_id_key ON _ps_trace.tag (key, _prom_ext.jsonb_digest(value)) INCLUDE (id, key_id);
+CREATE TABLE _ps_trace.tag (
+    id BIGINT NOT NULL GENERATED ALWAYS AS IDENTITY,
+    key_id BIGINT NOT NULL,
+    tag_type ps_trace.tag_type NOT NULL,
+    key ps_trace.tag_k NOT NULL REFERENCES _ps_trace.tag_key (key) ON DELETE CASCADE,
+    value ps_trace.tag_v NOT NULL
+);
+CREATE UNIQUE INDEX tag_key_value_id_key_id_key_idx ON _ps_trace.tag (key, _prom_ext.jsonb_digest(value)) INCLUDE (id, key_id);
+GRANT SELECT ON TABLE _ps_trace.tag TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.tag TO prom_writer;
 
 -- We are replacing domain with custom type
 DROP DOMAIN ps_trace.trace_id CASCADE;


migration/migration/015-tracing-compression.sql
```diff
@@ -15,8 +15,17 @@ ANALYZE _ps_trace.instrumentation_lib;
 TRUNCATE TABLE _ps_trace.schema_url CASCADE;
 ANALYZE _ps_trace.schema_url;
 
-ALTER TABLE _ps_trace.tag DROP CONSTRAINT tag_key_value_id_key_id_key;
+CREATE TABLE _ps_trace.tag
+(
+    id BIGINT NOT NULL GENERATED ALWAYS AS IDENTITY,
+    tag_type ps_trace.tag_type NOT NULL,
```
JamesGuthrie, line 21: By reordering `tag_type` and `key_id` we would get better alignment.

```
> SELECT
  pg_column_size(row (1::BIGINT, 1::INT2, 1::BIGINT, ''::text, 'null'::jsonb)) AS base_row,                                                       
  pg_column_size(row (1::BIGINT, 1::BIGINT, 1::INT2, ''::text, 'null'::jsonb)) AS aligned_row
;

 base_row | aligned_row
----------+-------------
       58 |          52
(1 row)
```

migration/migration/012-tracing.sql
```diff
@@ -66,30 +66,8 @@ GRANT SELECT ON TABLE _ps_trace.tag TO prom_reader;
...
-            GRANT SELECT ON TABLE _ps_trace.tag_%s TO prom_reader
-            $sql$, _i);
-        EXECUTE format($sql$
-            GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _ps_trace.tag_%s TO prom_writer
-            $sql$, _i);
-    END LOOP;
-END
-$block$
-;
+-- we used to create a bunch of partitions here, but since we drop
```
JamesGuthrie, line 69: I don't think that we should be retroactively modifying the migration scripts. If we do, we lose the ability to reason about what was present when. I might be in the minority here though.

JamesGuthrie (changes requested):

migration/migration/012-tracing.sql
jgpruitt: agreed. `012-tracing.sql` has already been released into the wild. Granted our new migration tracking system hasn't, but I THINK that since the 0.0.0 script prepopulates the migration table and includes 12, this migration will probably get skipped.

sumerman: My point is that it does a lot of useless work when it's not skipped, but I won't insist on having it.

jgpruitt:
The tag_maps operators are coded to take advantage of partition elimination. Should we modify that in this PR, or is the new implementation of tag_maps going to address it?

sumerman:
> The tag_maps operators are coded to take advantage of partition elimination. Should we modify that in this PR, or is the new implementation of tag_maps going to address it?

I don't think we need to change them: the same WHERE conditions are necessary to make use of the unique index.

sumerman:
> Gotcha. Makes sense. What about the `ps_trace.jsonb` functions though? I think those might be improved if we don't need to worry about partition elimination.

We are likely to drop them in a follow-up to the current tag map PR

migration/incremental/015-tracing-compression.sql
```diff
@@ -15,8 +15,16 @@ ANALYZE _ps_trace.instrumentation_lib;
...
-ALTER TABLE _ps_trace.tag DROP CONSTRAINT tag_key_value_id_key_id_key;
-CREATE UNIQUE INDEX tag_key_value_id_key_id_key ON _ps_trace.tag (key, _prom_ext.jsonb_digest(value)) INCLUDE (id, key_id);
+CREATE TABLE _ps_trace.tag (
+    id BIGINT NOT NULL GENERATED ALWAYS AS IDENTITY,
+    key_id BIGINT NOT NULL,
+    tag_type ps_trace.tag_type NOT NULL,
+    key ps_trace.tag_k NOT NULL REFERENCES _ps_trace.tag_key (key) ON DELETE CASCADE,
+    value ps_trace.tag_v NOT NULL
+);
+CREATE UNIQUE INDEX tag_key_value_id_key_id_key_idx ON _ps_trace.tag (key, _prom_ext.jsonb_digest(value)) INCLUDE (id, key_id);
```
jgpruitt, line 25: Did you happen to benchmark storing the digest in a column vs a functional index?

sumerman: No, I didn't.

jgpruitt: if we have to change the jsonb_digest function, will we have to drop and recreate the functional index?

I'm wondering whether the function is so fast that storing the digest in a column is not worth the storage overhead, or if storing it would make a noticeable difference in query performance. WDYT?

sumerman: > if we have to change the jsonb_digest function, will we have to drop and recreate the functional index?

Correct.

> I'm wondering whether the function is so fast that storing the digest in a column is not worth the storage overhead, or if storing it would make a noticeable difference in query performance. WDYT?

If we don't store it, how do we enforce the uniqueness constraint?

jgpruitt: it's stored in the index, but not the table. I was thinking about the difference between the two statements below, however I now realize that there's no way around doing the digest on the right side, and postgres just looks at the value stored in the index for the left side (it doesn't have to recompute). So, I withdraw my question. lol

`select * from tag t where jsonb_digest(t.value) = jsonb_digest(_thing_we_want_to_find);`

`select * from tag t where t.digest =  jsonb_digest(_thing_we_want_to_find);`

cevian (approved):

JamesGuthrie (approved):

PR #210, opened 2022-04-28
Merge current master into develop

JamesGuthrie:


cevian:
I'm ok with these changes but won't this make merging develop into master more difficult? I am actually not 100% sure

JamesGuthrie:
> I'm ok with these changes but won't this make merging develop into master more difficult? I am actually not 100% sure

I think that it should make merging develop into master easier.

PR #214, opened 2022-04-29
Fix unqualified casts in SQL scripts

JamesGuthrie:


sumerman (approved):
the change itself looks good, but AFAICT it's far from being exhaustive.

cevian (approved):

JamesGuthrie:
> the change itself looks good, but AFAICT it's far from being exhaustive.

Do you have examples of cases which you think are not being correctly caught?

sumerman:
> > the change itself looks good, but AFAICT it's far from being exhaustive.
> 
> Do you have examples of cases which you think are not being correctly caught?

E.g. `007-matcher-operators.sql` has more unqualified casts.

JamesGuthrie:
> > > the change itself looks good, but AFAICT it's far from being exhaustive.
> > 
> > 
> > Do you have examples of cases which you think are not being correctly caught?
> 
> E.g. `007-matcher-operators.sql` has more unqualified casts.

Interestingly, `::int` does not qualify as an unqualified cast, because the parser expands it to `pg_catalog.integer`.

PR #218, opened 2022-04-29
JIT off by default

niksajakovljevic:
PG JIT compilation doesn't play nicely with TimescaleDB planner
and can cause a huge slowdown for specific queries so we are turning it off.
We might revisit this once JIT issues are fixed in TimescaleDB. (it might not be all TimescaleDB fault as PG itself can perform poorly when JIT is on for specific queries)

Setting JIT on database level will also disable it for 3rd party apps connecting to database (which is a good thing imho)

Related TimescaleDB issue https://github.com/timescale/timescaledb/issues/4222

### DIFF
--- a/migration/incremental/019-jit-off.sql
+++ b/migration/incremental/019-jit-off.sql
@@ -1,0 +1,8 @@
+-- PG JIT compilation doesn't play nicely with TimescaleDB planner and can cause a huge 
+-- slowdown for specific queries so we are turning it off.
+-- We might revisit this once JIT issues are fixed in TimescaleDB */
+DO $$
+    BEGIN
+        EXECUTE format('ALTER DATABASE %I SET jit = off', current_database());
+    END
+$$;


cevian (approved):
LGTM. We should figure out how to document this along with search_path changes. cc: @JamesGuthrie 

We should also make sure others agree with this change

migration/incremental/019-jit-off.sql
```diff
@@ -0,0 +1,8 @@
+-- PG JIT compilation doesn't play nicely with TimescaleDB planner and can cause a huge
```
JamesGuthrie, line 1: Do we have bug reports against timescaledb for the jit issues, can we link them here as well so that it will be easier to assess in future if the jit issues have been resolved?

niksajakovljevic: I linked one issue reported by Mat however I believe I found additional case (need still to confirm).

niksajakovljevic: So we don't have a definitive list of issues yet

JamesGuthrie: Github descriptions aren't reflected in the code, and aren't easy to find from the code. I would have preferred this to have been in the migration file.

PR #221, opened 2022-05-02
Use pgx 0.4.4

JamesGuthrie:
This bumps to pgx 0.4.4 with our contributed versioned .so change.


sumerman:
As a side note: this change will also require changing the build script in `timescale-ha` as it has to pick and install pgx versions.

PR #224, opened 2022-05-02
Handle pre-release releases

cevian:
On pre-release:
- send packages to the timescaledb-exp package cloud repo
- mark the github release as pre-release

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -56,10 +56,20 @@ jobs:
           tag="$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].version')"; \
           version="${tag}-snapshot"; \
         fi;
+        # a dash indicates a pre-release version in semver
+        if [[ "$version" =~ '-' ]]; then \
+          repo='timescale/timescaledb-exp'; \
+          ghopts='--prerelease --draft'; \
+        else \
+          repo='timescale/timescaledb'; \
+          ghopts=''; \
+        fi;
         echo "::set-output name=version::${version}"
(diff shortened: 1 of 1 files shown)


jgpruitt (approved):

.github/workflows/release.yaml
```diff
@@ -56,10 +56,19 @@ jobs:
           tag="$(cargo metadata --format-version 1 --no-deps | jq -r '.packages[0].version')"; \
           version="${tag}-snapshot"; \
         fi;
+        if [[ "$version" =~ '-' ]]; then \
```
JamesGuthrie, line 59: I don't understand why this check tells us that it's a pre-release.

JamesGuthrie (changes requested):

.github/workflows/release.yaml
cevian: Because we follow server and pre-releases always follow Major.Minor.Patch-(pre-release tag) so something with a dash is a pre-release and something without isn't. Or am I missing something?

JamesGuthrie: Thanks, sounds good. I think I was the one missing something 😂

JamesGuthrie (approved):

sumerman (approved):

.github/workflows/release.yaml
paulfantom: Can we have that comment in code? It is not clear now, it won't be in the future.

paulfantom (approved):
Overall LGTM, just one nitpicky comment

PR #225, opened 2022-05-02
Rename tracing enums

cevian:
This makes the enum names much more user-friendly on the SQL level.
We made them lowercase and removed the prefix.

```
SELECT
  *
FROM ps_trace.span
WHERE start_time > NOW() - INTERVAL '30m'
AND status_code = 'STATUS_CODE_ERROR'
```

Now becomes

```
SELECT
  *
FROM ps_trace.span
WHERE start_time > NOW() - INTERVAL '30m'
AND status_code = 'error'
```


jgpruitt:
is there a corresponding PR on the connector-side?

jgpruitt (approved):

JamesGuthrie (approved):

PR #230, opened 2022-05-04
Run docker workflow on tag push

JamesGuthrie:


sumerman (approved):

pgnickb:
Could you please explain how this works in the commit message or somewhere? I genuinely lack competence to understand what is going on here.

JamesGuthrie:
> Could you please explain how this works in the commit message or somewhere? I genuinely lack competence to understand what is going on here.

Whoops, sorry I missed this before merging.

Github action workflows can be configured to be triggered on different events (documentation about this [here](https://docs.github.com/en/actions/using-workflows/triggering-a-workflow)). We had it configured so that the docker workflow, which builds and pushes our development docker images, would build on pull requests, and pushes to some branches, but not on pushing a new tag to the repo. In this change I added a trigger based on any tag. You can find a more detailed explanation of the syntax for `push` [here](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#push).

PR #234, opened 2022-05-04
dump/restore w/o superuser

jgpruitt:
### Ensure that a non-superuser can dump/restore the database

- updated build.rs to allow for a 999-last.sql file in the idempotent dir for tasks we ALWAYS want to run last
- matched some logic from cloud to allow for elevated privileges to the database owner to allow for dump/restore
- beefed up the logic in the snapshot to be more strict and cover more aspects of the database
- 013-remote-commands.sql and 014-apply-config.sql combined into 999-last.sql
- added code to 999-last.sql to ensure that prom_admin has privileges needed for dump/restore
- promscale_post_restore addresses ownership and privilege differences between a fresh install and a restore


e2e/scripts/after-create.sql
```diff
@@ -0,0 +1,54 @@
+\set ECHO all
+\set ON_ERROR_STOP 1
+
+-- create a user named bob
```
cevian, line 4: nit: does it make sense to use the name tsdbadmin instead of bob? Makes it more like cloud.

e2e/scripts/after-create.sql
```diff
@@ -0,0 +1,54 @@
...
+        create user bob login;
+    end if;
+end;
+$block$;
+alter database db owner to bob;
+grant all on database db to bob;
+
+create extension if not exists timescaledb with schema public;
+
+-- The pre_restore and post_restore function can only be successfully executed by a very highly privileged
```
cevian, line 30: Let's point to the script in the docker repo that this is emulating

e2e/tests/dump-restore-test.rs
```diff
@@ -288,47 +294,80 @@ fn dump_db(container: &PostgresContainer, db: &str, username: &str, path: &Path)
...
+        .args(["-d", db])
+        .arg("--no-password")
+        .arg("--no-psqlrc")
+        .arg("--no-readline")
+        .arg("--echo-all")
+        // do not ON_ERROR_STOP=1. some statements in the restore will fail. for example, setting
+        // the comment on the timescaledb extension. this is highly unfortunate, but seemingly
+        // unavoidable according to googling. eat the errors and continue. we will rely on the
+        // snapshot to determine whether the restore was successful
+        //.args(["-v", "ON_ERROR_STOP=1"]) don't uncomment me!
```
cevian, line 316: we should remember to document this. Can we get a list of expected errors for documentation?

migration/idempotent/011-maintenance.sql
```diff
@@ -648,6 +648,46 @@ BEGIN
...
+                INNER JOIN _timescaledb_catalog.hypertable h2
+                ON (h1.schema_name = '_ps_trace'
+                AND h1.table_name = ANY(ARRAY['span', 'link', 'event'])
+                AND h1.compressed_hypertable_id = h2.id)
+            )
+            SELECT format('GRANT ALL PRIVILEGES ON TABLE %I.%I TO prom_admin', x.schema_name, x.table_name)
+            FROM x
+            UNION ALL
+            SELECT format('GRANT ALL PRIVILEGES ON TABLE %I.%I TO %I', x.schema_name, x.table_name,
+                CASE
```
cevian, line 679: actually I don't think this method of determining user works well with pgextwlist. In other places I've changed this to just use session_user. Can we change this here as well?

migration/idempotent/999-last.sql
```diff
@@ -0,0 +1,271 @@
...
+
+    ALTER TABLE _ps_trace.span OWNER TO prom_admin; --required for restores to attach child tables
+    ALTER TABLE _ps_trace.link OWNER TO prom_admin; --required for restores to attach child tables
+    ALTER TABLE _ps_trace.event OWNER TO prom_admin; --required for restores to attach child tables
+    ALTER TABLE _prom_catalog.series OWNER TO prom_admin; --required for restores to attach child tables
+
+END$$;
+
+-- make sure prom_admin can do a dump/restore
+DO $$
```
cevian, line 229: hmm why do we need to do this, won't tsdbadmin have this already granted by the after-create scripts?

migration/incremental/023-privileges.sql
```diff
@@ -0,0 +1,14 @@
...
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _prom_catalog.remote_commands TO prom_admin;
+GRANT SELECT ON TABLE _ps_catalog.migration TO prom_reader;
+
+GRANT EXECUTE ON FUNCTION _prom_ext.num_cpus() TO prom_reader;
+GRANT EXECUTE ON FUNCTION _prom_ext.jsonb_digest(JSONB) TO prom_reader;
+GRANT EXECUTE ON FUNCTION _prom_ext.prom_delta(TIMESTAMPTZ, TIMESTAMPTZ, BIGINT, BIGINT, TIMESTAMPTZ, DOUBLE PRECISION) TO prom_reader;
+GRANT EXECUTE ON FUNCTION _prom_ext.prom_increase(TIMESTAMPTZ, TIMESTAMPTZ, BIGINT, BIGINT, TIMESTAMPTZ, DOUBLE PRECISION) TO prom_reader;
+GRANT EXECUTE ON FUNCTION _prom_ext.prom_rate(TIMESTAMPTZ, TIMESTAMPTZ, BIGINT, BIGINT, TIMESTAMPTZ, DOUBLE PRECISION) TO prom_reader;
+GRANT EXECUTE ON FUNCTION _prom_ext.vector_selector(TIMESTAMPTZ, TIMESTAMPTZ, BIGINT, BIGINT, TIMESTAMPTZ, DOUBLE PRECISION) TO prom_reader;
+GRANT EXECUTE ON FUNCTION _prom_ext.rewrite_fn_call_to_subquery(internal) TO prom_reader;
```
cevian, line 11: can't we do this in the idempotent scripts next to the function definition?

templates/idempotent-wrapper.sql
```diff
@@ -7,5 +7,6 @@ BEGIN
 -- and the comparison is indentation-sensitive.
 {{body}}
 
+RAISE LOG 'Applied idempotent script {{filename}}';
```
cevian, line 10: This may create too many messages?

cevian (commented):
Some questions

e2e/scripts/after-create.sql
jgpruitt: i can't believe you don't like "bob"!
jk. yeah, I'll change it

migration/idempotent/999-last.sql
jgpruitt: yes. on cloud. but not for folks using the docker images or the linux packages, right?

migration/incremental/023-privileges.sql
jgpruitt: i think the issue here is that these are defined by pgx, right @JamesGuthrie ?

e2e/tests/dump-restore-test.rs
jgpruitt: yep

templates/idempotent-wrapper.sql
jgpruitt: removed

migration/idempotent/011-maintenance.sql
```diff
@@ -648,6 +648,46 @@ BEGIN
...
+                    WHEN current_setting('server_version_num')::INT < 130000 THEN quote_ident(session_user)
+                    ELSE '@extowner@' -- proper quotation is handled by the pg itself
```
cevian, line 681: I think we should use session own in any pg version. Why? Because pgextwlist is sometimes used for 13 and 14 and this will break if it is.

migration/idempotent/999-last.sql
```diff
@@ -0,0 +1,271 @@
...
+    , 'prom_series'
+    , 'ps_tag'
+    , 'ps_trace'
+    ];
+BEGIN
+    -- grant schema privileges
+    FOREACH _schema IN ARRAY _schemas
+    LOOP
+        EXECUTE format('ALTER DEFAULT PRIVILEGES IN SCHEMA %I GRANT ALL PRIVILEGES ON TABLES TO prom_admin', _schema);
+        EXECUTE format('ALTER DEFAULT PRIVILEGES IN SCHEMA %I GRANT ALL PRIVILEGES ON SEQUENCES TO prom_admin', _schema);
```
cevian, line 197: I am not exactly sure whether we need to do this as it will only apply to future tables and can't we just apply that when we create new tables? DEFAULT PRIVILEGES seem to be a weird pain in the ass as it's only applied when tables are created by certain users. I wonder if it's best to avoid and be explicit at table creation time.

migration/idempotent/999-last.sql
cevian: Yeah but that seems like a timescaledb issue. In the ha-docker image it should work because after-create script is part of that. And on-prem people usually have superuser.

Fundamentally, I am a bit uncomfortable setting permissions on internal schemas of another extension even if it is timescaledb

cevian: I think for non-superuser dump restore we have to just target cloud

cevian (changes requested):

migration/idempotent/011-maintenance.sql
jgpruitt: gotcha.

migration/idempotent/999-last.sql
jgpruitt: > Fundamentally, I am a bit uncomfortable setting permissions on internal schemas of another extension even if it is timescaledb

yeah, agreed.

migration/idempotent/011-maintenance.sql
```diff
@@ -648,6 +648,122 @@ BEGIN
...
+                ON (h1.schema_name = ANY(ARRAY['prom_data', '_ps_trace'])
+                AND h1.compressed_hypertable_id = h2.id)
+            )
+            SELECT format('GRANT ALL PRIVILEGES ON TABLE %I.%I TO prom_admin', x.schema_name, x.table_name)
+            FROM x
+            UNION ALL
+            SELECT format('ALTER TABLE %I.%I OWNER TO prom_admin', x.schema_name, x.table_name)
+            FROM x
+            UNION ALL
+            SELECT format('GRANT ALL PRIVILEGES ON TABLE %I.%I TO %I', x.schema_name, x.table_name, quote_ident(session_user))
```
pgnickb, line 680: Using both `%I` and `quote_ident` will break whenever there is something to quote:
```SQL
select format('%I', quote_ident('&'));
┌─────────┐
│ format  │
├─────────┤
│ """&""" │
└─────────┘
(1 row)
```

migration/idempotent/011-maintenance.sql
```diff
@@ -648,6 +648,122 @@ BEGIN
...
+                INNER JOIN _timescaledb_catalog.chunk k
+                ON (h.id = k.hypertable_id)
+            )
+            SELECT format('GRANT ALL PRIVILEGES ON TABLE %I.%I TO prom_admin', x.schema_name, x.table_name)
+            FROM x
+            UNION ALL
+            SELECT format('ALTER TABLE %I.%I OWNER TO prom_admin', x.schema_name, x.table_name)
+            FROM x
+            UNION ALL
+            SELECT format('GRANT ALL PRIVILEGES ON TABLE %I.%I TO %I', x.schema_name, x.table_name, quote_ident(session_user))
```
pgnickb, line 759: This has the same issue as the previous one with `%I` and `quote_ident`. Perhaps worth fixing in case this is ever uncommented.

pgnickb (changes requested):
I realized I don't understand the `remote_commands` machinery, so I just trust it works and does what's promised.

My review was mostly sanity-checking the code.
I'd perhaps take on the task of tidying-up the formatting of SQL in some places to make it more reader friendly, but that is not urgent.

e2e/scripts/after-create.sql
```diff
@@ -0,0 +1,57 @@
...
+    select setting::int / 10000
+    into strict _version
+    from pg_settings where name = 'server_version_num';
+
+    -- trusted extensions were introduced in postgres v3
+    -- prior to that you must be a superuser to create extension
+    if _version < 13 then
+        create user tsdbadmin login superuser;
+    else
+        create user tsdbadmin login;
```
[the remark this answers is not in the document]
jgpruitt: fixed

migration/idempotent/011-maintenance.sql
```diff
@@ -648,6 +648,78 @@ BEGIN
             EXECUTE _sql;
         END LOOP
         ;
+
+        -- when restoring from backup, the user running the restore script ends up owning many of the objects
+        -- the following code addresses this by adjusting ownership and privileges to make them match what
+        -- they would have been had the database been created anew rather than restored
+
+        FOR _sql IN
```
cevian, line 656:
```suggestion

        --compressed hypertables
        FOR _sql IN
```

cevian (approved):

PR #216, opened 2022-04-29
WIP: Try out different rust caching techniques

JamesGuthrie:

PR #235, opened 2022-05-04
Improve build caching

JamesGuthrie:
This change integrates a number of improvements to caching in our build
pipelines. The results of different comparisons can be seen here: [1].

Overall, testing changes to caching is _hard_ because there are a lot of
different caching mechanisms in place. We already use caching in Docker
layers, and for "local" (not in Docker) Rust compilation. These existing
mechanisms can confound the result of a run.

The main change is integrating `sccache` into all aspects of the build.
We configure it to use an S3 bucket as the backing store for cached
artifacts, which allows the cache to be persisted over multiple CI runs.

The following is a breakdown of the major experiments and their results:

- ci.yaml

  This workflow consists primarily of Rust compilation directly on the
  github actions runner. Here can use the `swatinem/rust-cache`
  action to cache the intermediate compile outputs. We evaluated the
  build times of the following configurations:

  1. No caching at all                             14m23s [ci-1]
  2. Only `swatinem/rust-cache`                     5m46s [ci-2]
  3. Only `sccache`                                 9m16s [ci-3]
  4. Both `swatinem/rust-cache` and `sccache`       5m19s [ci-4]

- release.yaml

  This workflow consists primarily of Rust compilation inside a Docker
  container. This _can_ profit from Docker layering, but this is not
  always effective. Here we compared two configurations, where Docker
  layer caching was not in effect:

  1. No caching at all                             38m19s [r-1]
  2. `sccache` for Rust builds in Docker           22m50s [r-2]

We also ran both lint.yaml and docker.yaml in an "sccache" and
"no-sccache" setup to compare the differences. There the differences are
less pronounced, with `sccache` builds being _marginally_ slower.

[1]: https://github.com/timescale/promscale_extension/pull/216

[ci-1]: https://github.com/timescale/promscale_extension/actions/runs/2269259208
[ci-2]: https://github.com/timescale/promscale_extension/actions/runs/2269259211
[ci-3]: https://github.com/timescale/promscale_extension/actions/runs/2269259214
[ci-4]: https://github.com/timescale/promscale_extension/actions/runs/2269259212

[r-1]: https://github.com/timescale/promscale_extension/actions/runs/2269259218
[r-2]: https://github.com/timescale/promscale_extension/actions/runs/2269259202

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -16,6 +16,11 @@ jobs:
         - version: "14"
         - version: "13"
         - version: "12"
+    env:
+      RUSTC_WRAPPER: sccache
+      SCCACHE_BUCKET: promscale-extension-sccache
+      AWS_ACCESS_KEY_ID: ${{ secrets.PROMSCALE_EXTENSION_SCCACHE_AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.PROMSCALE_EXTENSION_SCCACHE_AWS_SECRET_ACCESS_KEY }}
     steps:
       - uses: actions/checkout@v2
       - name: Install latest nightly
@@ -23,6 +28,13 @@ jobs:
         with:
             toolchain: stable
             override: true
+
+      - name: Setup sccache
+        run: |
+          curl -L "https://github.com/mozilla/sccache/releases/download/v0.2.15/sccache-v0.2.15-x86_64-unknown-linux-musl.tar.gz" | tar zxf -
+          chmod +x sccache-*/sccache
+          sudo mv sccache-*/sccache /usr/local/bin/sccache
+          sccache --show-stats
 
       - uses: Swatinem/rust-cache@v1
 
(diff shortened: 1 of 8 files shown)


sumerman (commented):
Perhaps I'm missing something, but the diff between ci-2 and ci-4 doesn't seem significant enough to justify the added complexity for the `ci.yml` and `lint.yml`. 

Docker bits look good.

JamesGuthrie:
> Perhaps I'm missing something, but the diff between ci-2 and ci-4 doesn't seem significant enough to justify the added complexity for the `ci.yml` and `lint.yml`.

There is one case in which having `sccache` could be beneficial, although it's a bit contrived. If the file-cache is mostly invalidated (e.g. when changing Rust versions), then it will have to recompile a bunch of stuff. If the build fails 99% of the way through, everything which was already built will not have been cached in the file cache, but it would be cached in the `sccache` cache.

sumerman (approved):
I'm still not entirely convinced whether `sscache` is necessary for `ci` and `lint` workflows, but I don't think my hesitation should block this PR. 

I wonder what others think about the edge-case `sscache` addresses (see above).

jgpruitt (approved):

PR #238, opened 2022-05-05
Update extension comment

JamesGuthrie:
This was harder than I was expecting it to be. Open to other suggestions.

### DIFF
--- a/templates/promscale.control
+++ b/templates/promscale.control
@@ -1,5 +1,5 @@
 # promscale extension
-comment = 'Promscale support functions'
+comment = 'tables, types and functions supporting Promscale'
 default_version = '@CARGO_VERSION@'
 module_pathname = '$libdir/promscale'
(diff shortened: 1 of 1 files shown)


sumerman (approved):

jgpruitt (approved):
Not crazy about the description, but can't think of anything better except plain "Promscale".

JamesGuthrie:
@cevian any good ideas?

templates/promscale.control
```diff
@@ -1,5 +1,5 @@
 # promscale extension
-comment = 'Promscale support functions'
+comment = 'Promscale\'s database-side heavy lifter'
```
cevian, line 2:
```suggestion
comment = 'tables, types and functions supporting the Promscale project'
```

pgnickb (approved):

Issue #239, opened 2022-05-05
Provide packages for multiple releases


JamesGuthrie:
When we upload our packages to packagecloud, we tag them as either `debian/bullseye`, or `el/7`.

This means that, for debian-based distributions, the package can only be installed on bullseye by default (not any of focal/impish/jammy/etc.).

JamesGuthrie:
We managed to get this in for Debian and Ubuntu. For RHEL distributions there is more that we can do, but that work is being tracked in https://github.com/timescale/promscale_extension/issues/251.

PR #240, opened 2022-05-05
Provide packages for all debian-based distros, disable arm64

JamesGuthrie:
We are already building a .deb artifact, which we can publish for
multiple distros, so we just loop through all of the desired distros,
pushing the artifact to package cloud for that distro.

Currently coverage for RHEL-based distros is more "hope and pray" than
actually known to work. The CentOS-based build has been modified to test
the built rpm against the same CentOS version which built it. In future
this will be expanded, so the current implementation is a little hacky.

The arm64 build is run on a qemu VM on top of the github runner, so
takes 3+ hours to complete. This makes fixing any issues in the build an
annoyingly long game of whack-a-mole.

Fixes https://github.com/timescale/promscale_extension/issues/239


sumerman (commented):
I agree that we probably don't have to re-build the same thing for every version of distro, but publishing untested artefacts doesn't feel right. For instance, deb packages specify the version of PostgreSQL as their dependency. Even if all PG versions we build for are available for all distros in the list, this can change in the future without us noticing.  I imagine it should be possible to extend "Test Package" phase to include a run through all `pkg_cloud_tags`.

cevian:
I agree with @sumerman we do need to test all of these. The tests don't all need to run on all prs or branches but should run on some.

.github/workflows/release.yaml
```diff
@@ -120,30 +130,34 @@ jobs:
...
       env:
-        EXTENSION_DOCKER_IMAGE: ${{ steps.metadata.outputs.image }}-test
+        PG_VERSION: ${{ matrix.postgres.version }}
+        DOCKER_PLATFORM: ${{ matrix.os.name }}/${{ matrix.arch.platform }}
+        PACKAGE_TO_TEST: artifacts/${{ steps.metadata.outputs.outfile }}
+        IMAGE_BASE: ${{ steps.metadata.outputs.image }}-test
+        DOCKER_FILE: dist/${{ matrix.os.pkg_type }}.test.dockerfile
         CONNECTOR_DOCKER_IMAGE: "timescale/promscale:0.11.0-alpha"
-      run: ./tools/smoke-test "${EXTENSION_DOCKER_IMAGE}" "${CONNECTOR_DOCKER_IMAGE}" "${DOCKER_PLATFORM}"
+      run: |
```
sumerman, line 139: IMO this is script is too big to keep in GH action

dist/rpm.dockerfile
```diff
@@ -35,27 +35,20 @@ case "${OS_VERSION}" in
...
-            yum install -y rubygems ruby-devel
-        else
-            yum install -y epel-release rubygems ruby-devel
-        fi
+        yum install -y rubygems ruby-devel glibc-langpack-en glibc-locale-source
         ;;
 esac
 
 # Disable postgres
-if command -v dnf; then dnf -qy module disable postgresql; fi
```
sumerman, line 50: what was it and why we don't need it? 😅

Makefile
```diff
@@ -123,22 +130,20 @@ endif
 		.
 
 .PHONY: release-tester
-release-tester: dist/$(DOCKERFILE) ## Build image used for testing a specific release package
+release-tester: dist/$(TESTDOCKERFILE) ## Build image used for testing a specific release package
```
cevian, line 133: why the dependency on `dist/$(TESTDOCKERFILE)` those are committed, don't need to be built?

dist/rpm.dockerfile
cevian: I had the same question

dist/rpm.dockerfile
```diff
@@ -35,27 +35,20 @@ case "${OS_VERSION}" in
...
+
+EOF
 
+RUN <<EOF
 # System dependencies
 yum install -y \
     gettext \
     rpm-build \
-    rpm-devel \
-    rpmlint \
```
cevian, line 57: don't we need rpm lint if we are lifting aka this [line](https://github.com/timescale/promscale_extension/blob/develop/tools/package#L296) of tools/package?

dist/tester/timescale_timescaledb.repo
```diff
@@ -1,5 +1,6 @@
 [timescale_timescaledb]
 name=timescale_timescaledb
+baseurl=https://packagecloud.io/timescale/timescaledb/el/7/$basearch
```
cevian, line 3: why was this change needed.

cevian (approved):
Many of the same questions as Nick, overall looks good. None of my comments are blockers, so approving.

.github/workflows/release.yaml
JamesGuthrie: Agree. I would like to punt this as a tech-debt item to rework later. I will open an issue for it.

.github/workflows/release.yaml
```diff
@@ -120,30 +128,34 @@ jobs:
...
+        skip_arm64="debian/stretch ubuntu/hirsute ubuntu/impish"; \
```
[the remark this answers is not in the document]
JamesGuthrie: It's arm64-only, I will remove it from here for now.

dist/rpm.dockerfile
JamesGuthrie: Sorry this was a stray change which I made when fiddling with the rhel 8 build. I've reverted it.

dist/rpm.dockerfile
```diff
@@ -110,15 +101,9 @@ case "$(uname -m)" in
...
-if [ "${OS_NAME}" = "fedora" ]; then
-    PG_REPO="https://download.postgresql.org/pub/repos/yum/reporpms/F-${OS_VERSION}-${PG_ARCH}/pgdg-fedora-repo-latest.noarch.rpm"
-else
-    PG_REPO="https://download.postgresql.org/pub/repos/yum/reporpms/EL-${OS_VERSION}-${PG_ARCH}/pgdg-redhat-repo-latest.noarch.rpm"
-fi
-
-# Install supported Postgres versions
-yum install -y "${PG_REPO}"
+# Install supported Postgres versions \
+echo "https://download.postgresql.org/pub/repos/yum/reporpms/EL-${OS_VERSION}-${PG_ARCH}/pgdg-redhat-repo-latest.noarch.rpm"
```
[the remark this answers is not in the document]
JamesGuthrie: I've reverted this too.

dist/tester/timescale_timescaledb.repo
JamesGuthrie: This was previously dynamically added to the file in the docker build:

```
echo "baseurl=https://packagecloud.io/timescale/timescaledb/el/${OS_VERSION}/\$basearch" >> /etc/yum.repos.d/timescale_timescaledb.repo
```

The reason behind this was to support both centos 7 and 8 builds. We're no longer building centos 8, and we need to find another solution for rhel 8 anyway, so I would prefer to keep this not dynamic for now.

dist/rpm.dockerfile
JamesGuthrie: I've reverted this.

Makefile
JamesGuthrie: The docker command which is invoked by this stage does depend on the contents of the docker file. When testing this locally (and making changes to the dockerfile), one would want it to rebuild.

.github/workflows/release.yaml
JamesGuthrie: https://github.com/timescale/promscale_extension/issues/252

cevian (approved):

sumerman (approved):

PR #242, opened 2022-05-05
Correct the privileges on tracing tables

jgpruitt:
In the new tracing schema, prom_writer has select, insert, update, and delete.
It should only have select and insert.
Prom_modifier has no explicit privileges, but it should have select, insert, update, and delete.


pgnickb (approved):
LGTM

cevian (approved):

PR #245, opened 2022-05-06
More table privilege fixes

jgpruitt:


cevian (approved):
LGTM but let's run connector tests before we merge

PR #247, opened 2022-05-08
Introduce materialization for operations

cevian:
This commit introduces a materialization to record
the parent-child relation between operations. Namely for
and span of operation P that is the parent of a span with
operation C, we record the 10-minute bucketed relation:
- bucket time
- id of operation P
- id of operation C
- count()

This gives us a way to quickly aggregate the service map.
The materialization is done like a continous aggregate. But,
we couldn't use continuous aggs since they don't support
self-joins.

This materialization support real-time aggregation where the
materialization is combined with a freshly-calculated aggregation
of the not-yet materialized head of the data. This is available
in the _ps_trace.operation_stats view.

A limitation of this approach is that we don't support invalidations
so if a new span comes in for a time-period that has already been
materialized it is never corrected. Currently we materialize
data that is no newer than 10-minutes prior to the most
recent data.

A few assumptions are made in the materialization:
- That a child span's start_time is >= parent span's
  start time. This seem obvious but I couldn't find
  it documented in the spec.
- A child span's start_time is < parent span's start_time + 1 hour.
  This is more questionable. But I think it's ok if we
  document the limitation.

The background job to materialize this data is set up to be separate
from the rest of the jobs for isolation and also because we always
want to run this every 10 minutes and don't want parallelism.

Fixed parameters we may want to make configurable later:
- Size of bucket (currently 10 minutes)
- Size aggregated in one loop/txn (1 bucket)
- Invalidation window (10 minutes)
- chunk size of materialized hypertable (7 days)


pgnickb:
> A child span's start_time is < parent span's start_time + 1 hour.
> This is more questionable. But I think it's ok if we
> document the limitation.

> Fixed parameters we may want to make configurable later:

Perhaps worth adding the max parent age is worth it as well. We never know what the end system is going to be like. I reckon on a system where the delta of >1h is the case it wont be as significant of a performance penalty either.

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+    LIMIT 1;
+
+    --we materialize one bucket at a time until we are done. We do this so that
+    --we can make steady progress even if we are pretty far behind.
+    WHILE (_start + _bucket) <= _global_end LOOP
+        PERFORM _prom_catalog.set_app_name(pg_catalog.format('promscale maintenance: materializing operations: start %L', _start));
+        _query_start := clock_timestamp ();
+        INSERT INTO _ps_trace.operation_materialization
+        SELECT
+            public.time_bucket(_bucket, parent.start_time) bucket,
```
pgnickb, line 60:
```suggestion
            public.time_bucket(_bucket, parent.start_time) as bucket,
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+            parent.operation_id as parent_operation_id,
+            child.operation_id as child_operation_id,
+            count(*) as cnt
+        FROM
+            _ps_trace.span parent
```
pgnickb, line 65:
```suggestion
            _ps_trace.span as parent
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+        INNER JOIN
+            _ps_trace.span child ON (parent.span_id = child.parent_span_id AND parent.trace_id = child.trace_id)
```
pgnickb, line 67:
```suggestion
            _ps_trace.span as child ON (parent.span_id = child.parent_span_id AND parent.trace_id = child.trace_id)
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+    --bucket twice. Consider this if the "real-time" part of the query gets too
+    --expensive.
+    SELECT public.time_bucket(_bucket, start_time) - _bucket INTO _global_end
+    FROM _ps_trace.span
+    ORDER BY start_time DESC
+    LIMIT 1;
+
+    --we materialize one bucket at a time until we are done. We do this so that
+    --we can make steady progress even if we are pretty far behind.
+    WHILE (_start + _bucket) <= _global_end LOOP
```
pgnickb, line 55: I'm not certain this should be `<=` and not just `<`. Can you please elaborate on this?

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+    SELECT parent_start_time + interval '10 minutes' --time_bucket returns the lower end of the bucket so add an interval
+    FROM _ps_trace.operation_materialization
+    ORDER BY parent_start_time DESC
+    LIMIT 1
+$func$
+LANGUAGE SQL;
+
+--This procedure updates the materialization
+CREATE OR REPLACE PROCEDURE _ps_trace.execute_materialize_operations(log_verbose boolean = false, log_timing boolean = true)
+AS $func$
```
pgnickb, line 18:
```suggestion
SET search_path = pg_catalog
LANGUAGE plpgsql VOLATILE AS 
$func$
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+            END IF;
+        END IF;
+
+        --save our work!
+        COMMIT;
+        _start = _start + _bucket;
+    END LOOP;
+END
+$func$
+LANGUAGE PLPGSQL;
```
pgnickb, line 93:
```suggestion
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+                RAISE LOG 'materializing operations: from % to %', _start, _start + _bucket;
+            END IF;
+        END IF;
+
+        --save our work!
+        COMMIT;
+        _start = _start + _bucket;
+    END LOOP;
+END
+$func$
```
pgnickb, line 92:
```suggestion
$func$;
```

migration/idempotent/015-materialize_operations.sql
pgnickb: Let's keep the signature in the top of the function, so that it's more reader friendly and is easier to edit. (there is a corresponding edit at the bottom of the func)

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+LANGUAGE PLPGSQL;
+GRANT EXECUTE ON PROCEDURE _ps_trace.execute_materialize_operations(boolean, boolean) TO prom_maintenance;
+
+
+--This is the view that we can query. It's a "real time" view (like in continous aggregates)
+--which means it combines the materialized portion with a fresh rollup of the data from the
+--underlying table for the data that has not-yet-been-materialized. This means it should be
+--the same as a fresh rollup in all cases except invalidation (e.g. changes to data that
+--has already been materialized, which we don't handle yet).
+CREATE OR REPLACE VIEW _ps_trace.operation_stats AS --TODO: should this be public?
```
pgnickb, line 102: Is there any reason for this to exist if users can't benefit from it?

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+SELECT
+    parent_start_time as parent_start_time,
+    parent_operation_id as parent_operation_id,
+    child_operation_id as child_operation_id,
+    cnt bigint
+FROM  _ps_trace.operation_materialization
+WHERE parent_start_time < COALESCE(_ps_trace.materialize_watermark(), '-infinity'::timestamp with time zone)
```
pgnickb, line 109:
```suggestion
WHERE parent_start_time < COALESCE(_ps_trace.materialize_watermark(), '-infinity'::pg_catalog.timestamptz)
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+    public.time_bucket('10 minute', parent.start_time) parent_start_time,
+    parent.operation_id as parent_operation_id,
+    child.operation_id as child_operation_id,
+    count(*) as cnt
+FROM
+    _ps_trace.span parent
+INNER JOIN
+    _ps_trace.span child ON (parent.span_id = child.parent_span_id AND parent.trace_id = child.trace_id)
+WHERE
+    parent.start_time >= COALESCE(_ps_trace.materialize_watermark(), '-infinity'::timestamp with time zone)
```
pgnickb, line 121:
```suggestion
    parent.start_time >= COALESCE(_ps_trace.materialize_watermark(), '-infinity'::pg_catalog.timestamptz)
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+    --assumes the child span starts on or after the parent starts
+    AND child.start_time >= COALESCE(_ps_trace.materialize_watermark(), '-infinity'::timestamp with time zone)
```
pgnickb, line 123:
```suggestion
    AND child.start_time >= COALESCE(_ps_trace.materialize_watermark(), '-infinity'::pg_catalog.timestamptz)
```

migration/idempotent/015-materialize_operations.sql
```diff
@@ -0,0 +1,167 @@
...
+SELECT
+    parent_start_time as parent_start_time,
+    parent_operation_id as parent_operation_id,
+    child_operation_id as child_operation_id,
+    cnt bigint
+FROM  _ps_trace.operation_materialization
+WHERE parent_start_time < COALESCE(_ps_trace.materialize_watermark(), '-infinity'::timestamp with time zone)
+UNION ALL
+SELECT
+    public.time_bucket('10 minute', parent.start_time) parent_start_time,
```
pgnickb, line 112: Perhaps worth to put the interval into an in-linable config-function.

migration/incremental/023-materialize_operations.sql
```diff
@@ -0,0 +1,31 @@
+--need this index on span to search by start_time
+CREATE INDEX span_start_time ON _ps_trace.span(start_time);
+
+CREATE TABLE _ps_trace.operation_materialization (
+    parent_start_time timestamptz,
+    parent_operation_id bigint,
+    child_operation_id bigint,
+    cnt bigint,
+    --todo what else,
+    UNIQUE(parent_start_time, parent_operation_id, child_operation_id)
```
pgnickb, line 10: Perhaps worth upgrading to a `primary key` to enforce `NOT NULL`s.

migration/incremental/023-materialize_operations.sql
```diff
@@ -0,0 +1,31 @@
...
+CREATE OR REPLACE PROCEDURE _ps_trace.execute_materialize_operations_job(job_id int, config jsonb)
+AS $$
+BEGIN
+END
+$$ LANGUAGE PLPGSQL;
+
+--keep this separate from the other jobs for independence and also since the right
+--schedule here is always "ever 10 min". We never want this to run in parallel so
+--it's always just one job making progress. This job is usally pretty quiet so enable
+--logging by default (should chirp once every 10 min).
```
pgnickb, line 30: Not sure I agree with this. It is quiet indeed, but what benefit does it provide? Perhaps it's better to expose the duration (or the end_time) in the job table (`_ps_trace.operation_materialization`) so that the user can assess the performance change of this over time and consider changing the bucket size.
If this is a file log entry then I'd see this as yet another line I wouldn't care about, since a single entry doesn't give you any information at all.

PR #253, opened 2022-05-10
Do trace compression through our own functions

cevian:
Move trace compression to our functions rather than policy.
There are 2 reasons:
- We can't run compression policy on tables owned by prom_admin
  because that user doesn't have LOGIN rights and timescale requires
  those rights.
- This allows us better control over when chunks are compressed. Our
  current function requires the chunk not to be the last chunk AND
  be an hour long.


jgpruitt (approved):
The dump/restore test is failing for some reason, but it LGTM otherwise.

pgnickb:
Could you please elaborate on the autoexplain stuff? Does it provide any value for this workload?

pgnickb (approved):

JamesGuthrie (approved):

sumerman (approved):

PR #258, opened 2022-05-10
Support for reuse temp table for trace ingest 

sumerman:
Creates a temporary table (if it doesn't exist), suitable for tracing data ingestion.
Suppresses corresponding DDL logging, otherwise, the PG log may get unnecessarily verbose.

Addresses https://github.com/timescale/promscale/pull/1333

### DIFF
--- a/migration/idempotent/008-tracing-functions.sql
+++ b/migration/idempotent/008-tracing-functions.sql
@@ -569,6 +569,31 @@ $func$
 LANGUAGE plpgsql;
 GRANT EXECUTE ON FUNCTION ps_trace.put_instrumentation_lib(text, text, bigint) TO prom_writer;
 
+-- Creates a temporary table (if it doesn't exist), suitable for tracing data ingestion.
+-- Supresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
+CREATE OR REPLACE FUNCTION _ps_trace.ensure_trace_ingest_temp_table(_temp_table_name text, _proto_table_name text)
+    RETURNS void
+    SECURITY DEFINER
+    VOLATILE
+    SET search_path = pg_catalog, pg_temp
+AS $func$
+BEGIN
+    IF _proto_table_name NOT IN ('span', 'link', 'event') THEN
+        RAISE EXCEPTION 'Invalid table prototype specified: %', _proto_table_name;
+    END IF;
+
+    SET LOCAL log_statement = 'none';
+
+    EXECUTE format($sql$CREATE TEMPORARY TABLE IF NOT EXISTS %I ON COMMIT DELETE ROWS AS TABLE _ps_trace.%I WITH NO DATA$sql$,
+                    _temp_table_name, _proto_table_name);
+    EXECUTE format($sql$GRANT SELECT, INSERT ON TABLE %I TO prom_writer$sql$,
+                    _temp_table_name);
+END;
+$func$
+LANGUAGE plpgsql;
+REVOKE ALL ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) TO prom_writer;
+
 CREATE OR REPLACE FUNCTION ps_trace.delete_all_traces()
     RETURNS void
     SECURITY DEFINER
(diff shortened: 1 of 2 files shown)


migration/idempotent/008-tracing-functions.sql
```diff
@@ -569,6 +569,39 @@ $func$
...
+        RAISE EXCEPTION 'Invalid table prototype specified: %', _proto_table_name;
+    END IF;
+
+    _log_setting = current_setting('log_statement');
+    SET log_statement = 'none';
+
+    EXECUTE format($sql$CREATE TEMPORARY TABLE IF NOT EXISTS %I ON COMMIT DELETE ROWS AS TABLE _ps_trace.%I WITH NO DATA$sql$,
+                    _temp_table_name, _proto_table_name);
+
+    IF _log_setting = 'none' THEN
```
pgnickb, line 593: could be simplified into `set_config`, eliminating IFs

migration/idempotent/008-tracing-functions.sql
```diff
@@ -569,6 +569,39 @@ $func$
...
+    _log_setting text;
+BEGIN
+    IF _proto_table_name NOT IN ('span', 'link', 'event') THEN
+        RAISE EXCEPTION 'Invalid table prototype specified: %', _proto_table_name;
+    END IF;
+
+    _log_setting = current_setting('log_statement');
+    SET log_statement = 'none';
+
+    EXECUTE format($sql$CREATE TEMPORARY TABLE IF NOT EXISTS %I ON COMMIT DELETE ROWS AS TABLE _ps_trace.%I WITH NO DATA$sql$,
```
pgnickb, line 590: Any reason we actually avoid using `CREATE TABLE x(LIKE y);`?

pgnickb (approved):

migration/idempotent/008-tracing-functions.sql
sumerman: Does it have the exact same behaviour? Why didn't you use it in your sample then?

migration/idempotent/008-tracing-functions.sql
sumerman: I've copied @jgpruitt code. Is there a difference?

migration/idempotent/008-tracing-functions.sql
sumerman: actually I should just use `SET LOCAL`

migration/idempotent/008-tracing-functions.sql
```diff
@@ -569,6 +569,28 @@ $func$
...
+    END IF;
+
+    SET LOCAL log_statement = 'none';
+
+    EXECUTE format($sql$CREATE TEMPORARY TABLE IF NOT EXISTS %I ON COMMIT DELETE ROWS AS TABLE _ps_trace.%I WITH NO DATA$sql$,
+                    _temp_table_name, _proto_table_name);
+END;
+$func$
+LANGUAGE plpgsql;
+GRANT EXECUTE ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) TO prom_writer;
```
JamesGuthrie, line 592: I think that generally it's good practice to revoke exec from PUBLIC on security definer functions.

JamesGuthrie (approved):

cevian (approved):

migration/idempotent/008-tracing-functions.sql
jgpruitt: That code was actually there already. I just modified it from `ON COMMIT DROP` to `ON COMMIT DELETE ROWS`.

This is interesting. This code is using the `CREATE TABLE AS` [statement](https://www.postgresql.org/docs/current/sql-createtableas.html) combined with the `TABLE` [statement](https://www.postgresql.org/docs/current/sql-select.html#SQL-TABLE). `TABLE foo` == `select * from foo`.

The `CREATE TABLE LIKE` statement would do essentially the same thing without running a `select * from foo`. Although, it looks like the LIKE route also copies not null constraints and the CREATE TABLE AS doesn't appear to.

jgpruitt: I think we should change this to use the LIKE approach. Running a needless `SELECT *` seems like a bad idea.

jgpruitt (changes requested):

jgpruitt (approved):

migration/idempotent/008-tracing-functions.sql
pgnickb: I didn't use it because mine had a more broad context and I decided to completely avoid using `format` or the friends. In this context it is less relevant, as the target value doesn't come from the user.

PR #262, opened 2022-05-11
Fix takeover script to handle tracing data

cevian:


sumerman (approved):

pgnickb (approved):
I don't see any obvious issues with this. It'd be nice to have some more context for the review, so that I could also do a logical sanity check. From the SQL and Regex point of view it looks good

Harkishen-Singh (approved):

PR #266, opened 2022-05-11
Merge release 0.5.0 into master

cevian:


JamesGuthrie (changes requested):
Can you please add the generated sql schema file to this commit with `cargo pgx schema --out sql/promscale--0.5.0.sql`.

sumerman (approved):

cevian:
@JamesGuthrie done but can't say I loved it since we didnt do that for the alphas

JamesGuthrie:
> @JamesGuthrie done but can't say I loved it since we didnt do that for the alphas

Ack. If we don't do this now, the`promscale--0.5.0.sql` file will be "lost to the ether" (ish, we could always get it back). We may also want to commit the upgrade script symlinks, but it's too late for me to pass judgement on that, and we can quite easily do that retroactively.

Also: holy large file batman.

JamesGuthrie (approved):
LGTM. 🤞

jgpruitt (approved):

PR #272, opened 2022-05-18
Ensure that frozen incremental files are not modified

jgpruitt:
The sql scripts in migration/incremental/ are not idempotent. They are meant to be run once
and only once on any given promscale installation. We track whether these files have
been applied in the _ps_catalog.migration table. Once we have released a version of
promscale, all the incremental scripts are considered frozen. They have been applied to
databases in the wild, and we don't want to create a situation in which some databases were
created using one version of a script, while others used a different version.
This test is meant to keep us from accidentally changing an incremental file after it has
been released. When cutting a new release, append to the list of files below the filenames
and MD5 hashes of all the new files going out with that release.

### DIFF
--- a/e2e/Cargo.toml
+++ b/e2e/Cargo.toml
@@ -11,6 +11,8 @@ testcontainers = "0.12.0"
 test-generator = "0.3.0"
 similar = "2.1.0"
 regex = "1.5.5"
+md-5 = "0.10.1"
+hex-literal = "0.3.4"
 
 [build-dependencies]
 build-deps = "^0.1"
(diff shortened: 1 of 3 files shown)


e2e/Cargo.toml
```diff
@@ -11,6 +11,8 @@ testcontainers = "0.12.0"
 test-generator = "0.3.0"
 similar = "2.1.0"
 regex = "1.5.5"
+md-5 = "0.10.1"
```
sumerman, line 14: It's probably fine, but I'd suggest using SHA512

PR #273, opened 2022-05-19
More optimal query plan when generating dynamic SQL

niksajakovljevic:
Somehow PG planner was picking non-optimal plan for a given function (
performing seq scan instead of index only scan) which resulted in a
really poor performance. Generating and executing dynamic SQL generates optimal plan.

Old plan:
```
   Aggregate  (cost=298493.79..298494.05 rows=1 width=128) (actual time=2816.778..2816.780 rows=1 loops=1)
          CTE cte
            ->  Result  (cost=298490.80..298493.43 rows=10 width=72) (actual time=2816.760..2816.764 rows=13 loops=1)
                  ->  Sort  (cost=298490.80..298490.83 rows=10 width=72) (actual time=2816.759..2816.761 rows=13 loops=1)
                        Sort Key: kv.key COLLATE "C", kv.value COLLATE "C"
                        Sort Method: quicksort  Memory: 26kB
                        ->  Merge Left Join  (cost=298490.44..298490.63 rows=10 width=72) (actual time=2816.748..2816.755 rows=13 loops=1)
                              Merge Cond: (kv.key = lkp.key)
                              ->  Sort  (cost=298465.14..298465.16 rows=10 width=68) (actual time=2816.657..2816.659 rows=13 loops=1)
                                    Sort Key: kv.key COLLATE "C"
                                    Sort Method: quicksort  Memory: 25kB
                                    ->  Hash Right Join  (cost=0.26..298464.97 rows=10 width=68) (actual time=0.031..2816.648 rows=13 loops=1)
                                          Hash Cond: ((l.key = kv.key) AND (l.value = kv.value))
                                          ->  Seq Scan on label l  (cost=0.00..223462.26 rows=10000326 width=68) (actual time=0.006..969.334 rows=10000207 loops=1)
                                          ->  Hash  (cost=0.11..0.11 rows=10 width=64) (actual time=0.018..0.018 rows=13 loops=1)
                                                Buckets: 1024  Batches: 1  Memory Usage: 9kB
                                                ->  Function Scan on kv  (cost=0.01..0.11 rows=10 width=64) (actual time=0.010..0.012 rows=13 loops=1)
                              ->  Sort  (cost=25.30..25.34 rows=13 width=23) (actual time=0.087..0.088 rows=16 loops=1)
                                    Sort Key: lkp.key COLLATE "C"
                                    Sort Method: quicksort  Memory: 26kB
                                    ->  Seq Scan on label_key_position lkp  (cost=0.00..25.06 rows=13 width=23) (actual time=0.010..0.076 rows=16 loops=1)
                                          Filter: (metric_name = $1)
                                          Rows Removed by Filter: 975
          ->  CTE Scan on cte  (cost=0.00..0.20 rows=10 width=72) (actual time=2816.762..2816.767 rows=13 loops=1)
```

New plan:
```Result  (cost=28.88..35.57 rows=13 width=72) (actual time=0.109..0.111 rows=13 loops=1)
   ->  Sort  (cost=28.88..28.91 rows=13 width=72) (actual time=0.107..0.108 rows=13 loops=1)
         Sort Key: kv.key, kv.value
         Sort Method: quicksort  Memory: 26kB
         ->  Hash Left Join  (cost=6.09..28.64 rows=13 width=72) (actual time=0.051..0.096 rows=13 loops=1)
               Hash Cond: (kv.key = lkp.key)
               ->  Nested Loop Left Join  (cost=0.57..23.08 rows=13 width=68) (actual time=0.030..0.071 rows=13 loops=1)
                     ->  Function Scan on kv  (cost=0.01..0.14 rows=13 width=64) (actual time=0.008..0.009 rows=13 loops=1)
                     ->  Index Only Scan using label_key_value_id_key on label l  (cost=0.56..1.76 rows=1 width=70) (actual time=0.004..0.004 rows=1 loops=13)
                           Index Cond: ((key = kv.key) AND (value = kv.value))
                           Heap Fetches: 13
               ->  Hash  (cost=5.48..5.48 rows=4 width=36) (actual time=0.016..0.016 rows=13 loops=1)
                     Buckets: 1024  Batches: 1  Memory Usage: 9kB
                     ->  Bitmap Heap Scan on label_key_position lkp  (cost=1.28..5.48 rows=4 width=36) (actual time=0.006..0.007 rows=13 loops=1)
                           Recheck Cond: (metric_name = 'avalanche_metric_mmmmm_0_0'::text)
                           Heap Blocks: exact=1
                           ->  Bitmap Index Scan on label_key_position_metric_name_key_pos_key  (cost=0.00..1.28 rows=4 width=0) (actual time=0.003..0.003 rows=13 loops=1)
                                 Index Cond: (metric_name = 'avalanche_metric_mmmmm_0_0'::text)
```

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -1106,36 +1106,40 @@ CREATE OR REPLACE FUNCTION _prom_catalog.get_or_create_label_ids(metric_name TEX
     VOLATILE
     SET search_path = pg_catalog, pg_temp
 AS $$
-        WITH cte as (
-        SELECT
-            -- only call the functions to create new key positions
-            -- and label ids if they don't exist (for performance reasons)
-            lkp.pos as known_pos,
-            coalesce(l.id, _prom_catalog.get_or_create_label_id(kv.key, kv.value)) label_id,
-            kv.key key_str,
-            kv.value val_str
-        FROM ROWS FROM(unnest(label_keys), UNNEST(label_values)) AS kv(key, value)
-            LEFT JOIN _prom_catalog.label l
-               ON (l.key = kv.key AND l.value = kv.value)
-            LEFT JOIN _prom_catalog.label_key_position lkp ON
-            (
-                    lkp.metric_name = get_or_create_label_ids.metric_name AND
-                    lkp.key = kv.key
+    BEGIN
+        RETURN QUERY EXECUTE format($query$
+        WITH cte AS (
+            SELECT
+                -- only call the functions to create new key positions
+                -- and label ids if they don't exist (for performance reasons)
+                lkp.pos as known_pos,
+                coalesce(l.id, _prom_catalog.get_or_create_label_id(kv.key, kv.value)) label_id,
+                kv.key key_str,
+                kv.value val_str
+            FROM ROWS FROM(unnest(%3$L::text[]), unnest(%4$L::text[])) AS kv(key, value)
+                LEFT JOIN _prom_catalog.label l
+                ON (l.key = kv.key AND l.value = kv.value)
+                LEFT JOIN _prom_catalog.label_key_position lkp ON
+                (
+                        lkp.metric_name = %1$L AND
+                        lkp.key = kv.key
+                )
+            ORDER BY kv.key, kv.value
             )
-        ORDER BY kv.key, kv.value
-        )
-        SELECT
-           case when count(*) = count(known_pos) Then
-              array_agg(known_pos)
-           else
-              _prom_catalog.get_new_pos_for_key(get_or_create_label_ids.metric_name, get_or_create_label_ids.metric_table, array_agg(key_str), false)
-           end as poss,
-           array_agg(label_id) as label_ids,
-           array_agg(key_str) as keys,
-           array_agg(val_str) as vals
-        FROM cte
+            SELECT
+            case when count(*) = count(known_pos) Then
+                array_agg(known_pos)
+            else
+                _prom_catalog.get_new_pos_for_key(%1$L, %2$L, array_agg(key_str), false)
+            end as poss,
+            array_agg(label_id) as label_ids,
+            array_agg(key_str) as keys,
+            array_agg(val_str) as vals
+            FROM cte
+        $query$, metric_name, metric_table, label_keys, label_values);
+    END
 $$
-LANGUAGE SQL;
+LANGUAGE PLPGSQL;
 COMMENT ON FUNCTION _prom_catalog.get_or_create_label_ids(text, name, text[], text[])
 IS 'converts a metric name, array of keys, and array of values to a list of label ids';
 GRANT EXECUTE ON FUNCTION _prom_catalog.get_or_create_label_ids(TEXT, NAME, text[], text[]) TO prom_writer;


migration/incremental/025-change-lable-ids-func.sql
```diff
@@ -0,0 +1,39 @@
+CREATE OR REPLACE FUNCTION _prom_catalog.get_or_create_label_ids(metric_name TEXT, metric_table NAME, label_keys text[], label_values text[])
```
JamesGuthrie, line 1: You don't need to redefine this function in incremental.

migration/idempotent/001-base.sql
```diff
@@ -1106,36 +1106,40 @@ CREATE OR REPLACE FUNCTION _prom_catalog.get_or_create_label_ids(metric_name TEX
...
+            end as poss,
+            array_agg(label_id) as label_ids,
+            array_agg(key_str) as keys,
+            array_agg(val_str) as vals
+            FROM cte
+        $query$, metric_name, metric_table, label_keys, label_values);
+    END
 $$
-LANGUAGE SQL;
+LANGUAGE PLPGSQL VOLATILE;
```
JamesGuthrie, line 1142: Tests are failing because `VOLATILE` here is redundant (it was already declared before the function body).

```suggestion
LANGUAGE PLPGSQL;
```

migration/idempotent/001-base.sql
```diff
@@ -1106,36 +1106,40 @@ CREATE OR REPLACE FUNCTION _prom_catalog.get_or_create_label_ids(metric_name TEX
...
-            kv.value val_str
-        FROM ROWS FROM(unnest(label_keys), UNNEST(label_values)) AS kv(key, value)
-            LEFT JOIN _prom_catalog.label l
-               ON (l.key = kv.key AND l.value = kv.value)
-            LEFT JOIN _prom_catalog.label_key_position lkp ON
-            (
-                    lkp.metric_name = get_or_create_label_ids.metric_name AND
-                    lkp.key = kv.key
+    BEGIN
+        RETURN QUERY EXECUTE format($query$
```
JamesGuthrie, line 1110: I know that there is a lot of "black magic" around the query planner and optimising queries, but I have a few issues:

1. The fact that we've restructured this code for a specific outcome is not evident either in the code (e.g. as a comment), or in the tests (to assert that we're actually getting the desired query plan).
2. Based on the description of the commit, I don't understand why this works, or why this is the best way to achieve the desired outcome.

JamesGuthrie (changes requested):

migration/idempotent/001-base.sql
niksajakovljevic: 1.  While testing I've tried turning off generic plan `set plan_cache_mode=force_custom_plan;` and also forcing planner to not use sequential scan `set enable_seqscan=off`, however non of that helped 🤷 . It seems that somehow planner gets confused when planning a query that uses arguments. The PR description explains the difference about query plans. If you feel strongly about it I could add code comment about it as well?
Making tests for function plan is a bit tricky. The `EXPLAIN` of a function call does not show inner query plan (like it does for a normal SQL query). The only way (that I am aware of) to get it is to use `auto_explain` library that will print execution plan to PG log...so it requires quite a bit of work...
2. Not really sure what part is confusing. I shared two plans - one that does sequential scan and another that does the index scan. Index scan is almost always preferred (unless there is a very little data to scan). Our benchmarks are showing from 10X-1000X improvement (depending on number of labels) for this fix.
What else do you feel we should add to make it more clear?

cevian (changes requested):
I think there is another solution. will discuss on slack

niksajakovljevic:
It seems that the root cause is that NAME type uses C collation which confuses planner. I've opened another PR to tackle this for all of our functions https://github.com/timescale/promscale_extension/pull/310

PR #289, opened 2022-05-23
Optimize metric view performance

jgpruitt:


JamesGuthrie (changes requested):
Can we split out this change into (at least two) component pieces?

Right now I see three pieces:
1. Fixup `NAME -> TEXT` change
2. Drop old TSDB compatibility
3. Optimize metric view

I would really like to get 1. and 2. in ASAP, and from what I understand, it's not clear if we're happy with the current status of 3., so that might slow things down a bit.

jgpruitt:
see this PR instead https://github.com/timescale/promscale_extension/pull/334

Issue #274, opened 2022-05-19
Missing upgrade script for 0.3.2 to 0.5.0


tomhughes:
Promscale 0.11.0 is refusing to start claiming the extension is too old even though I have 0.5.0 installed. Trying to update it manually fails:

```
promscale=# alter extension promscale update;
ERROR:  extension "promscale" has no update path from version "0.3.2" to version "0.5.0"
```

and the package does indeed appear to be missing an SQL script to do the update, as does this repository.

sumerman:
Hi Tom, which one of the packages did you use to install the extension? 

The SQL script is not present in the repository, because it’s generated in the build process.

tomhughes:
The deb from the timescaledb repository at https://packagecloud.io/timescale/timescaledb/ubuntu.

tomhughes:
The github doesn't seem to have it either:

```
% dpkg --contents promscale-extension-0.5.0.pg14.debian11.x86_64.deb
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/lib/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/lib/postgresql/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/lib/postgresql/14/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/lib/postgresql/14/lib/
-rwxr-xr-x 0/0         1267968 2022-05-12 01:21 ./usr/lib/postgresql/14/lib/promscale-0.5.0.so
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/share/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/share/postgresql/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/share/postgresql/14/
drwxr-xr-x 0/0               0 2022-05-12 01:21 ./usr/share/postgresql/14/extension/
-rw-r--r-- 0/0           11063 2022-05-12 01:21 ./usr/share/postgresql/14/extension/promscale--0.3.2.sql
-rw-r--r-- 0/0           48512 2022-05-12 01:21 ./usr/share/postgresql/14/extension/promscale--0.0.0.sql
[...]

PR #290, opened 2022-05-23
Stop all background workers for the duration of the upgrade

cevian:
This commit stops bgws while upgrading the Promscale extension.
This prevents lock contention between jobs and upgrades and thus
seems safer.

### DIFF
--- a/migration/bootstrap/002-stop-bgw.sql
+++ b/migration/bootstrap/002-stop-bgw.sql
@@ -1,0 +1,25 @@
+-- stop all background workers in the current db so they don't interfere
+-- with the upgrade (by e.g. holding locks).
+--
+-- Note: this stops the scheduler (and thus all jobs)
+-- and waits for the current txn (the one doing the upgrade) to finish
+-- before starting it up again. Thus, no jobs will be running during
+-- the upgrade.
+
+-- TimescaleDB itself does the same thing:
+-- https://github.com/timescale/timescaledb/blob/72d03e6f7d30cc4794c9263445f14199241e2eb5/sql/updates/pre-update.sql#L34
+DO
+$stop_bgw$
+DECLARE
+    _is_timescaledb_installed boolean = false;
+BEGIN
+    SELECT count(*) > 0
+    INTO STRICT _is_timescaledb_installed
+    FROM pg_extension
+    WHERE extname='timescaledb';
+
+    IF _is_timescaledb_installed THEN
+        PERFORM _timescaledb_internal.restart_background_workers();
+    END IF;
+END;
+$stop_bgw$;
--- a/build.rs
+++ b/build.rs
@@ -102,9 +102,12 @@ fn render_bootstrap(manifest_dir: &str) -> String {
     ext_schema_path.push("migration/bootstrap/001-create-ext-schema.sql");
     let ext_schema_sql = wrap(ext_schema_path.as_path(), &Incremental);
 
+    let stop_bgw_sql = include_str!("migration/bootstrap/002-stop-bgw.sql").to_string();
+
     let mut result = String::from("");
     result.push_str(&migration_table_sql);
     result.push_str(&ext_schema_sql);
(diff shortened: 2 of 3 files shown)


cevian:
This will address some of the potential slowness seen in https://github.com/timescale/promscale_extension/issues/274

pgnickb (dismissed):

migration/bootstrap/002-stop-bgw.sql
```diff
@@ -0,0 +1,12 @@
...
+-- with the upgrade (by e.g. holding locks).
+--
+-- Note: this stops the scheduler (and thus all jobs)
+-- and waits for the current txn (the one doing the upgrade) to finish
+-- before starting it up again. Thus, no jobs will be running during
+-- the upgrade.
+
+-- TimescaleDB itself does the same thing:
+-- https://github.com/timescale/timescaledb/blob/72d03e6f7d30cc4794c9263445f14199241e2eb5/sql/updates/pre-update.sql#L34
+SELECT _timescaledb_internal.restart_background_workers();
```
JamesGuthrie, line 11: This doesn't work without timescaledb. I know that we're planning on not supporting installations without timescaledb, but that will also require improving our test infra at least a little. I would suggest the following:

```diff
+-- stop all background workers in the current db so they don't interfere
+-- with the upgrade (by e.g. holding locks).
+--
+-- Note: this stops the scheduler (and thus all jobs)
+-- and waits for the current txn (the one doing the upgrade) to finish
+-- before starting it up again. Thus, no jobs will be running during
+-- the upgrade.
+
+-- TimescaleDB itself does the same thing:
+-- https://github.com/timescale/timescaledb/blob/72d03e6f7d30cc4794c9263445f14199241e2eb5/sql/updates/pre-update.sql#L34
+DO
+$stop_bgw$
+DECLARE
+    _is_timescaledb_installed boolean = false;
+BEGIN
+    SELECT count(*) > 0
+    INTO STRICT _is_timescaledb_installed
+    FROM pg_extension
+    WHERE extname='timescaledb';
+
+    IF _is_timescaledb_installed THEN
+        SELECT _timescaledb_internal.restart_background_workers();
+    END IF;
+END;
+$stop_bgw$;
```

JamesGuthrie (changes requested):
The background workers should also be stopped at the beginning of `promscale--0.0.0.sql`.

JamesGuthrie:
Implementation of #291

pgnickb:
Sorry for a stupid question, but I failed to get the answer by just glancing through the code of `ts_bgw_db_workers_restart`.. Is it blocking or does it always return immediately upon sending the message? And more relevantly, do we want to set a `statement_timeout` for this operation in case it takes too long?

JamesGuthrie:
> Sorry for a stupid question, but I failed to get the answer by just glancing through the code of `ts_bgw_db_workers_restart`.. Is it blocking or does it always return immediately upon sending the message? And more relevantly, do we want to set a `statement_timeout` for this operation in case it takes too long?

It looks like it _can_ be blocking, if the message queue to the background worker process is full. The comment on `ts_bgw_message_send_and_wait` (called by `ts_bgw_db_workers_restart`) says:
```c
/*
 * Write element to queue, wait/error if queue is full
 * consumes message and deallocates
 */
extern bool
ts_bgw_message_send_and_wait(BgwMessageType message_type, Oid db_oid)
```

@cevian On another note, I tested this change out in my test environment, by adding the "stop bgw" snippet to the `promscale--0.0.0.sql` script. It stopped the background workers, and they were run again when the timeout was reached.

JamesGuthrie:
I also added a commit to make CI pass. Feel free to squash it down.

JamesGuthrie (approved):

pgnickb (approved):

JamesGuthrie:
@cevian are you waiting on anything to merge this?

cevian:
Connector test https://github.com/timescale/promscale/actions/runs/2457393370

PR #313, opened 2022-05-26
Fix compile from source

### LINKED ISSUE #311: Instructions to compile from source broken
The instructions say:

```
make
sudo make install
```

These instructions are broken.

JamesGuthrie:
- We forgot to update the version in the documentation

- The installation broke between 0.3.0 and 0.5.0. The reason is that the
  recommended way to install Rust/rustup is in a per-user directory.
  This means that invoking anything via sudo doesn't have Rust available
  to it. Instead of trying to (effectively) `sudo cargo ...`, separate
  out `make` and `make install`, so that `sudo make install` works.

Fixes #311

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,13 @@
 ARCH ?= $(shell uname -m)
 # This little workaround allows usage of the '#' character on both gnu make 3.8 and 4.3+
 H := \#
-EXT_VERSION ?= $(shell cargo pkgid | cut -d'$H' -f2 | cut -d':' -f2)
+# We want to be able to execute `sudo make install`, but cargo and rustc are probably both
+# not available for sudo, so we use `command -v` to only run them if they are available
+EXT_VERSION ?= $(shell command -v cargo >/dev/null && cargo pkgid | cut -d'$H' -f2 | cut -d':' -f2)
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -97,10 +99,8 @@ package: promscale.control ## Generate extension artifacts for packaging
 	cargo pgx package --pg-config ${PG_CONFIG}
 
 .PHONY: install
-install: promscale.control ## Install the extension in the Postgres found via pg_config
-	cargo pgx schema pg${PG_BUILD_VERSION} --out sql/promscale--${EXT_VERSION}.sql --release
-	bash create-upgrade-symlinks.sh
-	cargo pgx install --pg-config ${PG_CONFIG}
+install: ## Install the extension in the Postgres found via pg_config
```
sumerman, line 102: don't we need to build it first?

JamesGuthrie: Yes, but `make install` only installs. You have to run `make package` first (as in the instructions).

sumerman: Is there a downside to making a dependency chain like: `install` -> `/target/release/promscale-pg${PG_BUILD_VERSION}` -> `package`? Non-sudo `make install` is sometimes useful for local development.

sumerman (approved):

Issue #288, opened 2022-05-23
Storage parameter of tag_map is set incorrectly.


cevian:
Tag_map have a plain storage type, which means they cannot be toasted. This can lead to "row is too big" errors as reported in https://timescaledb.slack.com/archives/C011FC0PPC5/p1653314379747189

Note JSONB itself is marked as extended and so the setting for tag_map was an oversight.

This needs to be fixed by:

- [ ] ALTER TYPE on tag_map to change it's storage type (probably `extended` or `main`)
- [ ] ALTER TABLE on any tables that use tag_map to change the storage type of the associated columns.

Would be good to add a test. I think this is triggered if there are a lot of tags associated with a span.

sumerman:
Even though, we should indeed change `tag_map` storage type, the real offender is likely `tag_v`

PR #314, opened 2022-05-26
Fixing storage of tag_map and tag_v types

sumerman:
Addresses #288


migration/incremental/025-tag-map-storage-type.sql
```diff
@@ -0,0 +1,19 @@
+-- tag_v
+UPDATE pg_catalog.pg_type AS t
```
pgnickb, line 2: Any reason to do this over a plain
```SQL
alter type tag_map set (storage = extended);
```

sumerman: Alter type is not supported on pg12

pgnickb: Are you opposed to making the current code an exception for v12, rather than the default behavior?

sumerman: I don't see the benefit of added branching, but it surely increases complexity.

pgnickb: I find it hard to agree that manually updating catalog tables reduces complexity.

migration/incremental/025-tag-map-storage-type.sql
```diff
@@ -0,0 +1,23 @@
...
+
+-- tag_v
+UPDATE pg_catalog.pg_type AS t 
+SET typstorage = 'x' 
+WHERE t.oid = '_ps_trace.tag_v'::regtype::oid;
+
+UPDATE pg_catalog.pg_attribute AS a
+SET attstorage = 'x'
+WHERE a.attrelid = '_ps_trace.tag'::regclass::oid
+  AND a.atttypid = '_ps_trace.tag_v'::regtype::oid;
```
pgnickb, line 13: `tag_v` is never stored, so this is essentially a no-op.

migration/incremental/025-tag-map-storage-type.sql
sumerman: I would agree if this line was gone entirely, but it's "this complex line" or "alter type". Which is in my book strictly more complex than just "this complex line".

pgnickb: I'd agree if at one point pg12 won't be gone. And once it is gone we can go ahead and remove all such branches given we can find them.

jgpruitt (approved):

pgnickb (approved):

migration/incremental/025-tag-map-storage-type.sql
pgnickb: Actually I wasn't confused. We have two distinct types:
- `ps_trace.tag_v` which is used in `tag` table
- `_ps_trace.tag_v` which shouldn't ever be used anywhere.

JamesGuthrie: I don't think that's correct. Yes, the two types do exist. `ps_trace.tag_v` _was_ used in the `_ps_trace.tag` table (in `012-tracing.sql`), but it was replaced by `_ps_trace.tag_v` in `015-tracing-redesign.sql`.

PR #316, opened 2022-05-30
Fix occasional e2e test flakiness by increasing trace data retention

JamesGuthrie:


sumerman (approved):
LGTM. A nit: for something that we use in every single test file we can create a shared `.sql` script, that other files may include via `\i`

PR #323, opened 2022-06-01
Drop old versions of func/proc where signature changed

jgpruitt:
We changed the signatures of some functions and procedures. We changed the data types from `name` to `text`. We did so with `create or replace function`. Since the signature changed, postgres ran this as a `create` and we ended up with two versions of each function/procedure. This adds a migration that drops the old versions.


JamesGuthrie (approved):
LGTM!

PR #326, opened 2022-06-01
Remove some code paths for deprecated versions

jgpruitt:
Timescaledb is still optional, but if installed, version 2.6.1 or greater is required. This commit removes some code paths that only existed to support timescaledb < 2.6.1.


migration/idempotent/001-base.sql
```diff
@@ -2554,218 +2554,149 @@ LANGUAGE PLPGSQL;
...
-                FROM (
-                    SELECT DISTINCT dht.node_name -- only ping each node once
-                    FROM dht
-                ) x
-            )
-            SELECT
-                dht.table_name::text as hypertable_name,
-                dht.node_name::text as node_name,
-                up.node_up
+IF NOT _prom_catalog.is_timescaledb_installed() THEN
```
JamesGuthrie, line 2560: I know this is being nitpicky, but is there a reason that the contents of the BEGIN are not nested?

jgpruitt: Yeah, I'm 50/50 on it. When I'm looking at an individual function, the extra nest was distracting. But I don't feel strongly either way. WDYT? @JamesGuthrie

JamesGuthrie: I would prefer being able to easily see where the `BEGIN/END` block begins and ends, so I would be for indenting the contents one level.

jgpruitt: okay, i'll indent, but I will also wrap each function individually so you don't have to scroll a ton to figure out why it's indented

sumerman (approved):

PR #334, opened 2022-06-02
Metric view optimization

jgpruitt:
The prom_info.metric view can be slow when there are many chunks. The original query uses a number of functions/views. By "inlining" these and modifying them to be less generic (e.g. we KNOW that metrics only have a single dimension), we can speed things up. Furthermore, calling pg_total_relation_size is really slow relative to the rest of the query. We don't need to call pg_total_relation_size for compressed chunks. The size of compressed chunks should be static and is already recorded in a table.

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -158,7 +158,7 @@ CREATE OR REPLACE FUNCTION _prom_catalog.is_multinode()
     RETURNS BOOLEAN
     SET search_path = pg_catalog, pg_temp
 AS $func$
-    SELECT count(*) > 0 FROM timescaledb_information.data_nodes
+    SELECT EXISTS (SELECT 1 FROM timescaledb_information.data_nodes)
 $func$
 LANGUAGE sql STABLE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.is_multinode() TO prom_reader;
@@ -1470,7 +1470,7 @@ CREATE OR REPLACE FUNCTION _prom_catalog.get_metric_retention_period(schema_name
 AS $$
     SELECT COALESCE(m.retention_period, _prom_catalog.get_default_retention_period())
     FROM _prom_catalog.metric m
-    WHERE id IN (SELECT id FROM _prom_catalog.get_metric_table_name_if_exists(schema_name, get_metric_retention_period.metric_name))
+    WHERE id IN (SELECT id FROM _prom_catalog.get_metric_table_name_if_exists(get_metric_retention_period.schema_name, get_metric_retention_period.metric_name))
     UNION ALL
     SELECT _prom_catalog.get_default_retention_period()
     LIMIT 1
@@ -1485,7 +1485,7 @@ CREATE OR REPLACE FUNCTION _prom_catalog.get_metric_retention_period(metric_name
     SET search_path = pg_catalog, pg_temp
 AS $$
     SELECT *
-    FROM _prom_catalog.get_metric_retention_period('prom_data', metric_name)
+    FROM _prom_catalog.get_metric_retention_period('prom_data', get_metric_retention_period.metric_name)
 $$
 LANGUAGE SQL STABLE PARALLEL SAFE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.get_metric_retention_period(TEXT) TO prom_reader;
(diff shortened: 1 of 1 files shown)


migration/idempotent/001-base.sql
```diff
@@ -2775,11 +2775,65 @@ BEGIN
...
+                SELECT lkp.key
+                FROM _prom_catalog.label_key_position lkp
+                WHERE lkp.metric_name = m.metric_name
+                ORDER BY lkp.key
+            ) label_keys,
+            coalesce(m.retention_period, _prom_catalog.get_default_retention_period()) as retention_period,
+            _timescaledb_internal.to_interval(d.interval_length) as chunk_interval,
+            x.compressed_interval,
+            x.total_interval,
+            x.before_compression_bytes,
```
JamesGuthrie, line 2795: This used to be called `before_compression_total_bytes`, and the line below used to be called `after_compression_total_bytes`.

migration/idempotent/001-base.sql
```diff
@@ -2775,11 +2775,65 @@ BEGIN
...
+            x.after_compression_bytes,
+            (x.total_chunk_bytes + pg_total_relation_size(format('%I.%I'::text, h.schema_name, h.table_name)::regclass))::bigint as total_size_bytes,
+            pg_size_pretty(x.total_chunk_bytes + pg_total_relation_size(format('%I.%I'::text, h.schema_name, h.table_name)::regclass)) as total_size,
+            (1.0 - (x.after_compression_bytes::NUMERIC / x.before_compression_bytes::NUMERIC)) * 100 as compression_ratio,
+            x.total_chunks,
+            x.compressed_chunks
+        FROM _prom_catalog.metric m
+        INNER JOIN _timescaledb_catalog.hypertable h ON (m.table_name = h.table_name)
+        INNER JOIN _timescaledb_catalog.dimension d ON (h.id = d.hypertable_id)
+        LEFT OUTER JOIN LATERAL
```
JamesGuthrie, line 2805: (Note: I am requesting changes, but not because of this, this is just discussion)

`LATERAL`  always raises alarm bells for me. In effect the lateral query will be executed once for every row on the left-hand side. I guess in this case it is bounded to #metrics, so it might not be too bad, but if we're going to be returning the results for all metrics, then it could be more efficient to do this differently?

Did you explore using a CTE on the thing in the lateral and grouping by dimension id, and then joining that directly? Something like:

```
WITH dimension_sizes AS (
                SELECT
                d.dimension_id
                count(*) as total_chunks,
                ...
                GROUP BY ds.dimension_id
),
SELECT ...
        FROM _prom_catalog.metric m
        INNER JOIN _timescaledb_catalog.hypertable h ON (m.table_name = h.table_name)
        INNER JOIN _timescaledb_catalog.dimension d ON (h.id = d.hypertable_id)
        LEFT JOIN dimension_sizes ON dimension_sizes.dimension_id = d.id;
```

JamesGuthrie (changes requested):

migration/idempotent/001-base.sql
JamesGuthrie: I looked into this a little further. On my test system with 612 metrics it looks as though my suggested query is a little faster (maybe 10%), and uses less memory.

migration/idempotent/001-base.sql
JamesGuthrie: Just to clarify: this is the same as the "no timescaledb" branch, but different to the "multinode" branch.

migration/idempotent/001-base.sql
jgpruitt: how many chunks do you have?

migration/idempotent/001-base.sql
jgpruitt: the function's signature determines the actual names and it does not include `total`. I'll change the multinode branch

jgpruitt:
I think I may have uncovered an issue in `_timescaledb_catalog.compression_chunk_size`. The `compressed_toast_size` column is not matching what the system functions report. Need to solve this so our answer matches the old one

migration/idempotent/001-base.sql
```diff
@@ -1470,7 +1470,7 @@ CREATE OR REPLACE FUNCTION _prom_catalog.get_metric_retention_period(schema_name
 AS $$
     SELECT COALESCE(m.retention_period, _prom_catalog.get_default_retention_period())
     FROM _prom_catalog.metric m
-    WHERE id IN (SELECT id FROM _prom_catalog.get_metric_table_name_if_exists(schema_name, get_metric_retention_period.metric_name))
+    WHERE id IN (SELECT id FROM _prom_catalog.get_metric_table_name_if_exists(get_metric_retention_period.schema_name, get_metric_retention_period.metric_name))
```
pgnickb, line 1473: Usually `in` can always be rewritten as `exists` and it will be at least as slow, but might be much, much faster. Given it is a function, there is no way for pg to early interrupt the call, but I think it still might be worth it to rewrite here.

PR #341, opened 2022-06-07
Bump the version of timescaledb to 2.7.0 in our Makefile

jgpruitt:
## Description

The latest version of timescaledb is 2.7.0, but our Makefile is still building with 2.6.x. Fix this.

## Merge requirements


JamesGuthrie (changes requested):
There's also an instance of this in `alpine.Dockerfile`

JamesGuthrie (approved):

PR #345, opened 2022-06-08
Fix a bug with the tag_v type operator class clashing with the typical use case

pgnickb:
## Description

Fixes https://github.com/timescale/o11y-team-data-platform/issues/56

We want to redefine the opclass for tag_v type to not use `=` operator.
This will improve type coercion for the typical use case.
Simply dropping the opclass should be safe since we don't expect anyone
to ever *index* using this opclass.

NOTE: We currently have two distinct types: 
- _ps_trace.tag_v and 
-  ps_trace.tag_v
This is not ideal.

## Merge requirements

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -980,10 +980,10 @@ $$
           RAISE '% is not a metric. unable to drop it.', metric_name_to_be_dropped;
         END IF;
         RAISE NOTICE 'deleting "%" metric with metric_id as "%" and table_name as "%"', metric_name_to_be_dropped, deletable_metric_id, hypertable_name;
-        EXECUTE FORMAT('DROP VIEW prom_series.%1$I;', hypertable_name);
-        EXECUTE FORMAT('DROP VIEW prom_metric.%1$I;', hypertable_name);
-        EXECUTE FORMAT('DROP TABLE prom_data_series.%1$I;', hypertable_name);
-        EXECUTE FORMAT('DROP TABLE prom_data.%1$I;', hypertable_name);
+        EXECUTE FORMAT('DROP VIEW IF EXISTS prom_series.%1$I;', hypertable_name);
+        EXECUTE FORMAT('DROP VIEW IF EXISTS prom_metric.%1$I;', hypertable_name);
+        EXECUTE FORMAT('DROP TABLE IF EXISTS prom_data_series.%1$I;', hypertable_name);
+        EXECUTE FORMAT('DROP TABLE IF EXISTS prom_data.%1$I;', hypertable_name);
         DELETE FROM _prom_catalog.metric WHERE id=deletable_metric_id;
         -- clean up unreferenced labels, label_keys and its position.
         DELETE FROM _prom_catalog.label_key_position WHERE metric_name=metric_name_to_be_dropped;


migration/idempotent/007-tracing-tags.sql
```diff
@@ -98,7 +98,10 @@ $do$;
 
 DO $do$
 BEGIN
-	CREATE OPERATOR ps_trace.= (
+    -- We need to drop both the operator and the operator class if they exist
+    DROP OPERATOR CLASS IF EXISTS btree_tag_v_ops USING btree;
```
JamesGuthrie, line 102: I would suggest that you move both of these DROP statements to a new incremental migration, otherwise you're going to need to continue to implement incremental migration logic in this block.

JamesGuthrie (changes requested):
Can you give your commit a nice name and some description?

pgnickb:
> Can you give your commit a nice name and some description?

It's still a draft for precisely that reason :-)

migration/idempotent/007-tracing-tags.sql
```diff
@@ -307,10 +310,11 @@ BEGIN
     AS
```
JamesGuthrie, line 310: As you're dropping the operator class, can you schema-qualify the `btree_tag_v_ops` operator class?

sumerman (commented):
- Aren't we in danger of running into the exact same problem with other operators? <> in particular (although that one, unlike the rest, seems to have the support function declared)?
- It would be great, if you don't mind, to convert the test to use pgtap

migration/idempotent/007-tracing-tags.sql
```diff
@@ -68,19 +68,6 @@ COMMENT ON FUNCTION ps_trace.tag_v_eq(_ps_trace.tag_v, pg_catalog.jsonb)
 IS 'This function is a part of custom _ps_trace.tag_v type which is a wrapper for 
 the built-in jsonb. It is the same as its jsonb_ namesake, but has a support function attached.';
 
-CREATE OR REPLACE FUNCTION ps_trace.tag_v_eq(_ps_trace.tag_v, _ps_trace.tag_v)
```
JamesGuthrie, line 71: I'm almost tempted to suggest that we keep these function definitions in here (_and_ have them in the `incremental`).

My reasoning is as follows:

In general, we want to have all function definitions in `idempotent`. That way we have one place to look to understand which functions exist and why. In this specific case, we _need_ the function to exist in the `incremental` because we're adding an operator which needs to know about the function, and operators should be added in `incremental` (otherwise we will have a hell of a time managing them).

@sumerman @pgnickb WDYT?

JamesGuthrie (changes requested):
Please squash the commits down to one commit with a nice commit message.

migration/idempotent/007-tracing-tags.sql
sumerman: I had this thought too but then realized this specific set of functions is translating regular operators 1-1 and is unlikely to change. Therefore I don't have a strong opinion.

pgnickb:
@sumerman 
> Aren't we in danger of running into the exact same problem with other operators? <> in particular (although that one, unlike the rest, seems to have the support function declared)?
None of the `(tag_v, tag_v)` operators have support functions just yet (although they should, of course). Otherwise, you're absolutely right. I keep being confused and thinking that these operators are in an invisible schema, but they aren't.

migration/idempotent/007-tracing-tags.sql
jgpruitt: We used to create a dummy function - the definition with a basically empty body - just prior to creating the operators that depended upon them. Then, in idempotent, we would replace the dummy function with the real body. That way the real code is in one place and in a place where it can be changed easily.

Maintaining the dummy functions was a pain in the ass, so we moved to the pattern of having the operators defined in idempotent as well by wrapping them in a BEGIN/END can catching the exception if they already existed. This way the function and the operator could be kept in the same area of code, and the function is defined in one place where it can be changed.

Why did we decide to move away from this pattern?

JamesGuthrie: Good point, thanks for weighing in @jgpruitt. I don't think that there was an explicit decision to move away from that pattern. My original suggestion was to just move the DROP statements to a new incremental, keeping the operator creation in the idempotent script.

pgnickb:
In an attempt to fix the working tree I've slightly messed things up. Please refer to https://github.com/timescale/promscale_extension/pull/462 from now on.

PR #349, opened 2022-06-09
Update function definitions for type-related PGX functions

JamesGuthrie:
## Description

These type-related extension functions are not emitted by PGX, so must
be added to our idempotent scripts in order to point the functions at
the most recent version of our versioned extension binary.

## Merge requirements


sumerman (approved):
Why aren't they?

JamesGuthrie:
> Why aren't they?

We tell PGX not to ([here](https://github.com/timescale/promscale_extension/blob/master/src/aggregates/gapfill_delta.rs#L35)). The reason is that we don't have a mechanism to idempotently `CREATE TYPE`. The other option would be basically to tell PGX to emit the `CREATE TYPE`, but wrap it in a `do / BEGIN / EXCEPTION WHEN` block, as we do in our hand-written SQL.

PR #365, opened 2022-06-17
Correct storage type for columns deriving from `tag_map` and `tag_v`

JamesGuthrie:
## Description

When we changed the storage type of `ps_trace.tag_map` and
`_ps_trace.tag_v` from `PLAIN` to `EXTENDED`, we didn't update all
`pg_attribute` entries. This meant that the definitions for our views
(`ps_trace.{event,link,span}`) were incorrect.

While we noticed this specifically with our views, it could also cascade
to tables or views which end users built on top of our types.

This change modifies the storage type for all `pg_attribute` entries of
type `ps_trace.tag_map` and `_ps_trace.tag_v`.

Note: should a user have purposefully set the storage type for a
specific column, this change will overwrite that. We deem this to be
very unlikely, and `EXTENDED` is almost certainly what you want.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ We use the following categories for changes:
 
 ## [Unreleased]
 
+### Changed
+
+- Correct storage type for attributes deriving from `tag_map` and `tag_v` [#365]
+
 ## [0.5.1] - 2021-06-08
 
 ### Added
--- a/migration/incremental/027-tag-map-storage-views.sql
+++ b/migration/incremental/027-tag-map-storage-views.sql
@@ -1,0 +1,14 @@
+-- When we changed the storage type of `ps_trace.tag_map` and
+-- `_ps_trace.tag_v` from `PLAIN` to `EXTENDED`, we didn't update all
+-- `pg_attribute` entries. This meant that the definitions for our views
+-- (`ps_trace.{event,link,span}`) were incorrect.
+--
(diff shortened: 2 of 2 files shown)


pgnickb (approved):

JamesGuthrie:
The test is failing because during clean install the `ps_trace.{event,link,span}` views don't exist yet (they're only added later when the `idempotent` scripts are run).

I see two options:

1. Lift the `WHERE a.attrelid  IN ('ps_trace.span'::regclass::oid, 'ps_trace.event'::regclass::oid, 'ps_trace.link'::regclass::oid)` constraint, so that all attributes of type `ps_trace.tag_map` are hit.
2. Drop the views if they exist.

I have the feeling that option 2 is simpler.

@jgpruitt @pgnickb WDYT?

jgpruitt:
I would recommend option 1. It is possible that users have built their own views on top of tracing, and option 1 would have the added benefit of catching them as well

jgpruitt:
furthermore, if users have built structures on top of our tracing views, option 2 will need to have the `cascade` option and will drop the user's structures as well

pgnickb (approved):
```SQL
select count(*) from pg_attribute where atttypid = 'ps_trace.tag_map'::regtype and attstorage != 'x';
┌───────┐
│ count │
├───────┤
│     0 │
└───────┘
```
Looks good

PR #370, opened 2022-06-21
Add _prom_ext.re2_match(string, pattern) -> bool

JamesGuthrie:
## Description

This is the same as #362, but using a thread local `Rc` instead of a `Mutex`.
## Merge requirements

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "arrayvec"
+version = "0.7.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8da52d66c7071e2e3fa2a1e5c6d088fec47b593032b254f5e980de8ea54454d6"
+
+[[package]]
 name = "askama"
 version = "0.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


pgnickb:
This looks nice. It's more stable. I also realized that I made a mistake in estimating the actual memory use. The postgres backend with full cache at maximum size allocates ~1.3GiB of memory. Not the end of the world also it is the edge case.

I ran this query:
```SQL
select _prom_ext.re2_match(i::text, i||repeat('.?', 8623)) from generate_series(1, 32) i;
/*
 * Time: 8517.210 ms (00:08.517)
 * NEW: Time: Time: 7258.472 ms (00:07.258)
 */
```
and it took ~7 seconds to finish. It of course runs much faster the second time, when the cache is already populated (under a second, mostly under 200ms).

On the previous version with the `mutex` it takes 8.7 seconds and the run with populated cache is in the same ballpark.

A more `mutex` acquisition intensive test:

```SQL
select count(*) from (select _prom_ext.re2_match(i::text, i||'.?') from generate_series(1, 128000) i) f(x)
 where x;
/*
 * OLD: Time: 31966.645 ms (00:31.967)
 * NEW: Time: 30453.967 ms (00:30.454)
 */
```
It's a small gain so I figured I'll run a more `mutex`-intensive test with cache already populated for all checks:

```SQL
select count(*) from (select _prom_ext.re2_match(i::text, i%32||'.?') from generate_series(1, 1280000) i) f(x) where x;
/*
 * OLD: Time: 4031.161 ms (00:04.031)
 * NEW: Time: 4169.404 ms (00:04.169)
 */
```

This is rather puzzling result, but I guess it's the price of the actual error handling that happens in the new version.

(For all benchmarks I ran each query on a clean cache state 5 times and picked the best time, lazy but "good enough" approach)

Either way, at this point I deem my initial request to remove the mutex a micro-optimization. It's nice though that now we have some tests to run. I feel much more confident in this code now.
Let me know which path you want to follow and I'll approve it.

JamesGuthrie:
Closing in preference of #362

Issue #371, opened 2022-06-22
Upgrade to 0.5.0 breaks because `prom_schema_migrations` table in unexpected schema


JamesGuthrie:
The `prom_schema_migrations` table was created (and used by) Promscale to keep track of the schema version installed by Promscale in the database.

In the migration to having all objects become extension-owned (Promscale 0.11.0, extension 0.5.0), we used the `prom_schema_migrations` table as a marker to know if we were on the upgrade path, or the fresh-install path.

This required coordination between the Promscale Connector, and the Promscale extension. The connector checked if the table was present, and triggered the "extension takeover" process (drop old extension, install extension at 0.0.0, upgrade extension to newest version). The extension takeover process would drop the table `public.prom_schema_migrations` to indicate that the upgrade process was complete.

Unfortunately, the original creation of the `prom_schema_migrations` table _was not_ schema qualified. This means that the schema that the table ends up in is dependent on the user's `search_path`. This can cause the upgrade to fail.

ramonguiu:
@JamesGuthrie , in practical terms what % of users would you expect to actually be impacted by this? Most users, just a handful doing some custom installation?
Since the Promscale Connector was responsible for creating that table on fresh install, were we setting the search_path at that point and so in theory no user should be impacted?

JamesGuthrie:
This was fixed in the extension in https://github.com/timescale/promscale_extension/pull/372, and in the connector in https://github.com/timescale/promscale/pull/1452.

PR #372, opened 2022-06-22
Correctly identify and drop `prom_schema_migrations`

JamesGuthrie:
## Description

The `prom_schema_migrations` table was not schema-qualified on creation,
so it could be in any schema. Find out the schema it's in and drop.

Part of: #371

## Merge requirements

### DIFF
--- a/sql/promscale--0.0.0.sql
+++ b/sql/promscale--0.0.0.sql
@@ -17,7 +17,19 @@ BEGIN
 END;
 $stop_bgw$;
 
-DROP TABLE public.prom_schema_migrations;
+DO
+$drop_prom_schema_migrations$
+DECLARE
+    schema TEXT;
+BEGIN
+    -- The prom_schema_migrations table was not schema-qualified on creation,
+    -- so it could be in any schema. Find out the schema it's in and drop.
+    SELECT schemaname INTO STRICT schema FROM pg_catalog.pg_tables WHERE tablename = 'prom_schema_migrations';
+    IF schema IS NOT NULL THEN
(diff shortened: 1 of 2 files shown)


jgpruitt (approved):
I think it's so unlikely that maybe we don't have to worry about it, but what if there is more than one `prom_schema_migrations` table (in different schemas)?

jgpruitt:
ahh, i see your PR on the connector side. disregard my question

sql/promscale--0.0.0.sql
```diff
@@ -17,7 +17,19 @@ BEGIN
...
-DROP TABLE public.prom_schema_migrations;
+DO
+$drop_prom_schema_migrations$
+DECLARE
+    schema TEXT;
+BEGIN
+    -- The prom_schema_migrations table was not schema-qualified on creation,
+    -- so it could be in any schema. Find out the schema it's in and drop.
+    SELECT schemaname INTO STRICT schema FROM pg_catalog.pg_tables WHERE tablename = 'prom_schema_migrations';
+    IF schema IS NOT NULL THEN
```
pgnickb, line 28: FYI, this line is a no-op.

JamesGuthrie: Why?

pgnickb: `STRICT` in the line above will error out if nothing is returned.

JamesGuthrie: Aha, indeed. Thank you!

pgnickb: Are we ok with the execution stopping here if we don't find the table?

PR #387, opened 2022-06-29
Helper function to create temporary tables for ingest

niksajakovljevic:
When ingesting we COPY into temporary tables to handle ON CONFLICT situations.
We need respective changes in Promscale connector.

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -3316,3 +3316,31 @@ END;
 $$
 LANGUAGE PLPGSQL;
 GRANT EXECUTE ON FUNCTION _prom_catalog.insert_metric_row(TEXT, TIMESTAMPTZ[], DOUBLE PRECISION[], BIGINT[]) TO prom_writer;
+
+-- Creates a temporary table (if it doesn't exist) used for ingestion of metrics and traces
+-- Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
+-- Temporary table is created using supplied table and schema as prototype.
+-- Returns temporary table name
+CREATE OR REPLACE FUNCTION _prom_catalog.create_ingest_temp_table(table_name TEXT, schema_name TEXT)
+    RETURNS TEXT
+    SECURITY DEFINER
+    VOLATILE
+    SET search_path = pg_catalog, pg_temp
+AS $func$             
+BEGIN
+    SET LOCAL log_statement = 'none';
+    EXECUTE format($sql$CREATE TEMPORARY TABLE IF NOT EXISTS %I (LIKE %I.%I) ON COMMIT DELETE ROWS$sql$,
+                 table_name, schema_name, table_name);
+    EXECUTE format($sql$GRANT SELECT, INSERT ON TABLE %I TO prom_writer$sql$,
+                 table_name);
+    RETURN table_name;
+END;
+$func$
+LANGUAGE plpgsql;
+REVOKE ALL ON FUNCTION _prom_catalog.create_ingest_temp_table(TEXT, TEXT) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION _prom_catalog.create_ingest_temp_table(TEXT, TEXT) TO prom_writer;
+COMMENT ON FUNCTION _prom_catalog.create_ingest_temp_table
+IS 'Creates a temporary table (if it doesn''t exist) used for ingestion of metrics or traces.
+Temporary table is created using supplied table and schema as prototype.
+Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.'
+'Returns temporary table name';
--- a/migration/idempotent/008-tracing-functions.sql
+++ b/migration/idempotent/008-tracing-functions.sql
@@ -628,6 +628,7 @@ COMMENT ON FUNCTION ps_trace.put_instrumentation_lib
 IS 'This function creates an entry in the _ps_trace.instrumentation_lib table if it doesn''t exit and returns its id.
 It is meant to be used during span creation, to obtain a valid value for mandatory instrumentation_lib_id attribute.';
 
+-- Note: This function is getting deprecated. New versions of connector will use _prom_catalog.create_ingest_temp_table.
 -- Creates a temporary table (if it doesn't exist), suitable for tracing data ingestion.
 -- Supresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
 CREATE OR REPLACE FUNCTION _ps_trace.ensure_trace_ingest_temp_table(_temp_table_name text, _proto_table_name text)
@@ -653,7 +654,8 @@ LANGUAGE plpgsql;
 REVOKE ALL ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) FROM PUBLIC;
 GRANT EXECUTE ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) TO prom_writer;
 COMMENT ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table
-IS 'Creates a temporary table (if it doesn''t exist), suitable for tracing data ingestion. 
+IS 'Note: This function is getting deprecated. New versions of connector will use _prom_catalog.create_ingest_temp_table.
+Creates a temporary table (if it doesn''t exist), suitable for tracing data ingestion. 
 Supresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.';
 
 CREATE OR REPLACE FUNCTION ps_trace.delete_all_traces()
--- a/docs/sql-api.md
+++ b/docs/sql-api.md
@@ -501,6 +501,13 @@ function integer **_prom_catalog.count_jsonb_keys**(j jsonb)
 ```
 function boolean **_prom_catalog.create_exemplar_table_if_not_exists**(metric_name text)
 ```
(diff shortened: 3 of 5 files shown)


migration/idempotent/001-base.sql
```diff
@@ -3316,3 +3316,35 @@ END;
...
+    SET search_path = pg_catalog, pg_temp
+AS $func$
+DECLARE
+    temp_table TEXT;               
+BEGIN
+    SET LOCAL log_statement = 'none';
+    -- since table_name gets sanitized by jackc/pgx we need to remove quotes
+    temp_table := FORMAT('%s_temp',REPLACE(table_name, '"', ''));
+    EXECUTE format($sql$CREATE TEMPORARY TABLE IF NOT EXISTS %I (LIKE %I.%I) ON COMMIT DELETE ROWS$sql$,
+                 temp_table, schema_name, table_name);
```
jgpruitt, line 3337: 1. if `table_name` has double-quotes around it, then they need to be stripped both for the temp_table, AND for the `(LIKE %I.%I)`
2. I'd use `trim(both from table_name, '"')`
3. The biggest issue for me is why the table_name is being passed with double-quotes in the first place. That seems wrong and should be fixed so this issue goes away

migration/incremental/028-remove-trace-create-temp-table-func.sql
```diff
@@ -0,0 +1,2 @@
+-- We added more generic function to be used by both metrics and traces so removing traces specific one
+DROP FUNCTION IF EXISTS _ps_trace.ensure_trace_ingest_temp_table(text, text) CASCADE;
```
jgpruitt, line 2: you need to remove this function from the idempotent scripts too or it will just get recreated

jgpruitt (changes requested):

migration/idempotent/001-base.sql
niksajakovljevic: 1. Not really since it's legal to write: `"myschema"."Mytable"`.
The problem only shows up if you try to concat values and one has quotes. `temp_table` becomes `"Mytable"_temp` and that doesn't work.
3. Not sure about that. I think it's a good practice to sanitize user input and that is what PGX Go driver is doing.

migration/incremental/028-remove-trace-create-temp-table-func.sql
niksajakovljevic: Actually removing this function will break older version of Promscale. For now I'll just mark it as deprecated and later we can decide what to do here (I think we need more broad discussion about extension & connector version compatibility...)

migration/idempotent/001-base.sql
jgpruitt: If your input already has double-quotes, and you use `%I` with format, you'll get double-double-quotes

image

jgpruitt: Which characters is pgx supposed to be escaping? double quotes are valid in a SQL string.

Here, I pass two strings as SQL parameters. One is `Mytable`, and one is `"Mytable"`. Notice that pgx does not wrap `Mytable` in double quotes. In order to end up with double quotes in the SQL string, they had to be passed in that way. 

So, my question is: where in our code are we wrapping the input in double quotes and why?

```
package main

import (
	"context"
	"fmt"
	"os"

	"github.com/jackc/pgx/v5"
)

func main() {
	// urlExample := "postgres://username:password@localhost:5432/database_name"
	conn, err := pgx.Connect(context.Background(), os.Getenv("DATABASE_URL"))
	if err != nil {
		fmt.Fprintf(os.Stderr, "Unable to connect to database: %v\n", err)
		os.Exit(1)
	}
	defer conn.Close(context.Background())

	var one string
	var two string
	err = conn.QueryRow(context.Background(), "select $1, $2", `Mytable`, `"Mytable"`).Scan(&one, &two)
	if err != nil {
		fmt.Fprintf(os.Stderr, "QueryRow failed: %v\n", err)
		os.Exit(1)
	}

	fmt.Printf("one: %s\n", one)
	fmt.Printf("two: %s\n", two)
}
```
image

jgpruitt (changes requested):
I don't feel that my original feedback has been addressed yet

migration/idempotent/001-base.sql
jgpruitt: Double quotes should surround a SQL identifier in a statement when mixed-casing should be preserved. This is a different concern from sanitizing user input.

I'm asking that the text input to the function should not be wrapped in double quotes.

niksajakovljevic: The code example that you've posted is not really relevant since the problem with quotes is on PG end while interacting with PGX. Here is PGX code that sanitizes argument https://github.com/jackc/pgx/blob/master/internal/sanitize/sanitize.go

niksajakovljevic: >I'm asking that the text input to the function should not be wrapped in double quotes.
We are not wrapping any of our inputs into double quotes

migration/idempotent/001-base.sql
```diff
@@ -3316,3 +3316,35 @@ END;
...
+    SECURITY DEFINER
+    VOLATILE
+    SET search_path = pg_catalog, pg_temp
+AS $func$
+DECLARE
+    temp_table TEXT;               
+BEGIN
+    SET LOCAL log_statement = 'none';
+    -- since table_name gets sanitized by jackc/pgx we need to remove quotes
+    temp_table := FORMAT('%s_temp',REPLACE(table_name, '"', ''));
```
JamesGuthrie, line 3335: I took a deeper look at how this function behaves, and I suspect that by appending `_temp` to the table name, we violate some assumptions about how we handle long metric names (longer than 62 chars).

I think that we should not append `_temp` to the temporary table name. It is uniquely addressable by using the `pg_temp` schema.

niksajakovljevic: Yes, that's what @jgpruitt already recommended.

JamesGuthrie: Hmmm... Okay I must have missed that...

migration/idempotent/008-tracing-functions.sql
```diff
@@ -628,6 +628,7 @@ COMMENT ON FUNCTION ps_trace.put_instrumentation_lib
 IS 'This function creates an entry in the _ps_trace.instrumentation_lib table if it doesn''t exit and returns its id.
 It is meant to be used during span creation, to obtain a valid value for mandatory instrumentation_lib_id attribute.';
 
+-- Note: This function is getting depracted. New versions of connector will use _prom_catalog.create_ingest_temp_table.
```
JamesGuthrie, line 631:
```suggestion
-- Note: This function is getting deprecated. New versions of connector will use _prom_catalog.create_ingest_temp_table.
```

migration/idempotent/008-tracing-functions.sql
```diff
@@ -653,7 +654,8 @@ LANGUAGE plpgsql;
 REVOKE ALL ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) FROM PUBLIC;
 GRANT EXECUTE ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table(text, text) TO prom_writer;
 COMMENT ON FUNCTION _ps_trace.ensure_trace_ingest_temp_table
-IS 'Creates a temporary table (if it doesn''t exist), suitable for tracing data ingestion. 
+IS 'Note: This function is getting depracted. New versions of connector will use _prom_catalog.create_ingest_temp_table.
```
JamesGuthrie, line 657:
```suggestion
IS 'Note: This function is getting deprecated. New versions of connector will use _prom_catalog.create_ingest_temp_table.
```

migration/idempotent/001-base.sql
```diff
@@ -3316,3 +3316,35 @@ END;
 $$
 LANGUAGE PLPGSQL;
 GRANT EXECUTE ON FUNCTION _prom_catalog.insert_metric_row(TEXT, TIMESTAMPTZ[], DOUBLE PRECISION[], BIGINT[]) TO prom_writer;
+
+-- Creates a temporary table (if it doesn't exist) used for ingestion of metrics and traces
+-- Supresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
```
JamesGuthrie, line 3321:
```suggestion
-- Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
```

migration/idempotent/001-base.sql
```diff
@@ -3316,3 +3316,35 @@ END;
...
+    RETURN temp_table;
+END;
+$func$
+LANGUAGE plpgsql;
+REVOKE ALL ON FUNCTION _prom_catalog.create_ingest_temp_table(TEXT, TEXT) FROM PUBLIC;
+GRANT EXECUTE ON FUNCTION _prom_catalog.create_ingest_temp_table(TEXT, TEXT) TO prom_writer;
+COMMENT ON FUNCTION _prom_catalog.create_ingest_temp_table
+IS 'Creates a temporary table (if it doesn''t exist) used for ingestion of metrics or traces.
+Temporary table is created using supplied table and schema as prototype.
+Supresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.'
```
JamesGuthrie, line 3349:
```suggestion
Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.'
```

jgpruitt (approved):

JamesGuthrie (approved):

PR #388, opened 2022-06-29
Remove `_bytes_total` calculations from telemetry.

Harkishen-Singh:

## Description

Fixes: https://github.com/timescale/promscale/issues/1455

This PR optimises the `_ps_catalog.promscale_sql_telemetry()` by removing the `_bytes_total` metrics that were taking 95% of the time, sometimes even blocking for several minutes.

### With this PR

```postgres
tsdb=> select _ps_catalog.promscale_sql_telemetry_without_sizes(); -- This is the one proposed in the PR.
promscale_sql_telemetry_without_sizes 
---------------------------------------------
 
(1 row)

Time: 341.549 ms

tsdb=> select _ps_catalog.promscale_sql_telemetry(); // Previous function, taking forever.
```

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ We use the following categories for changes:
 
 - Correctly identify and drop `prom_schema_migrations` [#372]
 - Use `cluster` label from `_prom_catalog.label` to report about HA setup [#398]
+- Optimize `promscale_sql_telemetry()` by removing `metric_bytes_total` and `traces_spans_bytes_total` metrics [#388]
 
 ## [0.5.2] - 2021-06-20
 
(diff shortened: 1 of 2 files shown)


JamesGuthrie (commented):
Please add a CHANGELOG entry.

What will happen on systems where this metric was being tracked until now? I assume that they will continue to report `metric_bytes_total` and `traces_spans_bytes_total`, but the number will not change. Is that what we want? Do we want to remove the telemetry entries?

ramonguiu:
> What will happen on systems where this metric was being tracked until now? I assume that they will continue to report `metric_bytes_total` and `traces_spans_bytes_total`, but the number will not change. Is that what we want? Do we want to remove the telemetry entries?

I wouldn't spend time on this since the telemetry data is for our own use. I think we live them there and we just ignore them.

ramonguiu:
> What will happen on systems where this metric was being tracked until now? I assume that they will continue to report `metric_bytes_total` and `traces_spans_bytes_total`, but the number will not change. Is that what we want? Do we want to remove the telemetry entries?

I wouldn't spend time on this since the telemetry data is for our own use. I think we live them there and we just ignore them.

JamesGuthrie (approved):

PR #390, opened 2022-07-01
Support cargo 1.62.0 pkgid format

JamesGuthrie:
## Description

We use `cargo pkgid`, plus some text transformation, to determine the
extension version.

In cargo 1.62.0 the output of `cargo pkgid` changed from:
`<...>#promscale:<ver>`
to:
`<...>#promscale@<ver>`

## Merge requirements

### DIFF
--- a/create-upgrade-symlinks.sh
+++ b/create-upgrade-symlinks.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 prev_versions=$(cat promscale.control | sed -n 's/# upgradeable_from = \(.*\)/\1/p' | sed "s/[[:space:]']//g" | tr ',' '\n')
-cur_version=$(cargo pkgid | cut -d'#' -f2 | cut -d':' -f2)
+cur_version=$(./extract-extension-version.sh | tr -d '\n')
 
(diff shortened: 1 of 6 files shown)


create-upgrade-symlinks.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 prev_versions=$(cat promscale.control | sed -n 's/# upgradeable_from = \(.*\)/\1/p' | sed "s/[[:space:]']//g" | tr ',' '\n')
-cur_version=$(cargo pkgid | cut -d'#' -f2 | cut -d':' -f2)
+# Note: we cut on both ':' and '@' here to support pre-1.62.0 and post 1.62.0 `cargo pkgid` output
+cur_version=$(cargo pkgid | cut -d'#' -f2 | cut -d':' -f2 | cut -d'@' -f2)
```
arajkumar, line 5: Can we pass it as an arg from makefile to this script instead of repeating this?

JamesGuthrie: I've avoided repetition in a slightly different way than suggested.

sumerman (approved):

PR #391, opened 2022-07-01
Bump Rust compiler toolchain to 1.62.0

JamesGuthrie:
## Description

As in title.

## Merge requirements


JamesGuthrie:
It seems as though bumping to rust 1.62.0 somehow (I'm not sure how) introduces a dependency on glibc 2.28 (when compiled with glibc >= 2.28).

sumerman:
AFAICT, it's already 1.64 since #523

PR #395, opened 2022-07-05
Allow `make gendoc` to use a local connection

JamesGuthrie:
## Description

This requires configuring a local database to connect to. Set either the
`POSTGRES_URL` env var to a postgres-formatted connection string, or the
`POSTGRES_{USER,HOST,PORT}` env vars.

To switch back to the docker variant, set `GENDOC_USE_DOCKER=true`.
## Merge requirements


sumerman (approved):
LGTM in general.

Two optional things that came to mind:
- I tripped up on `cb`. I'd suggest calling it either `callback` because then it's obvious or `f` as it is done in stdlib.
- I find the current set of four env variables quite unwieldy. I'd rather prefer a single connection string variable.

gendoc/src/main.rs
```diff
@@ -74,17 +75,58 @@ FROM pg_catalog.pg_operator o
...
     let node = run_postgres(&docker);
-    let mut connection = connect(&node);
+    let mut connection = connect_docker(&node);
+    cb(&mut connection);
+}
+
+fn main() {
+    let use_local = env::var("GENDOC_USE_LOCAL")
+        .map(|val| val.to_ascii_lowercase() == "true")
+        .unwrap_or(false);
```
sumerman, line 114: I would prefer this to default to `true`

JamesGuthrie: Done.

PR #396, opened 2022-07-05
Adds ability to configure epoch duration

Harkishen-Singh:

Fixes: https://github.com/timescale/promscale/issues/1399

## Description

This PR does 2 things:
1. Makes the epoch duration (used in series cache on Promscale side) configurable
2. Increases the default duration to 12 hours. This will lead to fewer resets of cache and less transactional load, hence should help in improving metric ingest performance

## Merge requirements

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -11,7 +11,8 @@ FROM
     ('metric_compression'       , (exists(select 1 from pg_catalog.pg_proc where proname = 'compress_chunk')::text)),
     ('trace_retention_period'   , (30 * INTERVAL '1 days')::text),
     ('ha_lease_timeout'         , '1m'),
-    ('ha_lease_refresh'         , '10s')
+    ('ha_lease_refresh'         , '10s'),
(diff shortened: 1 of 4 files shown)


migration/idempotent/001-base.sql
```diff
@@ -2073,14 +2074,20 @@ DECLARE
...
     deletion_epoch BIGINT;
+    epoch_duration INTERVAL;
 BEGIN
     next_epoch := present_epoch + 1;
     -- technically we can delete any ID <= current_epoch - 1
     -- but it's always safe to leave them around for a bit longer
     deletion_epoch := present_epoch - 4;
 
+    SELECT _prom_catalog.get_default_value('epoch_duration')::INTERVAL INTO epoch_duration;
+    IF epoch_duration IS NULL THEN
```
jgpruitt, line 2085: `_prom_catalog.get_default_value` will not return a null value. The value you put in the `_prom_catalog.initial_default` view is what will be returned if the user has not specified a default (even if the user specifies NULL as the default).

I would remove the IF statement and add change line 2084 to use `INTO STRICT`

jgpruitt (approved):

PR #399, opened 2022-07-06
ci: skip docker push when PR is from forked repos

arajkumar:
## Description

Root cause: GITHUB_TOKEN is authorized to push to ghcr only when it is run from the owner context.

Solution: Skip ghcr image push when CI is running on forked repo context.

## Merge requirements


.github/workflows/docker.yaml
```diff
@@ -63,7 +63,7 @@ jobs:
             "AWS_SECRET_ACCESS_KEY=${{ secrets.PROMSCALE_EXTENSION_SCCACHE_AWS_SECRET_ACCESS_KEY }}"
           context: .
           file: ${{matrix.base}}.Dockerfile
-          push: true
```
arajkumar, line 66: We can use the below snippet to find PR context
```
${{ github.event_name == 'pull_request' }}
```

PR #401, opened 2022-07-07
Pass prefix for temporary table used for ingest

niksajakovljevic:
It can happen that we insert both samples and exemplars within the same
transaction. That means we need a separate temp table for these two 
so passing a prefix shoud help with that.

### DIFF
--- a/docs/sql-api.md
+++ b/docs/sql-api.md
@@ -504,9 +504,12 @@ function boolean **_prom_catalog.create_exemplar_table_if_not_exists**(metric_na
 ### _prom_catalog.create_ingest_temp_table
 Creates a temporary table (if it doesn't exist) used for ingestion of metrics or traces.
 Temporary table is created using supplied table and schema as prototype.
-Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.Returns temporary table name
+Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
+Api user has to make sure that table_prefix is unique per session/connection.
+This is to prevent different truncated table names having same temp table.
+Returns temporary table name
 ```
-function text **_prom_catalog.create_ingest_temp_table**(table_name text, schema_name text)
+function text **_prom_catalog.create_ingest_temp_table**(table_name text, schema_name text, table_prefix text)
 ```
 ### _prom_catalog.create_label_key
 
--- a/e2e/tests/snapshots/golden_tests__testdata__create_ingest_temp_table.sql.snap
+++ b/e2e/tests/snapshots/golden_tests__testdata__create_ingest_temp_table.sql.snap
@@ -14,13 +14,13 @@ expression: query_result
      # Subtest: public.test_create_ingest_temp_table()
      ok 1 - table prom_data.cpu_usage exists
      ok 2 - temp table name is well-formed
-     ok 3 - temp table cpu_usage exists
+     ok 3 - temp table exe_cpu_usage exists
      ok 4 - table prom_data."aMultiCase""fun""Metric!" exists
      ok 5 - temp table name is well-formed
-     ok 6 - temp table "aMultiCase""fun""Metric!" exists
+     ok 6 - temp table "eXe_aMultiCase""fun""Metric!" exists
      ok 7 - table prom_data."aVeryLongMultiCaseMetricWhich""might""BeTruncatedIfIt'sTOOOOOO_3" exists
      ok 8 - temp table name is well-formed
-     ok 9 - temp table "aVeryLongMultiCaseMetricWhich""might""BeTruncatedIfIt'sTOOOOOO_3" exists
+     ok 9 - temp table "ThisIsLong_aVeryLongMultiCaseMetricWhich""might""BeTruncatedIfIt" exists
      1..9
  ok 1 - public.test_create_ingest_temp_table
  1..1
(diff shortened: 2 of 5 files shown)


e2e/tests/snapshots/golden_tests__testdata__create_ingest_temp_table.sql.snap
```diff
@@ -9,18 +9,20 @@ expression: query_result
  (3,"aVeryLongMultiCaseMetricWhich""might""BeTruncatedIfIt'sTOOOOOO_3",t)
 (3 rows)
 
-                                               runtests                                               
-------------------------------------------------------------------------------------------------------
+psql:/testdata/create_ingest_temp_table.sql:41: NOTICE:  identifier "ThisIsLong_aVeryLongMultiCaseMetricWhich"might"BeTruncatedIfIt'sTOOOOOO_3" will be truncated to "ThisIsLong_aVeryLongMultiCaseMetricWhich"might"BeTruncatedIfIt'"
```
JamesGuthrie, line 12: I am concerned about this truncation.

JamesGuthrie: Let me explain my concern in more detail.

Promscale handles long metric names by creating a unique truncated version of the metric name. This happens in `get_or_create_metric_table_name`. As we see in the test, `aVeryLongMultiCaseMetricWhich"might"BeTruncatedIfIt''sTOOOOOOOOOOOOOOOLONGZOMG!` gets truncated to `ThisIsLong_aVeryLongMultiCaseMetricWhich"might"BeTruncatedIfIt'sTOOOOOO_3`. This is still uniquely identifiable (thanks to `_3` suffix).

If we now prepend something to that unique identifier, and the suffix becomes truncated, it's no longer possible to uniquely identify the table. This is not just inconvenient, it could result in the same temporary table being reused for different metrics, which would result in totally unexpected data corruption.

niksajakovljevic: I see what you mean. It's a valid concern. However, in this case we create temporary tables which are bound to a session (a connection) and there we provide uniqueness guarantees by generating unique prefix (this is connector code). I hope this makes it clear?

niksajakovljevic: So to summarize during insert in one session all metrics will have unique temporary tables.

migration/idempotent/001-base.sql
```diff
@@ -3375,24 +3375,29 @@ GRANT EXECUTE ON FUNCTION _prom_catalog.insert_metric_row(TEXT, TIMESTAMPTZ[], D
...
     SET search_path = pg_catalog, pg_temp
-AS $func$             
+AS $func$
+DECLARE
+    temp_table TEXT;         
 BEGIN
     SET LOCAL log_statement = 'none';
+    -- api user has to make sure that table_prefix is unique per session
+    -- this is to prevent different truncated table names having same temp table
+    temp_table := CONCAT(table_prefix, table_name);
```
JamesGuthrie, line 3390: Let's explicitly truncate the identifier here. I've chosen 62 characters for the same reason as it has been chosen in `_prom_catalog.pg_name_unique`.
```suggestion
    temp_table := left(CONCAT(table_prefix, table_name), 62);
```

migration/idempotent/001-base.sql
```diff
@@ -3375,24 +3375,29 @@ GRANT EXECUTE ON FUNCTION _prom_catalog.insert_metric_row(TEXT, TIMESTAMPTZ[], D
...
     SECURITY DEFINER
     VOLATILE
     SET search_path = pg_catalog, pg_temp
-AS $func$             
+AS $func$
+DECLARE
+    temp_table TEXT;         
 BEGIN
     SET LOCAL log_statement = 'none';
+    -- api user has to make sure that table_prefix is unique per session
```
JamesGuthrie, line 3388: Please move this comment to the function comment. It's a critical component of the API that the caller must ensure uniqueness of the generated temp table name through the use of the prefix.

JamesGuthrie (changes requested):

docs/sql-api.md
```diff
@@ -504,9 +504,12 @@ function boolean **_prom_catalog.create_exemplar_table_if_not_exists**(metric_na
 ### _prom_catalog.create_ingest_temp_table
 Creates a temporary table (if it doesn't exist) used for ingestion of metrics or traces.
 Temporary table is created using supplied table and schema as prototype.
-Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.Returns temporary table name
+Suppresses corresponding DDL logging, otherwise PG log may get unnecessarily verbose.
+Api user has to make sure that table_prefix is unique per session/connection.
```
JamesGuthrie, line 508: As I commented [here](https://github.com/timescale/promscale/pull/1462#discussion_r916731637), the prefix must be globally unique.

niksajakovljevic: Not sure that's true. Temporary tables are bound to session/connection. It's by PG design.

niksajakovljevic: Basically temporary tables from other sessions are invisible so no need for global uniqueness.

JamesGuthrie: Ya, you're right. Sorry I missed that.

JamesGuthrie (approved):

PR #406, opened 2022-07-11
Pick branch for runing connector e2e tests

niksajakovljevic:
If changes in the extension require changes in the connector the e2e test
against the connector mater will fail thus failing the CI pipeline. 
This approach also has it flaws since we'd like to have tests against `master` to avoid breaking
compatibility between minor versions between connector and extension. 
However the main goal here is to
unblock CI pipeline and we can improve as we go.
The change here is pretty much what we had before https://github.com/timescale/promscale_extension/pull/394


sumerman:
I'm very much against this change. Let's at least have a discussion about this. cc @JamesGuthrie @jgpruitt

niksajakovljevic:
@sumerman This change is fixing currently broken CI pipeline introduced https://github.com/timescale/promscale_extension/pull/394  so we can unblock work. I agree that we can discuss more but I think we need to bring in people involved in (https://github.com/timescale/promscale): @Harkishen-Singh @antekresic @arajkumar @alejandrodnm 
Keep in mind that https://github.com/timescale/promscale_extension/pull/394 went in silently under the radar so it' good to have more people involved.

sumerman:
For the context: the main argument from the offline discussion for using promscale's master only is to try and prevent accidental compatibility breakage. I believe extra safety far outweighs minor during development, which could be worked around.

niksajakovljevic:
Imho compatibility between two should be more about versions, not latest master. I do agree that test against master is nice but it still does not provide guarantees for our users that versions will be compatible. So maybe first we should  agree on what kind of compatibility guarantees we want to provide.

sumerman (approved):
After further consideration, we settled on running matching branches on both repos.

alejandrodnm (approved):

PR #407, opened 2022-07-11
Add Docker-based development environment

JamesGuthrie:
## Description

This change adds a new docker-based long-lived developer environment.
Through the use of bind-mounting, hot-reloading, and locally-exposed
ports, you can edit sources locally and have them recompiled in the
container.

## Merge requirements

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,0 +1,31 @@
+# Development environment
+
+In order to get started developing the extension, you need a postgres install
+with timescaledb, and Rust dependencies. For more information, see the
+[compile from source](INSTALL.md#Compile From Source) install instructions.
+
+To spare you the effort of getting this set up yourself, we provide a docker
+image with all required dependencies, which allows you to just get started.
+
+Run `make devenv` to build the docker image, start it, and expose it on port
+54321 on your local machine. This docker image mounts the current directory
+into the `/code` directory in the container. By default, it runs postgres 14
+and continually recompiles and reinstalls the promscale extension on source
+modifications. This means that you can edit the sources locally, and run tests
+against the container.
+
+You can adjust the postgres version through the `DEVENV_PG_VERSION` env var,
+for example: `DEVENV_PG_VERSION=12 make devenv`
+
+The `POSTGRES_URL` environment variable is used by tests and tools in this repo
+to point to a specific postgres installation. If you want to use the image
+above, set `POSTGRES_URL=postgres://ubuntu@localhost:54321/`.
+
+The `devenv-url` and `devenv-export-url` make targets output the URL above in
+convenient formats, for example:
+
+- To connect to the devenv db with psql: `psql $(make devenv-url)`
+- To set the `POSTGRES_URL` for all subshells: `eval $(make devenv-export-url`
+
+To permanently configure `POSTGRES_URL` when you change into this directory,
+you may consider using a tool like [direnv](https://direnv.net/).
--- a/Makefile
+++ b/Makefile
@@ -224,3 +224,26 @@ setup-buildx: ## Setup a Buildx builder
 
 promscale.control: ## A hack to boostrap the build, some pgx commands require this file. It gets re-generated later.
 	cp templates/promscale.control ./promscale.control
+
+CONTAINER_NAME=promscale-extension-devcontainer
+
+.PHONY: devcontainer
+devcontainer:
+	docker build -f dev.Dockerfile -t ${CONTAINER_NAME} .
+
+DEVENV_PG_VERSION?=14
+
+.PHONY: devenv
+devenv: VOLUME_NAME=promscale-extension-build-cache
+devenv: devcontainer
+	docker volume inspect ${VOLUME_NAME} 1>/dev/null 2>&1 || docker volume create ${VOLUME_NAME}
+	docker run --rm -v ${VOLUME_NAME}:/tmp/target ubuntu bash -c "chmod a+w /tmp/target"
+	docker run -ti -e DEVENV_PG_VERSION=${DEVENV_PG_VERSION} --rm -v ${VOLUME_NAME}:/tmp/target -p54321:288${DEVENV_PG_VERSION} -v$(shell pwd):/code ${CONTAINER_NAME}
+
+.PHONY: devenv-export-url
+devenv-export-url:
+	@echo "export POSTGRES_URL=$(make devenv-url)"
+
+.PHONY: devenv-url
+devenv-url:
+	@echo "postgres://ubuntu@localhost:54321/"
--- a/README.md
+++ b/README.md
@@ -60,5 +60,10 @@ To compile the extension (see instructions below):
 
 ## Development
 
+To quickly setup a development environment, see [DEVELOPMENT.md](DEVELOPMENT.md)
 To understand more about how to write SQL migration files for this extension, consult [this](migration/README.md) guide.
 To get a better understanding of our CI pipeline see [this document](.github/workflows/README.md).
+
+## Releasing
+
+A full checklist of the steps necessary to release a new version of the extension is available in [RELEASING.md](RELEASING.md).
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -1,0 +1,61 @@
+# Note: in future we should use the timescaledb-ha image. Unfortunately it
+# doesn't have arm64 builds, so we're doing things from scratch.
+FROM ubuntu:22.04
+
+SHELL ["/bin/bash", "-eE", "-o", "pipefail", "-c"]
+
+RUN apt update && apt install -y sudo wget curl gnupg2 lsb-release
+
+# Setup a non-root user that we'll use
+RUN adduser --disabled-password --gecos "" ubuntu && \
+ usermod -aG sudo ubuntu && \
+ echo "ubuntu ALL=(ALL:ALL) NOPASSWD: ALL" > /etc/sudoers.d/ubuntu
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Install timescaledb
+RUN echo "deb http://apt.postgresql.org/pub/repos/apt/ $(lsb_release -c -s)-pgdg main" > /etc/apt/sources.list.d/pgdg.list
+RUN wget --quiet -O - https://www.postgresql.org/media/keys/ACCC4CF8.asc | apt-key add -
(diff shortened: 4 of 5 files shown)


devenv.sh
```diff
@@ -0,0 +1,32 @@
...
+cargo pgx start
+wait_for 28814
+psql -h localhost -p 28814 -d template1 -c 'CREATE EXTENSION IF NOT EXISTS timescaledb';
+createdb -h localhost -p 28814 "$(whoami)"
+# This allows entr to work correctly on docker for mac
+export ENTR_INOTIFY_WORKAROUND=true
+
+# Note: this is not a comprehensive list of source files
+SOURCE_FILES="src migration"
+find ${SOURCE_FILES} | entr -n cargo pgx install > ~/compile.log &
```
arajkumar, line 30: Should we run `make package` here?
```suggestion
find ${SOURCE_FILES} | entr -n make package > ~/compile.log &
```

sumerman: `cargo pgx install` puts the binaries into the correct pg instance's lib directory, while `make package` merely produces them inside the `target` directory. Plus `make package` performs a couple of extra actions that would slow down development.

arajkumar: Am I missing something here? `cargo pgx install` also compiles & installs our extension?

arajkumar: Looks like it does, ignore this comment!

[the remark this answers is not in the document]
JamesGuthrie: So "it depends" because it kinda depends on what the expected behaviour is. Depending on the change that you make to the extension code, it may be necessary to drop and recreate the extension (because the SQL definitions are out-of-sync with the code). I'm not sure that we're in the position now to be able to automatically do that correctly. Although I suppose it wouldn't hurt to just kill all backends.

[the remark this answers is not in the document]
JamesGuthrie: To rephrase:

If you're running tests, or `make gendoc` against the database, then it shouldn't be necessary to restart backends/postgres. If you're connecting a long-running thing like Promscale to the database, then it would probably make sense to restart the backend, but, depending on the change that you make, it might not be sufficient to restart the backend. You might additionally need to `DROP EXTENSION promscale CASCADE;` and restart Promscale. Obviously this would cause data loss.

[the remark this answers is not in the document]
JamesGuthrie: For now I would suggest that we don't restart anything, and see where that causes trouble. When we have trouble, we can solve that problem in its own context.

dev.Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+RUN chmod a+w /usr/share/postgresql/*/extension /usr/lib/postgresql/*/lib
+
+USER ubuntu
+
+# Install rust
+ENV RUST_VERSION=1.62.0
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --profile minimal --component rustfmt --default-toolchain ${RUST_VERSION}
+ENV PATH=/home/ubuntu/.cargo/bin:$PATH
+
+RUN cargo install cargo-pgx --git https://github.com/timescale/pgx --branch promscale-staging --rev ee52db6b
```
sumerman, line 34: Maybe not in this PR but we have to consolidate cargo pgx version references somehow somewhere.

sumerman (changes requested):
This is great! I'm requesting changes because I believe `build-volume` is a typo.

devenv.sh
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/env bash
+
+trap "trap - SIGTERM && kill -- -$$" SIGINT SIGTERM EXIT
+
+wait_for() {
+    echo "waiting for $1"
+
+    for _ in $(seq 10) ; do
+      if pg_isready -d postgres -h localhost -p "$1" -U postgres 1>/dev/null 2>&1; then
+        echo "connector up"
```
niksajakovljevic, line 10: `DB is up` ?

devenv.sh
```diff
@@ -0,0 +1,32 @@
...
+    done
+    echo
+    echo "FAIL waiting for $1"
+    exit 1
+}
+
+cargo pgx start
+wait_for 28814
+psql -h localhost -p 28814 -d template1 -c 'CREATE EXTENSION IF NOT EXISTS timescaledb';
+createdb -h localhost -p 28814 "$(whoami)"
```
niksajakovljevic, line 24: Any reason to not use existing `postgres` database?

dev.Dockerfile
```diff
@@ -0,0 +1,61 @@
...
+
+RUN cargo pgx init --pg14 /usr/lib/postgresql/14/bin/pg_config --pg13 /usr/lib/postgresql/13/bin/pg_config --pg12 /usr/lib/postgresql/12/bin/pg_config
```
niksajakovljevic, line 36: We install various PG versions but it seems we only use PG14. If there is not easy way to use different PG version maybe better only have PG14.

DEVELOPMENT.md
```diff
@@ -0,0 +1,19 @@
...
+[compile from source](INSTALL.md#Compile From Source) install instructions.
+
+To spare you the effort of getting this set up yourself, we provide a docker
+image with all required dependencies, which allows you to just get started.
+
+Run `make devenv` to build the docker image, start it, and expose it on port
+54321 on your local machine. This docker image mounts the current directory
+into the `/code` directory in the container. It runs postgres 14 and
+continually recompiles and reinstalls the promscale extension on source
+modifications. This means that you can edit the sources locally, and run tests
```
niksajakovljevic, line 14: It would be nice to provide commands for running test. Something like: 
`POSTGRES_URL=postgres://ubuntu@localhost:54321/ make test`? I actually can't find make target to just run tests

README.md
```diff
@@ -60,5 +60,10 @@ To compile the extension (see instructions below):
 
 ## Development
 
+To quickly setup a development environment, see [DEVELOPMENT.md][DEVELOPMENT.md]
```
alejandrodnm, line 63:
```suggestion
To quickly setup a development environment, see [DEVELOPMENT.md](DEVELOPMENT.md)
```

DEVELOPMENT.md
sumerman: ➕

Makefile
```diff
@@ -224,3 +224,13 @@ setup-buildx: ## Setup a Buildx builder
...
+
+.PHONY: devcontainer
+devcontainer:
+	docker build -f dev.Dockerfile -t devcontainer .
+
+.PHONY: devenv
+devenv: devcontainer
+	docker volume inspect build-volume 1>/dev/null 2>&1 || docker volume create build-cache
+	docker run --rm -vbuild-cache:/tmp/target ubuntu bash -c "chmod a+w /tmp/target"
+	docker run -ti --rm -vbuild-cache:/tmp/target -p54321:28814 -p8432:8000 -v$(shell pwd):/code devcontainer
```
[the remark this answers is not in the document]
alejandrodnm: I got the following messages from this container. Is that normal?

```
FATAL:  extension "timescaledb" must be preloaded
HINT:  Please preload the timescaledb library via shared_preload_libraries.

This can be done by editing the config file at: /home/ubuntu/.pgx/data-14/postgresql.conf
and adding 'timescaledb' to the list in the shared_preload_libraries config.
        # Modify postgresql.conf:
        shared_preload_libraries = 'timescaledb'

Another way to do this, if not preloading other libraries, is with the command:
        echo "shared_preload_libraries = 'timescaledb'" >> /home/ubuntu/.pgx/data-14/postgresql.conf

(Will require a database restart.)

If you REALLY know what you are doing and would like to load the library without preloading, you can disable this check with:
        SET timescaledb.allow_install_without_preload = 'on';
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
connection to server was lost
==> /home/ubuntu/.pgx/14.log <==
                shared_preload_libraries = 'timescaledb'

        Another way to do this, if not preloading other libraries, is with the command:
                echo "shared_preload_libraries = 'timescaledb'" >> /home/ubuntu/.pgx/data-14/postgresql.conf

        (Will require a database restart.)

        If you REALLY know what you are doing and would like to load the library without preloading, you can disable this check with:
                SET timescaledb.allow_install_without_preload = 'on';
2022-07-11 14:31:24.418 GMT [45] STATEMENT:  CREATE EXTENSION IF NOT EXISTS timescaledb

==> /home/ubuntu/compile.log <==
entr: broken inotify workaround enabled
Error:
   0: control file not found in current directory

Location:
   pgx-utils/src/lib.rs:557

  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ SPANTRACE ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

   0: cargo_pgx::command::install::install_extension with pg_version=14.14 release=false features=[]
      at cargo-pgx/src/command/install.rs:53
   1: cargo_pgx::command::install::execute
      at cargo-pgx/src/command/install.rs:36

```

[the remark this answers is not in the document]
JamesGuthrie: Oh damn, this was a leftover of my testing. Will remove.

DEVELOPMENT.md
JamesGuthrie: Yes, that would be in the scope of what Valeriy is working on.

README.md
JamesGuthrie: Thanks!

dev.Dockerfile
JamesGuthrie: It is easy to use them, but right now it requires some manual intervention. I'm a little on the fence as to what to do here. I'm not sure how much of a use-case it is to run different postgres versions, but I'm sure that somebody will need to test something against pg12 because it behaves differently for some reason. I think what I will do is make it so that you can set `PG_VERSION` in the container's env, and it will run that version. WDYT?

(Initially I was thinking that it would be cool to run all postgres versions and expose them all on different ports, but then when recompiling, we would effectively have to recompile 3 times, which seems like it's not worth the effort).

Makefile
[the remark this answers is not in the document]
sumerman: Hmm. Unsurprisingly, I got the same error message, but it's two pages up and I didn't notice it 🙈

DEVELOPMENT.md
JamesGuthrie: @sumerman I don't think that is possible, because make runs every command in a subshell, so the `export` is swallowed.

sumerman: Yeah, that sounds good!

sumerman (approved):
LGTM!

Issue #276, opened 2022-05-20
Allow `e2e` package's tests to run against local connection


JamesGuthrie:
Currently the `e2e` package's tests can only be run against a docker container. We would like to be able to run the tests against a local postgres instance.

As our end-to-end tests rely on timescaledb functionality, this requires that we setup a postgres/timescaledb instance locally. We don't have good instructions (or automation) for this at the moment.

JamesGuthrie:
See implementation notes for https://github.com/timescale/promscale_extension/issues/278 as they are relevant here as well.

JamesGuthrie:
Note: the `gendoc` target also relies on Docker in the same way, it would be nice to be able to run that against local too.

PR #408, opened 2022-07-11
Extracting golden-tests aka SQL tests into their own workspace.

sumerman:
## Description

A prerequisite for https://github.com/timescale/promscale_extension/issues/276

## Merge requirements


alpine.Dockerfile
```diff
@@ -65,8 +65,9 @@ USER postgres
 
 # Pre-build extension dependencies
 RUN cd ../ && cargo pgx new promscale && cd promscale
-COPY Cargo.* Makefile /build/promscale/
+COPY Cargo.* Makefile *.sh /build/promscale/
```
JamesGuthrie, line 68: Why is this necessary?

sql-tests/tests/snapshots/tests__testdata__create_ingest_temp_table.sql.snap
```diff
@@ -0,0 +1,29 @@
+---
```
JamesGuthrie, line 1: Why are these files not registered as a "move"? Do we also need to remove them from `e2e`?

sumerman: Thank you!

alpine.Dockerfile
sumerman: Not strictly necessary, but clears warning about missing `extract-extension-version.sh` during docker build

JamesGuthrie: Okay. I would have a _slight_ preference to copy exact file matches (as opposed to globs), so that we know what we're copying in. Being that this is a "build image" (which we copy the compiled artifacts from), I'm not really too concerned.

JamesGuthrie (approved):

ha.Dockerfile
```diff
@@ -59,9 +59,11 @@ RUN chown -R postgres:postgres /build
 USER postgres
 
 COPY --chown=postgres:postgres Cargo.* /build/promscale/
+COPY --chown=postgres:postgres extract-extension-version.sh /build/promscale/
```
JamesGuthrie, line 62: This duplicates the line below it.

sumerman: dammit! thank you!

PR #409, opened 2022-07-11
Add steps to build on MacOS 

arajkumar:
## Description

This commit adds steps to build and install promscale extension on MacOS.

## Merge requirements

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -54,7 +54,7 @@ See the Postgres [documentation](https://www.postgresql.org/download/linux/redha
 - `alpine` - legacy and local development -- Avoid if you can. It will eat your ~laundry~ collation.
 - `quick` and package building images are not published anywhere and are used for local development and building packages 
 
-## Compile From Source
+## Compile From Source on Ubuntu Linux
 
 For bare-metal installations, the full instructions for setting up PostgreSQL, TimescaleDB, and the Promscale Extension are:
 
@@ -104,7 +104,7 @@ For bare-metal installations, the full instructions for setting up PostgreSQL, T
     sudo apt-get install -y git
     git clone https://github.com/timescale/promscale_extension
     cd promscale_extension
-    git checkout 0.5.0
+    git checkout master
     ```
 1) Compile and install
     ```bash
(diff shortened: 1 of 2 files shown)


INSTALL.md
```diff
@@ -121,7 +121,55 @@ For bare-metal installations, the full instructions for setting up PostgreSQL, T
...
+1) Download this repo and change directory into it
+    ```bash
+    git clone https://github.com/timescale/promscale_extension
+    cd promscale_extension
+    git checkout master
+    ```
+1) Compile and install
+    ```bash
+    make package
+    sudo cp -r ./target/release/promscale-pg14/* /
```
JamesGuthrie, line 158: Does `sudo make install` not work here?

INSTALL.md
```diff
@@ -121,7 +121,55 @@ For bare-metal installations, the full instructions for setting up PostgreSQL, T
...
     curl -L -o promscale "https://github.com/timescale/promscale/releases/download/${LATEST_VERSION}/promscale_${LATEST_VERSION}_Linux_x86_64"
     chmod +x promscale
-    ./promscale --db-name promscale --db-password promscale --db-user promscale --db-ssl-mode allow --install-extensions
+    ./promscale --db.name promscale --db.password promscale --db.user promscale --db.ssl-mode allow --startup.install-extensions --startup.upgrade-prerelease-extensions
+    ```
+## Compile From Source on MacOS(Darwin, ARM64)
+
+For bare-metal installations, the full instructions for setting up PostgreSQL, TimescaleDB, and the Promscale Extension are:
+
+1) Follow [this doc]([url](https://docs.timescale.com/install/latest/self-hosted/installation-macos/#install-self-hosted-timescaledb-using-homebrew)) and install Timescale(which also install Postgres)
```
JamesGuthrie: In general I wouldn't recommend installing postgres and timescaledb via homebrew. It does work to get stuff setup, but then homebrew updates the postgres version and suddenly you don't have timescaledb anymore.

INSTALL.md
```diff
@@ -104,7 +104,7 @@ For bare-metal installations, the full instructions for setting up PostgreSQL, T
     sudo apt-get install -y git
     git clone https://github.com/timescale/promscale_extension
     cd promscale_extension
-    git checkout 0.5.0
+    git checkout master
```
JamesGuthrie, line 107: Please don't change this. These instructions are intended as installation instructions for end users. We don't want end users installing off of `master`, that is just not a good idea.

INSTALL.md
```diff
@@ -121,7 +121,55 @@ For bare-metal installations, the full instructions for setting up PostgreSQL, T
     LATEST_VERSION=$(curl -s https://api.github.com/repos/timescale/promscale/releases/latest | grep "tag_name" | cut -d'"' -f4)
     curl -L -o promscale "https://github.com/timescale/promscale/releases/download/${LATEST_VERSION}/promscale_${LATEST_VERSION}_Linux_x86_64"
     chmod +x promscale
-    ./promscale --db-name promscale --db-password promscale --db-user promscale --db-ssl-mode allow --install-extensions
+    ./promscale --db.name promscale --db.password promscale --db.user promscale --db.ssl-mode allow --startup.install-extensions --startup.upgrade-prerelease-extensions
```
JamesGuthrie, line 124: Again, this is unnecessary for end users.

JamesGuthrie (changes requested):

INSTALL.md
arajkumar: No, `--recursive` is not a recognised option in darwin.

arajkumar: You are hitting https://github.com/timescale/promscale_extension/pull/391. 

> I tried with make install and it kind of work.

Probably you installed `cp` from GNU utils through brew?

INSTALL.md
arajkumar: Does the behaviour of brew matches with apt/yum/def on Linux distros? Do you recommend any alternative to brew?

INSTALL.md
alejandrodnm: probably, I replace a lot of the default macos tools by the gnu counterparts installed via brew

INSTALL.md
alejandrodnm: I use [asdf](https://asdf-vm.com/guide/getting-started.html) for managing node, golang, rust and postgres versions in the past.

JamesGuthrie: > Does the behaviour of brew matches with apt/yum/def on Linux distros

I don't think so. It's a combination of two things:

1. homebrew installs postgres in a path corresponding to the version, and then sets up symlinks to those paths. As an example, on my machine:
```
> ls -l $(which pg_config)
lrwxr-xr-x 1 james 39 May 23 13:42 /opt/homebrew/bin/pg_config -> ../Cellar/postgresql/14.3/bin/pg_config
```

2. Timescaledb _isn't_ installed into the postgres directory by default, as a user you need to run `timescaledb_move.sh` yourself.

The combination of these things means that when a new version of PG is released, brew updates it, and switches symlinks, but timescaledb is not installed in the new postgres install.

I guess it is as simple as running `timescaledb_move.sh` after upgrading postgres.

JamesGuthrie (approved):

arajkumar:
@alejandrodnm Do you have any concerns?

alejandrodnm (approved):

PR #417, opened 2022-07-13
A port of TestSQLDropMetricChunk from promscale

sumerman:
## Description

Porting TestSQLDropMetricChunk from promscale into this repository because it mainly tests code that lives in this repository and is known to break when we simply change defaults. (See https://github.com/timescale/promscale/pull/1484)

One notable change between the original and this port is how the epoch is tested. This test just calls epoch_abort and asserts that it RAISEs an exception, while the original expects an actual ingestion failure.

Additionally, I fixed a minor bug in epoch_abort that caused a confusing message in RAISE.

## Merge requirements

### DIFF
--- a/sql-tests/testdata/drop_metric.sql
+++ b/sql-tests/testdata/drop_metric.sql
@@ -1,0 +1,133 @@
+\unset ECHO
+\set QUIET 1
+\i '/testdata/scripts/pgtap-1.2.0.sql'
+
+SELECT * FROM plan(49);
+CREATE EXTENSION promscale;
+
+--
+-- Moved from TestSQLDropMetricChunk
+--
+
+DO $$
+DECLARE
+ s1_series_id BIGINT;
+ s2_series_id BIGINT;
+ s3_series_id BIGINT;
+BEGIN
+ -- Avoid randomness in chunk interval size by setting explicitly.
+ PERFORM _prom_catalog.get_or_create_metric_table_name('test');
+ PERFORM public.set_chunk_time_interval('prom_data.test', interval '8 hours');
+ -- Set 1h epoch duration to prevent changing defaults from affecting this test's outcome.
+ PERFORM _prom_catalog.set_default_value('epoch_duration', (interval '1 hour')::text);
+
+
+ -- this series (s1) will be deleted along with it's label
+ SELECT f.series_id 
+  FROM _prom_catalog.get_or_create_series_id_for_kv_array('test', ARRAY['__name__', 'name1'], ARRAY['test', 'value1']) f
+  INTO STRICT s1_series_id;
+ SELECT f.series_id
+  FROM _prom_catalog.get_or_create_series_id_for_kv_array('test', ARRAY['__name__', 'name1'], ARRAY['test', 'value2']) f
+  INTO STRICT s2_series_id;
+ SELECT f.series_id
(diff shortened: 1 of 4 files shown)


sumerman:
If https://github.com/timescale/promscale_extension/pull/396 is merged first this will require a rebase and a fix.

sql-tests/testdata/drop_metric.sql
```diff
@@ -0,0 +1,124 @@
...
+ SELECT f.series_id
+  FROM _prom_catalog.get_or_create_series_id_for_kv_array('test', ARRAY['__name__', 'name1'], ARRAY['test', 'value2']) f
+  INTO STRICT s2_series_id;
+ SELECT f.series_id
+  FROM _prom_catalog.get_or_create_series_id_for_kv_array('test', ARRAY['__name__', 'name1'], ARRAY['test', 'value3']) f
+  INTO STRICT s3_series_id;
+
+ INSERT INTO prom_data.test(time,value,series_id)
+  VALUES
+   (E'2009-11-10 23:59:59.999+00',0.1,s1_series_id), -- this will be dropped (notice the - 1)
```
JamesGuthrie, line 36: What is this `E'...'` syntax? I'm not familiar with it?

sumerman: It's a PostgreSQL syntactic extension to allow C-style escapes (Section 4.1.2.2 https://www.postgresql.org/docs/14/sql-syntax-lexical.html). A leftover of me copy-pasting from a GUI client that is unnecessary here. Will remove.

sql-tests/testdata/drop_metric.sql
```diff
@@ -0,0 +1,124 @@
...
+   (E'2009-11-10 23:59:59.999+00',0.1,s3_series_id), -- this will be dropped (notice the - 1)
```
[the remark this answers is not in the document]
sumerman: It's copy-pasted comment, will rephrase it. It it's about the timestamp not being -1 second from 00:00

sql-tests/testdata/drop_metric.sql
```diff
@@ -0,0 +1,127 @@
...
+BEGIN
+ RETURN NEXT is(count(*), 2::BIGINT, msg) FROM prom_data.test;
+ RETURN NEXT is(count(*), 2::BIGINT, msg || ': one series should be removed') FROM _prom_catalog.series;
+ RETURN NEXT is(count(*), 0::BIGINT, msg || ': no series should be marked for deletion') FROM _prom_catalog.series WHERE delete_epoch IS NOT NULL;
+ RETURN NEXT is(count(*), 2::BIGINT, msg || ': unused labels should deleted') FROM _prom_catalog.label where key='name1';
+ RETURN;
+END;
+$fnc$;
+
+-- Now is the time when it should actually drop the chunks
```
JamesGuthrie, line 104: Could you explicitly comment why we expect the chunks to be dropped, and series to be removed?

I _think_ that it has to do with the fact that the current_epoch > delete_epoch + 4, and because we already ran 4 deletions above, we expect the fifth one to drop.

sumerman: done

JamesGuthrie (approved):
Awesome!

PR #425, opened 2022-07-15
Allow tests to specify docker image

jgpruitt:
## Description

Some tests (upgrade tests) need to use specific docker images. Modifies the PostgresTestHarness to allow tests to specify the docker image to be used. Also changes the default docker image to the "local" image. When developing, this is most likely the image you want to be running tests against.

## Merge requirements

### DIFF
--- a/test-common/src/lib.rs
+++ b/test-common/src/lib.rs
@@ -12,9 +12,40 @@ use testcontainers::{images, Container, Docker};
 /// A docker container running Postgres
 pub type PostgresContainer<'d> = Container<'d, Cli, GenericImage>;
 
+#[allow(dead_code)]
+#[derive(Debug)]
+enum ImageOrigin {
+    Local,
+    Latest,
+    Master,
+}
+
+#[allow(dead_code)]
+#[derive(Debug)]
+enum PgVersion {
+    V14,
+    V13,
+    V12,
+}
+
+fn postgres_image_uri(origin: ImageOrigin, version: PgVersion) -> String {
+    let prefix = match origin {
+        ImageOrigin::Local => "local/dev_promscale_extension:head-ts2-",
+        ImageOrigin::Latest => "timescaledev/promscale-extension:latest-ts2.7.0-",
+        ImageOrigin::Master => "ghcr.io/timescale/dev_promscale_extension:master-ts2-",
+    };
+    let version = match version {
+        PgVersion::V12 => "pg12",
+        PgVersion::V13 => "pg13",
+        PgVersion::V14 => "pg14",
+    };
+    format!("{}{}", prefix, version)
+}
+
 #[derive(Debug, Clone)]
 pub struct PostgresTestHarness {
     pub docker: Rc<Cli>,
+    image_uri: String,
     volumes: HashMap<String, String>,
     env_vars: HashMap<String, String>,
 }
@@ -27,9 +58,9 @@ impl PostgresTestHarness {
     /// Returns the name of the docker image to use for Postgres containers.
     /// If the `TS_DOCKER_IMAGE` environment variable is set, it will return that value.
     /// Otherwise, it returns a default image.
-    pub fn postgres_image_uri() -> String {
+    pub fn default_image_uri() -> String {
         env::var("TS_DOCKER_IMAGE").unwrap_or_else(|_| {
-            String::from("ghcr.io/timescale/dev_promscale_extension:master-ts2-pg14")
+            String::from(postgres_image_uri(ImageOrigin::Local, PgVersion::V14))
         })
     }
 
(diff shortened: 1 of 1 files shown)


test-common/src/lib.rs
```diff
@@ -23,29 +24,53 @@ impl PostgresTestHarness {
     const DB: &'static str = "postgres-db-test";
     const USER: &'static str = "postgres-user-test";
     const PASSWORD: &'static str = "postgres-password-test";
+    #[allow(dead_code)]
+    const IMAGE_URI_MASTER_BRANCH_14: &'static str =
+        "ghcr.io/timescale/dev_promscale_extension:master-ts2-pg14";
+    #[allow(dead_code)]
+    const IMAGE_URI_LOCAL_14: &'static str = "local/dev_promscale_extension:head-ts2-pg14";
```
sumerman, line 31: Do you need these constants for the upgrade tests or are they all for the sake of completeness?

jgpruitt: I think so. Either an env var will specify the image to use, OR we'll run with the default. The default is set to 14. If the env var is set to a v13 or v12, I need to detect it and use a "latest" image of the same version.

sumerman: I have a suggestion then. 

Firstly, you'd need to introduce two enums: 
```rust
enum ImageOrigin {
  Local,
  Master,
}

enum PgVersion {
  v14, v13, v12
}
```

Then you can have a function that returns an image URI based on these two parameters:
```rust
fn pg_image_uri(origin: ImageOrigin, version: PgVersion) -> String
```

jgpruitt: I like that. Updated

test-common/src/lib.rs
```diff
@@ -27,25 +67,26 @@ impl PostgresTestHarness {
...
     }
 
     fn init_docker() -> Cli {
         clients::Cli::default()
     }
 
-    fn prepare_postgres_image() -> GenericImage {
-        images::generic::GenericImage::new(Self::postgres_image_uri()).with_wait_for(
+    fn prepare_image() -> GenericImage {
+        images::generic::GenericImage::new(Self::default_image_uri()).with_wait_for(
```
sumerman, line 81:
```suggestion
    fn prepare_image(&self) -> GenericImage {
        images::generic::GenericImage::new(self.img_uri).with_wait_for(
```

sumerman (changes requested):
The part I'm requesting to change is around `prepare_image`. The rest is up to you.

test-common/src/lib.rs
jgpruitt: oh yeah! whoops. good catch

test-common/src/lib.rs
```diff
@@ -12,9 +12,49 @@ use testcontainers::{images, Container, Docker};
 /// A docker container running Postgres
 pub type PostgresContainer<'d> = Container<'d, Cli, GenericImage>;
 
+enum ImageOrigin {
```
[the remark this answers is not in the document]
jgpruitt: why is the allow dead_code necessary? We use the enums in the function which should mean it's not dead, right?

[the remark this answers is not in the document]
sumerman: IDK but when I was reviewing the PR I saw a bunch of linter complaints coming from CI

test-common/src/lib.rs
```diff
@@ -27,25 +56,26 @@ impl PostgresTestHarness {
...
     }
 
     fn init_docker() -> Cli {
         clients::Cli::default()
     }
 
-    fn prepare_postgres_image() -> GenericImage {
-        images::generic::GenericImage::new(Self::postgres_image_uri()).with_wait_for(
+    fn prepare_image(&self) -> GenericImage {
+        images::generic::GenericImage::new(self.img_uri()).with_wait_for(
```
sumerman, line 70:
```suggestion
        images::generic::GenericImage::new(self.img_uri).with_wait_for(
```

test-common/src/lib.rs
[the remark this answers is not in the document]
jgpruitt: ah yes. i see them too now.

test-common/src/lib.rs
```diff
@@ -94,7 +133,7 @@ impl PostgresTestHarness {
     }
 
     pub fn run(&self) -> PostgresContainer {
-        let mut img = Self::prepare_postgres_image();
+        let mut img = Self::prepare_image();
```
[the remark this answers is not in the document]
jgpruitt: got it.

test-common/src/lib.rs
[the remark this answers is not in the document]
jgpruitt: fixed

sumerman (approved):

PR #428, opened 2022-07-19
Use actual logging in dump-restore tests

JamesGuthrie:
## Description

This change replaces usages of `println!()` with the logging-specific
macros (`info!()` and `debug!()`), allowing test execution logging to be
controlled.

To set the log level, set the RUST_LOG environment to one of `debug`,
`info`, etc.

## Merge requirements

### DIFF
--- a/e2e/tests/dump-restore-test.rs
+++ b/e2e/tests/dump-restore-test.rs
@@ -1,3 +1,6 @@
+use crate::util::debug_lines;
+use duct::cmd;
+use log::info;
 use regex::Regex;
 use similar::{ChangeTag, TextDiff};
 use std::env;
@@ -6,6 +9,34 @@ use std::path::{Path, PathBuf};
 use std::process::Command;
 use test_common::PostgresContainer;
 use test_common::PostgresContainerBlueprint;
+
+mod util;
+
+/// Tests the process of dumping and restoring a database using pg_dump
+#[test]
+fn dump_restore_test() {
+    let _ = pretty_env_logger::try_init();
+
+    let postgres_test_harness = PostgresContainerBlueprint::new()
+        .with_volume(concat!(env!("CARGO_MANIFEST_DIR"), "/scripts"), "/scripts")
+        .with_db("db")
+        .with_user("postgres");
+
+    let dir = &env::temp_dir().join("promscale-dump-restore-test");
+    make_working_dir(dir);
+    let dump = dir.join("dump.sql");
+
+    // create the first container, load it with data, snapshot it, and dump it
+    let snapshot0 = first_db(&postgres_test_harness, dir, &dump);
+
+    // create the second container, restore into it, snapshot it, add more data
+    let snapshot1 = second_db(&postgres_test_harness, dir, &dump);
+
+    // don't do `assert_eq!(snapshot0, snapshot1);`
+    // it prints both entire snapshots and is impossible to read
+    let are_snapshots_equal = are_snapshots_equal(snapshot0, snapshot1);
+    assert!(are_snapshots_equal);
+}
 
 /// Runs a SQL script file in the docker container
 ///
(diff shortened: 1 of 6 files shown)


e2e/tests/dump-restore-test.rs
```diff
@@ -32,15 +33,19 @@ fn psql_file(container: &PostgresContainer, db: &str, username: &str, path: &Pat
         .arg("--echo-all")
         .args(["-v", "ON_ERROR_STOP=1"])
         .args(["-f", path.to_str().unwrap()])
-        .spawn()
-        .unwrap()
-        .wait()
+        .output()
         .unwrap();
+    String::from_utf8(output.stdout)
```
sumerman, line 38: could be useful to output stderr as well

e2e/tests/dump-restore-test.rs
```diff
@@ -91,11 +96,20 @@ fn psql_cmd(container: &PostgresContainer, db: &str, username: &str, cmd: &str)
         .arg("--echo-all")
         .args(["-v", "ON_ERROR_STOP=1"])
         .args(["-c", cmd])
-        .spawn()
-        .unwrap()
-        .wait()
+        .output()
         .unwrap();
-    assert!(exit.success(), "psql command '{}' failed: {}", cmd, exit);
+    String::from_utf8(output.stdout)
```
sumerman, line 101: same here

e2e/tests/dump-restore-test.rs
```diff
@@ -197,16 +211,20 @@ fn snapshot_db(container: &PostgresContainer, db: &str, username: &str, path: &P
         //.args(["-v", "ON_ERROR_STOP=1"]) don't uncomment me!
         .args(["-o", snapshot.to_str().unwrap()])
         .args(["-f", "/scripts/snapshot.sql"])
-        .spawn()
-        .unwrap()
-        .wait()
+        .output()
         .unwrap();
+    String::from_utf8(output.stdout)
```
sumerman, line 216: and here 😅

e2e/tests/dump-restore-test.rs
```diff
@@ -232,16 +250,20 @@ fn dump_db(container: &PostgresContainer, db: &str, username: &str, path: &Path)
         .arg("-v")
         .args(["-F", "p"])
         .args(["-f", dump.to_str().unwrap()])
-        .spawn()
-        .unwrap()
-        .wait()
+        .output()
         .unwrap();
+    String::from_utf8(output.stdout)
```
sumerman, line 255: and here

e2e/tests/dump-restore-test.rs
```diff
@@ -275,15 +297,19 @@ fn restore_db(container: &PostgresContainer, db: &str, username: &str, path: &Pa
         //.args(["-v", "ON_ERROR_STOP=1"]) don't uncomment me!
         .args(["-v", "VERBOSITY=verbose"])
         .args(["-f", dump.to_str().unwrap()])
-        .spawn()
-        .unwrap()
-        .wait()
+        .output()
         .unwrap();
+    String::from_utf8(output.stdout)
```
sumerman, line 302: boop boop

sumerman (approved):
Mostly LGTM. I have one nit, but it applies to multiple places.

e2e/tests/dump-restore-test.rs
JamesGuthrie: Ya, the problem that I ran into is that _ideally_ we would pipe stderr to stdout and output both like that. If we output them separately, then the output will not be interleaved. Unfortunately std doesn't provide tools to do this, but it looks like there might be some crates to handle that...

JamesGuthrie: done

sumerman (approved):

PR #429, opened 2022-07-20
Test Instance Abstraction

sumerman:
## Description

- Upgrading test containers
- Introducing an abstraction for the env-controlled test instance.

A prerequisite for https://github.com/timescale/promscale_extension/issues/276

## Merge requirements


sumerman:
> In general, I like the direction that this is going in!
> 
> My main comment is that I find it a little confusing to understand all of the different options that I have to instantiate/run tests (i.e. the API surface in `test-common/src/lib.rs` is large). I guess that this is a reflection of how many ways we currently run/instantiate things, but I would prefer it if we could simplify this to fewer APIs.

Thank you. I completely agree with your sentiment. Here I tried to make it better by removing one of the shorthands completely, restructuring the code a bit more and adding a few comments here and there.

PR #434, opened 2022-07-20
Upgrade tests should use the latest image for prior version

jgpruitt:
## Description

When testing the upgrade process, we want to use the "latest" image to create the version of the extension that we will be upgrading from. We want to use the local image by default or the one specified via env var to do the upgrade to the latest version.

## Merge requirements


jgpruitt:
see this PR instead: https://github.com/timescale/promscale_extension/pull/452

PR #436, opened 2022-07-21
Update RELEASING.md with notes from the last release cycle

pgnickb:
## Merge requirements

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -33,7 +33,7 @@ mv sql/promscale-0.5.5.sql sql/promscale--0.5.5.sql
 Create symlinks:
 
 ```bash
-for i in {0..4}; do ln -s sql/promscale--0.5.5.sql sql/promscale--0.5.${i}--0.5.5.sql; done
+./create-upgrade-symlinks.sh
 ```
 
 Force add the `.sql` files to the commit, as they're in `.gitignore`:
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -14,14 +14,43 @@ The following are step-by-step instructions for releasing the Promscale extensio
 - [ ] Tag and push merged release commit.
   - CI will trigger and create a draft release with assets attached.
 - [ ] Prepare extension release notes, share with team asking for feedback.
-- [ ] Manually tag and push `timescaledev/promscale-extension` docker image (see #354)
 - [ ] Attach release notes to draft release created above.
-- [ ] Publish the GitHub release
+- [ ] Publish the GitHub release.
 
 ## Post-Release
 - [ ] Prepare for the next development cycle:
   - Generate SQL schema and commit it (and upgrade symlinks to it) to the `sql` directory.
+<details>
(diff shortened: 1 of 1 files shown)


RELEASING.md
```diff
@@ -14,14 +14,43 @@ The following are step-by-step instructions for releasing the Promscale extensio
...
 - [ ] Prepare extension release notes, share with team asking for feedback.
-- [ ] Manually tag and push `timescaledev/promscale-extension` docker image (see #354)
 - [ ] Attach release notes to draft release created above.
-- [ ] Publish the GitHub release
+- [ ] Publish the GitHub release.
 
 ## Post-Release
 - [ ] Prepare for the next development cycle:
   - Generate SQL schema and commit it (and upgrade symlinks to it) to the `sql` directory.
+<details>
```
JamesGuthrie, line 23: I would prefer it if we would keep the original formatting of this file as a bullet point checklist. If something can't be done as a bullet point item, then we should add some tooling to allow this to be the case.

RELEASING.md
```diff
@@ -14,14 +14,43 @@ The following are step-by-step instructions for releasing the Promscale extensio
...
+<summary>Examples</summary>
+
+Generate SQL file:
+
+```bash
+cargo pgx schema --release
```
JamesGuthrie, line 29:
```suggestion
cargo pgx schema --release --out=sql/promscale--$(./extract-extension-version.sh | tr -d '\n').sql
```

RELEASING.md
```diff
@@ -14,14 +14,43 @@ The following are step-by-step instructions for releasing the Promscale extensio
...
+mv sql/promscale-0.5.5.sql sql/promscale--0.5.5.sql
+```
+
+Create symlinks:
+
+```bash
+for i in {0..4}; do ln -s sql/promscale--0.5.5.sql sql/promscale--0.5.${i}--0.5.5.sql; done
```
JamesGuthrie, line 36:
```suggestion
./create-upgrade-symlinks.sh
```

RELEASING.md
```diff
@@ -14,14 +14,43 @@ The following are step-by-step instructions for releasing the Promscale extensio
...
+```
+
+Force add the `.sql` files to the commit, as they're in `.gitignore`:
+
+```bash
+git add sql/*--0.5.5*.sql --force
```
JamesGuthrie, line 42:
```suggestion
git add --force sql
```

JamesGuthrie (changes requested):

RELEASING.md
pgnickb: I'm afraid this will cause unwanted files to be added. Like the `promscale-0.x.x.sql` (the single hyphen) file.

JamesGuthrie: Ya, you're right. As discussed, you can use `extract-extension-version.sh`.

sumerman:
Subsumed by #489

PR #443, opened 2022-07-26
Add guidance on testing methods

### LINKED ISSUE #383: Consensus and documentation on testing methods
We have too many options for testing, and it's not clear which is best suited to what: 
- pgx tests
- end-to-end golden tests
- end-to-end SQL tests
- end-to-end Rust tests

We should understand what the tradeoffs between all of the above testing methods are, potentially reducing or combining the different methods (note: if we make it possible to run e2e without docker https://github.com/timescale/promscale_extension/issues/276, this may increase the amount of stuff that we can combine). Finally, we should have documentation which recommends which methods should be used for what.

JamesGuthrie:
## Description

We have a number of different ways to test code in this repository, and
it's not immediately obvious which to use. This guidance should help
developers be aware of their choices, and make the right choice.

Closes #383.

## Merge requirements

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -1,4 +1,8 @@
-# Development environment
+# Development
+
+This document covers important topics for developing in this repository.
+
+## Development environment
 
 In order to get started developing the extension, you need a postgres install
 with timescaledb, and Rust dependencies. For more information, see the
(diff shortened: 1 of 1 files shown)


DEVELOPMENT.md
```diff
@@ -28,4 +32,35 @@ convenient formats, for example:
...
+- End-to-end SQL tests
+- End-to-end Rust tests
+
+pgx tests are found in the `src` directory, typically in the same source file
+as a function defined in Rust. These tests are solely intended to test the
+behaviour of Postgres functions implemented in Rust.
+
+End-to-end SQL tests are found in the `sql-tests` workspace, in the `testdata`
+directory. The tests are written in SQL and intended to be used to test
+functions which we write in SQL (in the `migration` directory). Each SQL file
```
sumerman, line 52: I don't think it's true. For example, we have SQL tests that provide additional coverage for support functions.

```suggestion
functions which we write in SQL (in the `migration` directory) or complex interactions involving both SQL and native functions. Each SQL file
```

DEVELOPMENT.md
```diff
@@ -28,4 +32,35 @@ convenient formats, for example:
...
+
+End-to-end Rust tests are found in the `e2e` directory. They are intended to
+be used as an upgrade to end-to-end SQL tests, as the test has full control
+over test setup, can create multiple connections. For more information, see the
+[README](e2e/README.md).
+
+### Which test method to use?
+
+If you're adding a Rust function, use pgx tests. If you're adding/testing a
+SQL function, try to write it as an end-to-end SQL test. If that is not
```
sumerman, line 65:
```suggestion
SQL migration, try to write it as an end-to-end SQL test. If that is not
```

DEVELOPMENT.md
JamesGuthrie: done

DEVELOPMENT.md
JamesGuthrie: done

sumerman (approved):

PR #444, opened 2022-07-26
Bump to our fork of pgx v0.4.5

JamesGuthrie:
## Description

pgx v0.4.5 added a feature to only output function creation with
`CREATE FUNCTION`, not `CREATE OR REPLACE FUNCTION` as previously. Our
SQL migration is dependent on the function being created with
`OR REPLACE`, so we provided a patch for this which has not yet been
accepted upstream.

Until we have figured out what to do, we will use our forked version of
pgx.

## Merge requirements


sumerman (approved):
LGTM as it is, but would be great to shove `cargo install cargo-pgx --git https://github.com/timescale/pgx --branch v0.4.5-create-replace-fix --rev 6ac8a216` into `./install-pgx.sh`. Not only will it ease future upgrades in this repository, but also simplify `timescale-ha` image moving forward.

sumerman:
I'm afraid [3c09318](https://github.com/timescale/promscale_extension/pull/444/commits/3c09318e20d921ead370a00b9bad4b7bfe7c2437) is missing the shell script itself.

PR #451, opened 2022-07-28
Run upgrade tests across released Docker containers

JamesGuthrie:
## Description

This change ensures that we simulate the way that our users will upgrade
the extension, using our publically-published canonical images.

## Merge requirements

### DIFF
--- a/e2e/tests/config/mod.rs
+++ b/e2e/tests/config/mod.rs
@@ -1,0 +1,10 @@
+//! This file contains fixed docker container versions for the upgrade tests which
+//! must be bumped every time that we release a new version of the extension.
+pub(crate) const ALPINE_WITH_EXTENSION_LAST_RELEASED_PREFIX: &str =
+    "timescaledev/promscale-extension:0.5.4-ts2.7.0-pg";
+pub(crate) const HA_WITH_LAST_RELEASED_EXTENSION_PG14: &str =
+    "timescale/timescaledb-ha:pg14.4-ts2.7.1-p2";
+pub(crate) const HA_WITH_LAST_RELEASED_EXTENSION_PG13: &str =
+    "timescale/timescaledb-ha:pg13.7-ts2.7.1-p2";
+pub(crate) const HA_WITH_LAST_RELEASED_EXTENSION_PG12: &str =
+    "timescale/timescaledb-ha:pg12.11-ts2.7.1-p2";
--- a/e2e/tests/upgrade-test.rs
+++ b/e2e/tests/upgrade-test.rs
@@ -1,36 +1,443 @@
+//! End-to-end tests to ensure that different upgrade paths have the same result
+//!
+//! ## Objective
+//! These tests ensure that a fresh install of the extension produces the same
+//! results as starting from a prior version of the extension and upgrading it.
+//! It does so by comparing the database objects in the database.
+//!
+//! For full correctness, we run this procedure for both for the previously
+//! released version of the extension, as well as the first released extension
+//! version. This should catch issues like those in the following situation:
+//!
+//!   Assume we have an first version 1 and a new version 3.
+//!   In version 2 we introduce the procedure foo().
+//!   In version 3 we drop the procedure foo().
+//!
+//!   We could incorrectly implement the procedure drop in two ways:
+//!   1. We use `DROP FUNCTION IF EXISTS foo();` (foo is a procedure, not a
+//!      function).
+//!   2. We use `DROP PROCEDURE foo();` (this is missing `IF EXISTS`)
+//!
+//!   Mistake 1 would only be caught by the version 2 -> version 3 upgrade
+//!   test, because version 1 has no function foo. Mistake 2 would only be
+//!   caught by the version 1 -> version 3 upgrade test, because version 1 has
+//!   no function foo.
+//!
+//! ## Approach
+//! We have a candidate version that has been built into a docker image. This
+//! may be build locally on a developer's machine, or in CI off a branch. We
+//! want to test the install vs upgrade using this candidate image.
+//!
+//! We start with a baseline. We use the candidate image to install the
+//! promscale extension from scratch at the default version, which should be
+//! the latest version, which should also be the candidate version. We
+//! "snapshot" the structure and data of the database to a text file.
+//!
+//! Next, we want to compare this baseline snapshot to one produced by an
+//! upgrade process. We want to mimic what a real-world user would do. So, we
+//! start with a "blessed" image -- ideally, a published release image. We
+//! create the promscale extension at a "prior" version and shutdown the
+//! container.
+//!
+//! We then create a new container using the candidate image and the existing
+//! data directory. We then update the promscale extension to the latest
+//! version. We snapshot the structure and data of the database to a file.
+//!
+//! The snapshots produced by the baseline and the upgrade should be identical,
+//! aside from some minor differences, which we manually clean up.
+//!
+//! ## Complexities
+//!
+//! * We need these tests to run locally on a developer's machine, and in CI.
+//! * We should ideally test both "flavors" of our images: HA and Alpine.
+//!   Images built on developers' machines with make are of the alpine
+//!   "flavor". Both "flavors" are used in CI.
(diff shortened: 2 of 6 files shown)


JamesGuthrie:
@sumerman I'm still working on tidying this up. Could you give me some preliminary (high-level) feedback on what you would like to see change?

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,427 @@
...
+    match flavor {
+        Flavor::Alpine => match from_version {
+            // Our timescaledev images are deprecated, but they are the best option for an alpine
+            // image at the moment. They also are nicely tagged with the exact extension version.
+            FromVersion::First => format!(
+                "timescaledev/promscale-extension:0.5.0-ts2-pg{}",
+                pg_version
+            ),
+            FromVersion::Prior => format!(
+                "timescaledev/promscale-extension:0.5.4-ts2.7.0-pg{}",
```
sumerman, line 350: I don't like that this is hardcoded. I would probably look for it in the control file in the `upgradable_from`.

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,438 @@
...
+    client
+        .execute("DROP EXTENSION IF EXISTS timescaledb_toolkit;", &[])
+        .expect("failed to drop the toolkit extension");
+}
+
+fn from_image(flavor: &Flavor, pg_version: &PgVersion, from_version: &FromVersion) -> String {
+    match flavor {
+        Flavor::Alpine => match from_version {
+            // Our timescaledev images are deprecated, but they are the best option for an alpine
+            // image at the moment. They also are nicely tagged with the exact extension version.
```
sumerman, line 344: Oh gosh. This sounds like a pain. Perhaps we can load this map from a file. I would lump alpine versions (from the previous comment) into the same file then.

sumerman: Doesn't have to be a text file: it could be a Rust module, but in a separate file with nothing else but image name picking logic.

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,438 @@
...
+        .with_image_uri(to_image_uri.clone())
+        .with_volume(script_dir, "/scripts")
+        .with_env_var("PGDATA", "/var/lib/postgresql/data")
+        .with_db("db")
+        .with_user("postgres");
+    let baseline_container = baseline_blueprint.run();
+    let mut baseline_client = connect(&baseline_blueprint, &baseline_container);
+
+    // determine the versions of the promscale extension we are supposed to upgrade from and to
+    let (from_version, to_version) = {
```
sumerman, line 194: The nested lets are a little terse to read

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,438 @@
...
+        .unwrap();
+    PgVersion::try_from(pg_version)
+}
+
+/// determines the first, last, and prior versions of the promscale extension available
+fn available_extension_versions(client: &mut Client) -> (Version, Version, Version) {
+    // version 0.1 won't parse correctly
+    // version 0.5.3 and pre-releases of 0.5.3 were pulled due to a bug the broke upgrades
+    let qry = r"select version 
+    from pg_available_extension_versions
```
sumerman, line 406: Shouldn't this be logged?

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,438 @@
...
+
+    // for the database we're upgrading, we need to use two containers and the data_directory
+    // needs to be persistent in this dir. we'll nest it inside the working directory
+    let data_dir = working_dir.join("db");
+    if !data_dir.exists() {
+        create_dir_all(&data_dir).expect("failed to create working dir and data dir");
+    }
+    // The HA image requires that we map chown 777'd parent directory in, while
+    // the alpine image requires that we map the data directory in.
+    let volume_map_path = match flavor {
```
sumerman, line 166: I would move this closer to the place where it's used.

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,438 @@
...
+        .with_db("db")
+        .with_user("postgres");
+    let to_container = to_blueprint.run();
+    info!("to_container id: {}", to_container.id());
+
+    // upgrade the timescaledb extension if we need to
+    if from_timescaledb_version != to_timescaledb_version {
+        info!(
+            "upgrading from timescaledb {} to {}",
+            from_timescaledb_version, to_timescaledb_version
```
sumerman, line 295: Why does it have to happen via psql?

sumerman (commented):
I really like where this is going. Just a couple of small nits and a few questions.

e2e/tests/upgrade-test.rs
JamesGuthrie: I'm actually not sure if it's necessary. The timescaledb upgrade is quite finnicky though, the upgrade command has to be the only command executed on the connection, otherwise weird stuff happens. I will fiddle with this a bit and see.

e2e/tests/upgrade-test.rs
JamesGuthrie: I'm not sure what you mean by "logged" here?

e2e/tests/upgrade-test.rs
JamesGuthrie: done

e2e/tests/upgrade-test.rs
```diff
@@ -1,36 +1,427 @@
...
+
+#[test]
+fn upgrade_prior_with_data_test() {
+    test_upgrade(FromVersion::Prior, true);
+}
+
+fn test_upgrade(from_version: FromVersion, with_data: bool) {
+    let _ = pretty_env_logger::try_init();
+
+    // in local development, we want local/dev_promscale_extension:head-ts2-pg14
```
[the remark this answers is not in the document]
JamesGuthrie: done

e2e/tests/upgrade-test.rs
sumerman: No idea why did the comment move up a few lines. It was meant for Err => continue down below at 417

JamesGuthrie: Thanks. Done.

e2e/tests/upgrade-test.rs
JamesGuthrie: I've done an initial pass at this. Still needs some documentation.

sumerman: Much appreciated 👍🏼

sumerman (approved):
🥳

e2e/tests/config/mod.rs
```diff
@@ -0,0 +1,5 @@
+pub(crate) const ALPINE_WITH_EXTENSION_0_5_4_PREFIX: &str =
+    "timescaledev/promscale-extension:0.5.4-ts2.7.0-pg";
+pub(crate) const HA_WITH_EXTENSION_0_5_4_PG14: &str = "timescale/timescaledb-ha:pg14.4-ts2.7.1-p2";
```
JamesGuthrie, line 3: @sumerman I'm not entirely happy with having the version number in the name of the constants here. On the one hand I want it to be obvious what the referenced version is, but on the other hand it means that the constant needs to be renamed every time the image is changed, which might not actually happen. WDYT? Any other ideas?

sumerman: I'm inclined to name them `*_WITH_EXTENSION_LAST_RELEASED_*`, but I'm also fine with leaving them as they are and renaming them if we find ourselves too annoyed by it.

JamesGuthrie: done

e2e/tests/upgrade-test.rs
jgpruitt: The timescaledb docs say that you need to pass the -X flag to psql for the upgrade to work. This prevents the extension from being loaded into the session. I tried it without psql first and it failed, complaining about the extension having been loaded already

jgpruitt:
This is awesome. Very happy

PR #462, opened 2022-08-09
Another attempt at fixing the bug with tag_v type

pgnickb:
## Description

Fixes timescale/o11y-team-data-platform#56

We want to redefine the opclass for tag_v type to not use = operator.
This will improve type coercion for the typical use case.
Simply dropping the opclass should be safe since we don't expect anyone
to ever index using this opclass.

NOTE: We currently have two distinct types:

_ps_trace.tag_v and
ps_trace.tag_v
This is not ideal.
## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ We use the following categories for changes:
 - Added `prom_api.get_default_chunk_interval()` and `prom_api.get_metric_chunk_interval(TEXT)` [#464].
 
 ### Fixed
-- Incorrect type coercion when using `tag_map` with `=` operator [#345]
+- Incorrect type coercion when using `tag_map` with `=` operator [#462]
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,9 @@ We use the following categories for changes:
 
 ## [Unreleased]
 
+### Fixed
+- Incorrect type coercion when using `tag_map` with `=` operator [#345]
+
 ### Changed
 
 - `ps_trace.delete_all_traces()` can only be executed when no Promscale connectors are running [#437].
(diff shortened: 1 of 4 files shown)


sumerman (commented):
I think this is good to go, but I'd love another pair of 👁️

migration/idempotent/007-tracing-tags.sql
```diff
@@ -209,20 +210,21 @@ the built-in jsonb. It is the same as its jsonb_ namesake, but relies on tag_map
 
 DO $do$
 BEGIN
-	CREATE OPERATOR ps_trace.<> (
+	DROP OPERATOR IF EXISTS ps_trace.<> (_ps_trace.tag_v, _ps_trace.tag_v)  CASCADE;
```
JamesGuthrie, line 213: Is there any reason why these `DROP` statements don't work in an incremental migration? I'm just concerned that we're going to accidentally get something wrong by removing these at some point in the future.

CHANGELOG.md
```diff
@@ -18,6 +18,9 @@ We use the following categories for changes:
 
 - Added `prom_api.get_default_chunk_interval()` and `prom_api.get_metric_chunk_interval(TEXT)` [#464].
 
+### Fixed
+- Incorrect type coercion when using `tag_map` with `=` operator [#345]
```
JamesGuthrie, line 22:
```suggestion
- Incorrect type coercion when using `tag_map` with `=` operator [#462]
```

JamesGuthrie (changes requested):

migration/idempotent/007-tracing-tags.sql
pgnickb: That's a good question. I was thinking about this as well. Ultimately, I'd prefer related code to be in one place, so it's easier to make changes.
Unfortunately it's hard to achieve that in the current structure of the repo.

JamesGuthrie: I would argue that having the DROP statements in an idempotent script makes it _harder_ to make changes.

pgnickb: I see your point. Let me do some adjustments

JamesGuthrie (approved):

PR #464, opened 2022-08-11
Add convenience functions for getting metric chunk interval

JamesGuthrie:
## Description

`SELECT prom_api.get_default_chunk_interval();` returns the current
default chunk interval for all metrics.

`SELECT prom_api.get_metric_chunk_interval(<metric name>);` returns the
chunk interval for the configured metric, or the default chunk interval
if not explicitly set.

## Merge requirements

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,10 @@ We use the following categories for changes:
 
 ## [Unreleased]
 
+### Added
+
+- Added `prom_api.get_default_chunk_interval()` and `prom_api.get_metric_chunk_interval(TEXT)` [#464].
+
 ### Changed
 
 - `ps_trace.delete_all_traces()` can only be executed when no Promscale connectors are running [#437].
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -1426,6 +1426,17 @@ COMMENT ON FUNCTION prom_api.set_default_chunk_interval(INTERVAL)
 IS 'set the chunk interval for any metrics (existing and new) without an explicit override';
 GRANT EXECUTE ON FUNCTION prom_api.set_default_chunk_interval(INTERVAL) TO prom_admin;
 
+CREATE OR REPLACE FUNCTION prom_api.get_default_chunk_interval()
+    RETURNS INTERVAL
+    SET search_path = pg_catalog, pg_temp
+AS $func$
+    SELECT _prom_catalog.get_default_value('chunk_interval')::pg_catalog.interval;
+$func$
+    LANGUAGE SQL;
+COMMENT ON FUNCTION prom_api.get_default_chunk_interval()
+    IS 'Get the default chunk interval for all metrics';
+GRANT EXECUTE ON FUNCTION prom_api.get_default_chunk_interval() TO prom_admin;
+
 CREATE OR REPLACE FUNCTION prom_api.set_metric_chunk_interval(metric_name TEXT, chunk_interval INTERVAL)
     RETURNS BOOLEAN
     VOLATILE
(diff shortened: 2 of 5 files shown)


migration/idempotent/001-base.sql
```diff
@@ -1447,6 +1458,34 @@ COMMENT ON FUNCTION prom_api.set_metric_chunk_interval(TEXT, INTERVAL)
...
+        SELECT table_name, default_chunk_interval
+        INTO _table_name, _is_default
+        FROM _prom_catalog.metric WHERE table_schema = 'prom_data' AND metric.metric_name = get_metric_chunk_interval.metric_name;
+        IF _is_default THEN
+            RETURN prom_api.get_default_chunk_interval();
+        END IF;
+        SELECT _timescaledb_internal.to_interval(d.interval_length)
+        INTO _chunk_interval
+        FROM _timescaledb_catalog.hypertable h
+            JOIN _timescaledb_catalog.dimension d on d.hypertable_id = h.id
```
jgpruitt, line 1479: There's no way a metric hypertable can have more than the one dimension on time, correct?

migration/idempotent/001-base.sql
```diff
@@ -1426,6 +1426,17 @@ COMMENT ON FUNCTION prom_api.set_default_chunk_interval(INTERVAL)
 IS 'set the chunk interval for any metrics (existing and new) without an explicit override';
 GRANT EXECUTE ON FUNCTION prom_api.set_default_chunk_interval(INTERVAL) TO prom_admin;
 
+CREATE OR REPLACE FUNCTION prom_api.get_default_chunk_interval()
```
jgpruitt, line 1429: Maybe not for this PR, but I wonder if we should have separate default chunk intervals for metrics and traces?

jgpruitt (approved):
Have a couple of questions but not blockers. Agree with @sumerman  on the `STRICT`. Otherwise, LGTM!

CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ We use the following categories for changes:
 
 ## [Unreleased]
 
+### Added
+
+- Added `prom_api.get_default_chunk_interval()` and `prom_api.get_metric_chunk_interval(TEXT)` [#TODO].
```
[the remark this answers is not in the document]
JamesGuthrie: done

migration/idempotent/001-base.sql
JamesGuthrie: We do have separate default chunk intervals, but the chunk interval for traces is set when the trace tables are created. We don't provide any additional functions for handling the trace chunk interval. From a naming point of view, these functions should probably be renamed to `set/get_default_metric_chunk_interval`, but that's kinda out of scope of this change.

migration/idempotent/001-base.sql
```diff
@@ -1447,6 +1458,34 @@ COMMENT ON FUNCTION prom_api.set_metric_chunk_interval(TEXT, INTERVAL)
...
+    _chunk_interval INTERVAL;
+    BEGIN
+        SELECT table_name, default_chunk_interval
+        INTO _table_name, _is_default
+        FROM _prom_catalog.metric WHERE table_schema = 'prom_data' AND metric.metric_name = get_metric_chunk_interval.metric_name;
+        IF _is_default THEN
+            RETURN prom_api.get_default_chunk_interval();
+        END IF;
+        SELECT _timescaledb_internal.to_interval(d.interval_length)
+        INTO _chunk_interval
```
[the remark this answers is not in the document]
JamesGuthrie: done

migration/idempotent/001-base.sql
JamesGuthrie: Good question. I'm pretty sure not. It's not so clear from this query, but I've rewritten this to only use the (public) `timescaledb_information.dimensions` table, which I think makes it more obvious.

sumerman (approved):

PR #471, opened 2022-08-23
Prepare Release 0.5.5

jgpruitt:
## Description

Preparing release 0.5.5

## Merge requirements

### DIFF
--- a/templates/promscale.control
+++ b/templates/promscale.control
@@ -11,4 +11,4 @@ trusted = true
 {%-endif%}
 # comma-separated list of previous versions this version can be upgraded from
 # directly. This is used to generate upgrade scripts.
-# upgradeable_from = '0.0.0', '0.5.0', '0.5.1', '0.5.2', '0.5.4'
+# upgradeable_from = '0.0.0', '0.5.0', '0.5.1', '0.5.2', '0.5.3', '0.5.4'
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,11 +14,14 @@ We use the following categories for changes:
 
 ## [Unreleased]
 
+## [0.5.5] - 2022-08-23
+
 ### Added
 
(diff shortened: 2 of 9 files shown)


arajkumar:
@jgpruitt Can you wait for https://github.com/timescale/promscale_extension/pull/472?

CC @ssola

templates/promscale.control
```diff
@@ -11,4 +11,4 @@ trusted = true
 {%-endif%}
 # comma-separated list of previous versions this version can be upgraded from
 # directly. This is used to generate upgrade scripts.
-# upgradeable_from = '0.0.0', '0.5.0', '0.5.1', '0.5.2', '0.5.4'
+# upgradeable_from = '0.0.0', '0.5.0', '0.5.1', '0.5.2', '0.5.3', '0.5.4'
```
sumerman, line 14: I was under impression 0.5.3 was not included for a reason 🤔

pgnickb: Yes. If we include it, tests will break since for this upgrade test one'd need to install 0.5.3 which isn't currently possible.

jgpruitt: `0.5.3` is not an installable version.
image

However, we DO get a `0.5.3` -> `0.5.5` upgrade path:

jgpruitt: WHY is the upgrade test passing?

jgpruitt: Ah. Because this query in the test does the correct thing:

jgpruitt: @sumerman @pgnickb does this satisfy your concerns, or do you think we should still leave 0.5.3 out?

sumerman: Yeah, thank for taking the time to explain 🙏🏼

jgpruitt:
Killing this PR since we decided to tag this release `0.6.0` and there are a few PRs that just got merged and need to be included.

PR #472, opened 2022-08-24
Add telemetry event to find jaeger span ingestion

arajkumar:
This commit adds a telemetry event named `promscale_traces_jaeger_span_types` to find types of jaeger spans ingested into promscale.

Fixes https://github.com/timescale/promscale/issues/1548

## Merge requirements

### DIFF
--- a/migration/idempotent/010-telemetry.sql
+++ b/migration/idempotent/010-telemetry.sql
@@ -116,6 +116,12 @@ $$
         SELECT public.approximate_row_count('_ps_trace.span')::TEXT INTO result;
         PERFORM _ps_catalog.apply_telemetry('traces_spans_total_approx', result);
 
+        -- According to [1], Trace spans processed by Jaeger collector will have an internal attribute named 'internal.span.format' with one of the values 'jaeger|zipkin|proto|otlp|unknown'[2]. We can use this to infer whether promscale's gRPC remote storage implementation has been used or not.
+        -- [1] https://github.com/jaegertracing/jaeger/issues/1490
+        -- [2] https://github.com/jaegertracing/jaeger/blob/b7088238c017e5a54896efbf5ed38959e885e0c5/cmd/collector/app/processor/interface.go#L56-L67
+        SELECT ARRAY_AGG(value)::TEXT INTO result FROM _ps_trace.tag WHERE key='internal.span.format' AND _prom_ext.jsonb_digest(value) IN (_prom_ext.jsonb_digest('"jaeger"'), _prom_ext.jsonb_digest('"zipkin"'), _prom_ext.jsonb_digest('"proto"'), _prom_ext.jsonb_digest('"otlp"'), _prom_ext.jsonb_digest('"unknown"'));
+        PERFORM _ps_catalog.apply_telemetry('traces_jaeger_span_types', result);
+
         -- Others.
         -- The -1 is to ignore the row summing deleted rows i.e., the counter reset row. 
         SELECT (count(*) - 1)::TEXT INTO result FROM _ps_catalog.promscale_instance_information;


migration/idempotent/010-telemetry.sql
```diff
@@ -115,6 +115,9 @@ $$
 
         SELECT public.approximate_row_count('_ps_trace.span')::TEXT INTO result;
         PERFORM _ps_catalog.apply_telemetry('traces_spans_total_approx', result);
+        -- Trace spans processed by jaeger collector will always have an internal attribute named 'internal.span.format' with possible values 'jaeger|zipkin|proto' according to https://github.com/jaegertracing/jaeger/issues/1490. We can use this to infer whether promscale's gRPC remote storage implementation has been used or not.
+        SELECT count(*)::TEXT INTO result FROM _ps_trace.tag WHERE key='internal.span.format';
+        PERFORM _ps_catalog.apply_telemetry('traces_jaeger_span_ingested', result);
```
Harkishen-Singh, line 120: Sorry, I didn't understand why are we doing `count(*)` here (unless we want to count types of ingestion, i.e., jaeger/zipkin/proto). Shouldn't this field be a boolean, like `count(*) > 0`?

Harkishen-Singh (changes requested):

migration/idempotent/010-telemetry.sql
arajkumar: Yes, the intention is to count the number of span types.

arajkumar: May be we can create one metric per type for a better clarity?

```
FOR type IN (SELECT value FROM _ps_trace.tag WHERE key='internal.span.format')
LOOP
PERFORM _ps_catalog.apply_telemetry('traces_jaeger_span_type_' || type, '1');
END LOOP;
```

arajkumar: Another simple approach is aggregating as string array,

```
SELECT array_agg(value) FROM _ps_trace.tag WHERE key='internal.span.format';
```

Harkishen-Singh: I like the array_agg value. I think, we should also change the name of telemetry, to something like `traces_jaeger_span_ingested_types` which suits more to its value, which is agg form of types of Jaeger traces.

Then, it should be good to merge.

arajkumar: I named the event as `promscale_traces_jaeger_span_types`. Hope that is fine.

migration/idempotent/010-telemetry.sql
```diff
@@ -116,6 +116,12 @@ $$
         SELECT public.approximate_row_count('_ps_trace.span')::TEXT INTO result;
         PERFORM _ps_catalog.apply_telemetry('traces_spans_total_approx', result);
 
+        -- Trace spans processed by jaeger collector will always have an internal attribute named 'internal.span.format' with possible values 'jaeger|zipkin|proto' according to https://github.com/jaegertracing/jaeger/issues/1490. We can use this to infer whether promscale's gRPC remote storage implementation has been used or not.
+        SELECT ARRAY_AGG(value)::TEXT INTO result FROM _ps_trace.tag WHERE key='internal.span.format';
+        IF result IS NOT NULL THEN
+            PERFORM _ps_catalog.apply_telemetry('traces_jaeger_span_types', result);
+        END IF;
```
Harkishen-Singh, line 123: We don't need the NULL check since [apply_telemetry()](https://github.com/timescale/promscale_extension/blob/e94e75347074a69ae514d8e4a1332d0e634ee365/migration/idempotent/010-telemetry.sql#L146) already handles it.

```suggestion
PERFORM _ps_catalog.apply_telemetry('traces_jaeger_span_types', result);
```

Harkishen-Singh (approved):
One change and then we can merge.

migration/idempotent/010-telemetry.sql
arajkumar: Actually NULL causes the value to be substituted with 0, I don't want that behaviour. 

https://github.com/timescale/promscale_extension/blob/master/migration/idempotent/010-telemetry.sql#L152-L154

migration/idempotent/010-telemetry.sql
```diff
@@ -116,6 +116,12 @@ $$
         SELECT public.approximate_row_count('_ps_trace.span')::TEXT INTO result;
         PERFORM _ps_catalog.apply_telemetry('traces_spans_total_approx', result);
 
+        -- Trace spans processed by jaeger collector will always have an internal attribute named 'internal.span.format' with possible values 'jaeger|zipkin|proto' according to https://github.com/jaegertracing/jaeger/issues/1490. We can use this to infer whether promscale's gRPC remote storage implementation has been used or not.
+        SELECT ARRAY_AGG(value)::TEXT INTO result FROM _ps_trace.tag WHERE key='internal.span.format';
```
sumerman, line 120: I have two problems with the proposed query:
- Firstly, I don't like that it puts an unordered array of potentially user-supplied values into the metadata table. E.g. nothing prevents a value from being a megabyte of garbage, even though it's not a common case. It's also a way for us to accidentally leak customer's data.
- Secondly, bitmap scan is suboptimal, since we know the values you expect. It should be possible to put both a key and a `_prom_ext.jsonb_digest` of value into the `WHERE` clause and get a plan with an index-only scan.

sumerman (changes requested):

migration/idempotent/010-telemetry.sql
sumerman: I think the metric per type approach discussed above may work if you further restrict it to a set of expected types.

sumerman: A boolean would also work

arajkumar: @sumerman Nice catch.

How about `SELECT ARRAY_AGG(value)::TEXT FROM _ps_trace.tag WHERE key='internal.span.format' and value in ('"jaeger"', '"zipkin"','"proto"');` ?

It actually uses index too,

```
Aggregate  (cost=3.58..3.60 rows=1 width=32) (actual time=0.063..0.064 rows=1 loops=1)
  ->  Index Scan using tag_key_value_id_key_id_key_idx on tag  (cost=0.27..3.58 rows=1 width=25) (actual time=0.026..0.032 rows=2 loops=1)
        Index Cond: ((key)::text = 'internal.span.format'::text)
        Filter: (value = ANY ('{"\"jaeger\"","\"zipkin\"","\"proto\""}'::jsonb[]))
Planning Time: 0.482 ms
Execution Time: 0.123 ms
```

migration/idempotent/010-telemetry.sql
```diff
@@ -116,6 +116,12 @@ $$
         SELECT public.approximate_row_count('_ps_trace.span')::TEXT INTO result;
         PERFORM _ps_catalog.apply_telemetry('traces_spans_total_approx', result);
 
+        -- According to [1], Trace spans processed by Jaeger collector will have an internal attribute named 'internal.span.format' with one of the values 'jaeger|zipkin|proto|otlp|unknown'[2]. We can use this to infer whether promscale's gRPC remote storage implementation has been used or not.
+        -- [1] https://github.com/jaegertracing/jaeger/issues/1490
+        -- [2] https://github.com/jaegertracing/jaeger/blob/b7088238c017e5a54896efbf5ed38959e885e0c5/cmd/collector/app/processor/interface.go#L56-L67
+        SELECT ARRAY_AGG(value)::TEXT INTO result FROM _ps_trace.tag WHERE key='internal.span.format' AND value IN ('"jaeger"', '"zipkin"','"proto"', '"otlp"', '"unknown"');
```
sumerman, line 122: This query doesn't always generate a plan with an Index Scan. I strongly suggest using all columns in the index explicitly:

```SQL
SELECT ARRAY_AGG(value)::TEXT 
FROM _ps_trace.tag 
WHERE key='internal.span.format'
AND _prom_ext.jsonb_digest(value) IN (_prom_ext.jsonb_digest('"jaeger"'),_prom_ext.jsonb_digest('"zipkin"'),...);
```

sumerman (changes requested):

migration/idempotent/010-telemetry.sql
arajkumar: FYI, I even tried `explain analyse SELECT ARRAY_AGG(value)::TEXT FROM _ps_trace.tag WHERE key='internal.span.format' AND _prom_ext.jsonb_digest(value) in (select _prom_ext.jsonb_digest(CONCAT('"',x::text,'"')::jsonb) from unnest (array['jaeger','zipkin', 'proto', 'otlp', 'unknown']) as x);`, but again index scanning is not predictable. I will just stay with the current approach and merge it.

PR #478, opened 2022-08-25
Use Cargo.lock when installing cargo-pgx

JamesGuthrie:
## Description

Using the lock file ensures that we're not affected by breaking releases
imposed by transitive dependencies of cargo-pgx.

## Merge requirements


JamesGuthrie:
FYI I'm racing this change against https://github.com/timescale/promscale_extension/pull/479. I hope that this change will work, as it's less invasive than the other.

sumerman (approved):
LGTM

(I really wish we could have switched to a newer compiler version)

PR #479, opened 2022-08-25
Bump compiler to Rust 1.59.0

JamesGuthrie:
## Description

A transitive dependency of cargo-pgx, which we install via
`cargo install`, has bumped its MSRV to 1.59.0. Unfortunately this
affects us because we're not using a lock file for the cargo-pgx
install.

## Merge requirements


JamesGuthrie:
Closing this in favour of https://github.com/timescale/promscale_extension/pull/478. This PR doesn't build, and the other is the smaller change.

Issue #350, opened 2022-06-09
Template hardcoded .so version in incremental/idempotent SQL


JamesGuthrie:
See #124

JamesGuthrie:
Note: we may need to also template this in `promscale--0.0.0.sql`.

PR #491, opened 2022-08-30
Automate bumping extension version across multiple files

JamesGuthrie:
## Description

As in title.

## Merge requirements

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -4,7 +4,9 @@ The following are step-by-step instructions for releasing the Promscale extensio
 
 ## Create an issue to track the release
 
-Create a new issue titled "Release <version to be released here>". Copy everything below this line into that issue, and use it to to keep track of the release tasks.
+Create a new issue titled "Release `<version to be released here>`". Copy everything below this line into that issue, and use it to keep track of the release tasks.
+
+---
 
 ## Pre-release
 - [ ] Create a git branch named `pre-release-x.x.x`
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -11,14 +11,9 @@ Create a new issue titled "Release `<version to be released here>`". Copy everyt
 ## Pre-release
 - [ ] Create a git branch named `pre-release-x.x.x`
 - [ ] Ensure `upgradeable_from` in `templates/promscale.control` contains the previously released version.
-- [ ] Update `CHANGELOG.md` with release version and date.
-- [ ] Update the package version in `Cargo.toml` to the version being released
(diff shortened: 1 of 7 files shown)


RELEASING.md
```diff
@@ -4,19 +4,16 @@ The following are step-by-step instructions for releasing the Promscale extensio
...
 - [ ] Ensure `upgradeable_from` in `templates/promscale.control` contains the previously released version.
-- [ ] Update `CHANGELOG.md` with release version and date.
-- [ ] Update the package version in `Cargo.toml` to the version being released
-- [ ] Update the version in the references to the .so in `migration/incremental/001-extension.sql`
-- [ ] Update the MD5 hash of the `migration/incremental/001-extension.sql` file in `e2e/tests/incremental-freeze-test.rs`
-- [ ] Update the version in the references to the .so in `migration/idempotent/014-extension-type-functions.sql`
-- [ ] Update the version in the references to the .so in `sql/promscale--0.0.0.sql`
+- [ ] Update `CHANGELOG.md` with release version and date, ensuring that all relevant changes are reflected.
+- [ ] Update the version in all places with `./update-version.sh <version-here>`
 - [ ] Freeze any new incremental sql scripts released in this version in `incremental_freeze_test`
```
sumerman, line 16: this seems to be done by the update version script

sumerman (approved):
LGTM, but I have a question: why not have it as a template that build.rs replaces?

JamesGuthrie:
> why not have it as a template that build.rs replaces?

I'm not sure that I follow. Could you explain what you're thinking?

RELEASING.md
sumerman: right

sumerman:
> > why not have it as a template that build.rs replaces?
> 
> I'm not sure that I follow. Could you explain what you're thinking?

In most cases, we are altering the version string in the module name. For generated SQL, this is done by PGX, but since we are in charge of hand-written.sql, we should be able to substitute a set string with the expected module name.

JamesGuthrie:
> > > why not have it as a template that build.rs replaces?
> > 
> > 
> > I'm not sure that I follow. Could you explain what you're thinking?
> 
> In most cases, we are altering the version string in the module name. For generated SQL, this is done by PGX, but since we are in charge of hand-written.sql, we should be able to substitute a set string with the expected module name.

Ah, I see what you mean. I guess your question is essentially "Why didn't you implement this as https://github.com/timescale/promscale_extension/issues/350"?. I don't have a good answer. I think that I thought "we haven't done it for so long, it must be complicated".

jgpruitt (approved):
Very cool.

PR #503, opened 2022-09-08
Manually Vacuum Freeze Compressed Chunks

jgpruitt:
## Description

Adding an "engine" to the connector that will periodically look for compressed chunks that have not been vacuumed and frozen. The engine will grab a batch of them and manually vacuum (freeze, analyze) them. A configurable amount of parallelism will be employed in working through the batch.

The goal of this engine is to help prevent txid wraparound

## Merge requirements


migration/idempotent/016-vacuum-engine.sql
```diff
@@ -0,0 +1,40 @@
...
+            t.vacuum_count,
+            t.autovacuum_count,
+            t.last_vacuum,
+            t.last_autovacuum,
+            k.relallvisible
+        FROM _timescaledb_catalog.chunk c
+        INNER JOIN _timescaledb_catalog.chunk cc ON (c.dropped = false and c.compressed_chunk_id = cc.id)
+        INNER JOIN
+        (
+            SELECT
```
sumerman, line 20: `INNER JOIN _timescaledb_catalog.dimension d ON (d.id = ds.dimension_id)` seems to be sufficient in my metric queries. It's the only one we have.

sumerman (approved):
Mostly LGTM with one nit

migration/idempotent/016-vacuum-engine.sql
jgpruitt: Agreed. However, a user may create hypertables of their own and define more than one dimension on them. This view does not return ONLY metric chunks, but chunks tied to any hypertable

PR #505, opened 2022-09-08
Fix migration when a metric is not compressed

### LINKED ISSUE #504: Upgrade from before 0.0.0 broken if compression disabled
The upgrade process from the pre-0.5.0 era is broken if compression is disabled on some metrics.

cevian:
Previously, the migration logic assumed the existance of a compressed hypertable. This is not the case when compression is disabled.

Solve this case and add tests.

Fixes #504 

## Merge requirements

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -16,7 +16,7 @@ Run `make devenv` to build the docker image, start it, and expose it on port
 54321 on your local machine. This docker image mounts the current directory
 into the `/code` directory in the container. By default, it runs postgres 14
 and continually recompiles and reinstalls the promscale extension on source
-modifications. This means that you can edit the sources locally, and run SQL 
+modifications. This means that you can edit the sources locally, and run SQL
 tests against the container.
 
 You can adjust the postgres version through the `DEVENV_PG_VERSION` env var,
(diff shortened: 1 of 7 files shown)


jgpruitt (approved):

jgpruitt:
We're changing an incremental file here. Since there's a bug in it, I guess we have to.

jgpruitt:
you'll have to update the file's md5 hash in the incremental freeze test

DEVELOPMENT.md
```diff
@@ -110,4 +110,16 @@ SQL migrations there is also `make docker-quick-NN` family of targets. It's fast
...
-Further details could be found in the [corresponding document](./e2e/README.md).
\ No newline at end of file
+Further details could be found in the [corresponding document](./e2e/README.md).
+
+## Known Issues
+
+Older versions of Rust misbehaved on Apple Sillicon during panic unwind. Leading to
+errors like `(signal: 11, SIGSEGV: invalid memory reference)` during test failures.
+The solution is to upgrade to Rust >1.64 via:
+- `rustup toolchain install nightly`
```
sumerman, line 120: it's in beta channel already :)

sumerman (approved):

DEVELOPMENT.md
cevian: I tried beta and it did not work for me

sumerman: Strange. I'm on `v1.64.0-beta.1` rn

jgpruitt (approved):

PR #508, opened 2022-09-13
Fix and simplify chunks_to_vacuum_freeze

jgpruitt:
## Description

Fix and simplify chunks_to_vacuum_freeze

- relallvisible is a count, not a boolean.
- order by relfrozenxid to remove a bunch of joins

## Merge requirements


jgpruitt:
tests are not reliably passing

PR #509, opened 2022-09-14
Fix chunks_to_freeze

jgpruitt:
## Description

Simplifies and fixes to the `_ps_catalog.chunks_to_freeze` view which identifies compressed chunks that are not fully frozen.

## Merge requirements

### DIFF
--- a/migration/idempotent/016-vacuum-engine.sql
+++ b/migration/idempotent/016-vacuum-engine.sql
@@ -1,40 +1,29 @@
 DO $block$
 BEGIN
     IF _prom_catalog.is_timescaledb_installed() THEN
-        CREATE OR REPLACE VIEW _ps_catalog.chunks_to_vacuum_freeze AS
+        CREATE OR REPLACE VIEW _ps_catalog.chunks_to_freeze AS
         SELECT
-            c.hypertable_id,
-            c.id as chunk_id,
-            cc.id as compressed_chunk_id,
+            cc.id,
             cc.schema_name,
             cc.table_name,
-            t.vacuum_count,
-            t.autovacuum_count,
-            t.last_vacuum,
-            t.last_autovacuum,
-            k.relallvisible
+            greatest
+            (
+                pg_catalog.pg_stat_get_last_vacuum_time(k.oid),
+                pg_catalog.pg_stat_get_last_autovacuum_time(k.oid)
+            ) as last_vacuum,
+            k.relpages,
+            k.relallvisible,
+            k.relfrozenxid
         FROM _timescaledb_catalog.chunk c
-        INNER JOIN _timescaledb_catalog.chunk cc ON (c.dropped = false and c.compressed_chunk_id = cc.id)
-        INNER JOIN
-        (
-            SELECT
-              d.id,
-              d.hypertable_id,
-              row_number() OVER (PARTITION BY d.hypertable_id ORDER BY d.id) as dimension_nbr
-            FROM _timescaledb_catalog.dimension d
-        ) d ON (c.hypertable_id = d.hypertable_id and d.dimension_nbr = 1)
-        INNER JOIN _timescaledb_catalog.dimension_slice ds on (d.id = ds.dimension_id)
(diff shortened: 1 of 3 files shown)


migration/idempotent/016-vacuum-engine.sql
```diff
@@ -1,40 +1,27 @@
...
-        INNER JOIN _timescaledb_catalog.dimension_slice ds on (d.id = ds.dimension_id)
-        INNER JOIN _timescaledb_catalog.chunk_constraint con on (ds.id = con.dimension_slice_id and con.chunk_id = c.id)
-        INNER JOIN pg_class k ON (k.oid = format('%I.%I', cc.schema_name, cc.table_name)::regclass::oid)
-        INNER JOIN pg_stat_all_tables t ON (t.relid = k.oid)
-        WHERE k.relallvisible = 0 -- not already fully frozen
-        ORDER BY
-            greatest(t.vacuum_count, t.autovacuum_count), -- chunks vacuumed fewest times first
-            greatest(t.last_vacuum, t.last_autovacuum) NULLS FIRST, -- chunks never vacuumed first followed by ones vacuumed furthest in the past
-            ds.range_start -- chunks that represent older data first
+        INNER JOIN pg_stat_all_tables t ON (t.schemaname = cc.schema_name and t.relname = cc.table_name)
```
cevian, line 18: so this join isn't strictly needed except for some information columns. Can we test how expensive it is (ands maybe remove it)

cevian (approved):
One suggestion, but overall looks good.

migration/idempotent/016-vacuum-engine.sql
jgpruitt: Yes, I'll test it. However, I was thinking of using the last vacuum dates to limit the results in the engine. If we just vacuumed within the last x minutes, maybe don't try again in case the all visible flag just hasn't updated yet

jgpruitt: I did an `explain (analyze, verbose, buffers)` on the view both with and without `pg_stat_all_tables` with 20,000 chunks, with a variety of them compressed. Including `pg_stat_all_tables`  approx doubles the cost and sometimes doubles the time.

```
┌──────────┬─────────────────────┐
│ n_chunks │ n_compressed_chunks │
├──────────┼─────────────────────┤
│    20200 │                 200 │
└──────────┴─────────────────────┘

with pg_stat_all_tables    cost=7606.74 time=243.331   
without pg_stat_all_tables cost=3733.28 time= 95.163

┌──────────┬─────────────────────┐
│ n_chunks │ n_compressed_chunks │
├──────────┼─────────────────────┤
│    22000 │                2000 │
└──────────┴─────────────────────┘

with pg_stat_all_tables    cost=8994.66 time=290.653 
without pg_stat_all_tables cost=3728.01 time=120.464 

┌──────────┬─────────────────────┐
│ n_chunks │ n_compressed_chunks │
├──────────┼─────────────────────┤
│    32000 │               12000 │
└──────────┴─────────────────────┘

with pg_stat_all_tables    cost=15636.17 time=550.237
without pg_stat_all_tables cost= 6743.28 time=543.999

┌──────────┬─────────────────────┐
│ n_chunks │ n_compressed_chunks │
├──────────┼─────────────────────┤
│    40000 │               20000 │
└──────────┴─────────────────────┘

with pg_stat_all_tables    cost=20877.86 time=767.847
without pg_stat_all_tables cost= 9032.48 time=596.447
```

jgpruitt: but again, I think I want to do something like this to try to prevent a tight loop where we vacuum something over and over because allvisible hasn't updated by the time we check again. WDYT?

```
WHERE (coalesce(t.last_vacuum, t.last_autovacuum) is null 
or now() - greatest(t.last_vacuum, t.last_autovacuum) > interval '10 minutes')
```

jgpruitt: oh wait....

I can use `pg_stat_get_last_vacuum_time()` and `pg_stat_get_last_autovacuum_time()`. That's much better. I'll do that.

sumerman (approved):

PR #512, opened 2022-09-19
Switch from logical to time-based epoch

JamesGuthrie:
## Description

This change adjusts the existing logical cache epoch mechanism to be
time-based.

Effectively the time-based epoch mechanism uses the current system time
(`now()`) to determine the current epoch. The primary advantage of this
mechanism is that it allows for exact time-based decisions to be made on
when to drop series data (the logical epoch mechanism relied on the fact
that the epoch was incremented on a fixed time schedule).

The `_prom_catalog.global_epoch` table contains one row tracking the
`current_epoch`, and `delete_epoch`, both TIMESTAMPTZ. This table
replaces `_prom_catalog.ids_epoch`.

The `delete_epoch` column in the `prom_data_series.<metric_name>` table
was replaced with the `mark_for_deletion_epoch` column.

`current_epoch` contains the timestamp of the current epoch, and is
updated every time we delete series.
`mark_for_deletion_epoch` contains either NULL, or a TIMESTAMPTZ, which
is the value of `current_epoch` when the decision was made to delete the
series in the future.
`delete_epoch` contains the maximum value of `mark_for_deletion_epoch`
for rows in `prom_data_series.<metric_name>` which were deleted.
## Merge requirements

### DIFF
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -488,7 +488,7 @@ BEGIN
     --these indexes are logically on all series tables but they cannot be defined on the parent due to
     --dump/restore issues.
     EXECUTE format('CREATE INDEX series_labels_%s ON prom_data_series.%I USING GIN (labels)', NEW.id, NEW.table_name);
-    EXECUTE format('CREATE INDEX series_mark_for_deletion_epoch_id_%s ON prom_data_series.%I (mark_for_deletion_epoch, id) WHERE mark_for_deletion_epoch IS NOT NULL', NEW.id, NEW.table_name);
+    EXECUTE format('CREATE INDEX series_mark_for_deletion_epoch_id_%s ON prom_data_series.%I (mark_for_deletion_epoch) INCLUDE (id) WHERE mark_for_deletion_epoch IS NOT NULL', NEW.id, NEW.table_name);
 
     EXECUTE format('ALTER TABLE prom_data_series.%1$I OWNER TO prom_admin', NEW.table_name);
     EXECUTE format('GRANT ALL PRIVILEGES ON TABLE prom_data_series.%I TO prom_admin', NEW.table_name);
--- a/migration/idempotent/001-base.sql
+++ b/migration/idempotent/001-base.sql
@@ -474,7 +474,7 @@ BEGIN
             id bigint NOT NULL,
             metric_id int NOT NULL,
             labels prom_api.label_array NOT NULL,
-            delete_epoch BIGINT NULL DEFAULT NULL,
+            mark_for_deletion_epoch TIMESTAMPTZ NULL DEFAULT NULL,
             CHECK(labels[1] = %2$L AND labels[1] IS NOT NULL),
             CHECK(metric_id = %3$L),
             CONSTRAINT series_labels_id_%3$s UNIQUE(labels) INCLUDE (id),
@@ -488,7 +488,7 @@ BEGIN
     --these indexes are logically on all series tables but they cannot be defined on the parent due to
     --dump/restore issues.
     EXECUTE format('CREATE INDEX series_labels_%s ON prom_data_series.%I USING GIN (labels)', NEW.id, NEW.table_name);
-    EXECUTE format('CREATE INDEX series_delete_epoch_id_%s ON prom_data_series.%I (delete_epoch, id) WHERE delete_epoch IS NOT NULL', NEW.id, NEW.table_name);
+    EXECUTE format('CREATE INDEX series_mark_for_deletion_epoch_id_%s ON prom_data_series.%I (mark_for_deletion_epoch, id) WHERE mark_for_deletion_epoch IS NOT NULL', NEW.id, NEW.table_name);
 
     EXECUTE format('ALTER TABLE prom_data_series.%1$I OWNER TO prom_admin', NEW.table_name);
     EXECUTE format('GRANT ALL PRIVILEGES ON TABLE prom_data_series.%I TO prom_admin', NEW.table_name);
@@ -866,7 +866,7 @@ AS
 $$
 BEGIN
     EXECUTE FORMAT(
-        'UPDATE prom_data_series.%1$I SET delete_epoch = current_epoch+1 FROM _prom_catalog.ids_epoch WHERE delete_epoch IS NULL AND id = ANY($1)',
+        'UPDATE prom_data_series.%1$I SET mark_for_deletion_epoch = current_epoch FROM _prom_catalog.global_epoch WHERE mark_for_deletion_epoch IS NULL AND id = ANY($1)',
         metric_table
     ) USING series_ids;
     RETURN;
@@ -1285,7 +1285,7 @@ AS $func$
 BEGIN
     EXECUTE FORMAT($query$
         UPDATE prom_data_series.%1$I
-        SET delete_epoch = NULL
+        SET mark_for_deletion_epoch = NULL
         WHERE id = $1
     $query$, metric_table) using series_id;
 END
@@ -1322,7 +1322,7 @@ BEGIN
     ), existing AS (
         SELECT
             id,
-            CASE WHEN delete_epoch IS NOT NULL THEN
+            CASE WHEN mark_for_deletion_epoch IS NOT NULL THEN
                 _prom_catalog.resurrect_series_ids(%1$L, id)
             END
         FROM prom_data_series.%1$I as series
@@ -1367,7 +1367,7 @@ BEGIN
     ), existing AS (
         SELECT
             id,
-            CASE WHEN delete_epoch IS NOT NULL THEN
+            CASE WHEN mark_for_deletion_epoch IS NOT NULL THEN
                 _prom_catalog.resurrect_series_ids(%1$L, id)
             END
         FROM prom_data_series.%1$I as series
@@ -1397,7 +1397,7 @@ BEGIN
     WITH existing AS (
         SELECT
             id,
-            CASE WHEN delete_epoch IS NOT NULL THEN
+            CASE WHEN mark_for_deletion_epoch IS NOT NULL THEN
                 _prom_catalog.resurrect_series_ids(%1$L, id)
             END
         FROM prom_data_series.%1$I as series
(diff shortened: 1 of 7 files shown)


migration/incremental/032-logical-epoch-to-time-epoch.sql
```diff
@@ -0,0 +1,96 @@
+CREATE TABLE _prom_catalog.global_epoch (
+    current_epoch TIMESTAMPTZ NOT NULL,
+    delete_epoch TIMESTAMPTZ NOT NULL,
+    -- force there to only be a single row
+    is_unique BOOLEAN NOT NULL DEFAULT true CHECK (is_unique = true),
+    UNIQUE (is_unique)
+);
```
jgpruitt, line 7: You can force a single row without the extra column like this:

CREATE TABLE _prom_catalog.global_epoch (
    current_epoch TIMESTAMPTZ NOT NULL,
    delete_epoch TIMESTAMPTZ NOT NULL
);
INSERT INTO _prom_catalog.global_epoch (current_epoch, delete_epoch) values ('-infinity', '-infinity');
CREATE UNIQUE INDEX global_epoch_unique_idx ON _prom_catalog.global_epoch ((true));
You have to insert the one row before creating the unique index, but I see that as a bonus because your code can assume the row will always already exist.

However, I'm not sure if/how it would work with restores, so maybe what you have is best.

migration/incremental/032-logical-epoch-to-time-epoch.sql
```diff
@@ -0,0 +1,96 @@
...
+        END IF;
+        -- this ensures that pristine and migrated DBs have the same values
+        -- it's also important that current_epoch > delete_epoch
+        INSERT INTO _prom_catalog.global_epoch (current_epoch, delete_epoch, is_unique)
+            VALUES ('2020-01-01 00:00:00 UTC', '1970-01-01 00:00:00 UTC', true);
+    END;
+$block$;
+
+ALTER TABLE _prom_catalog.series ADD COLUMN
+    mark_for_deletion_epoch TIMESTAMPTZ NULL DEFAULT NULL;
```
jgpruitt, line 29: I haven't seen `DEFAULT NULL` used. In some older versions of postgres, specifying a default value on an added column will cause the table to be rewritten. I wasn't sure if this was the case on v12 or if it would happen with `DEFAULT NULL`. So, I tried it...

```
create table foo
( bar int 
, baz text
);
insert into foo (bar, baz) select x, x::text from generate_series(1, 1000) x;
create table b4 as select ctid as _ctid, bar from foo;
alter table foo add column buz timestamptz null default null;
select
  foo.ctid,
  foo.bar,
  b4._ctid,
  b4.bar
from foo
full outer join b4 on (foo.ctid = b4._ctid)
;
```
It does NOT cause the rows to be rewritten. (which is good)

migration/incremental/032-logical-epoch-to-time-epoch.sql
```diff
@@ -0,0 +1,96 @@
...
+        _is_restore_in_progress = coalesce((SELECT setting::boolean from pg_catalog.pg_settings where name = 'timescaledb.restoring'), false);
+        IF _is_restore_in_progress THEN
+            -- if a restore is in progress, we want the value from the backup, not this hardcoded init value
+            RAISE NOTICE 'restore in progress. skipping insert into _prom_catalog.global_epoch';
+            RETURN;
+        END IF;
+        -- this ensures that pristine and migrated DBs have the same values
+        -- it's also important that current_epoch > delete_epoch
+        INSERT INTO _prom_catalog.global_epoch (current_epoch, delete_epoch, is_unique)
+            VALUES ('2020-01-01 00:00:00 UTC', '1970-01-01 00:00:00 UTC', true);
```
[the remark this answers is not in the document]
jgpruitt: you could use `VALUES ('epoch', '-infinity', true)` 

feels more like proper constants to me

https://www.postgresql.org/docs/current/datatype-datetime.html#DATATYPE-DATETIME-SPECIAL-VALUES

jgpruitt (changes requested):
Consider my feedback optional. Not marking "Approve" because I'd like someone who's worked on the caching proof to also review.

migration/incremental/032-logical-epoch-to-time-epoch.sql
```diff
@@ -0,0 +1,96 @@
...
+
+-- We're changing the signature of epoch_abort, it will be recreated in the
+-- idempotent migration script.
+DROP FUNCTION IF EXISTS _prom_catalog.epoch_abort(BIGINT);
+
+DROP FUNCTION IF EXISTS _prom_catalog.delete_expired_series(text, text, text, timestamptz, BIGINT, timestamptz);
+
+UPDATE _prom_catalog.series s
+SET mark_for_deletion_epoch = now() -- NOTE: we can't use current_epoch because we've initialized it to a value in the past
+FROM (SELECT current_epoch FROM _prom_catalog.global_epoch) as global_epoch
```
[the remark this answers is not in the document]
JamesGuthrie: Nothing. I think it was a left-over from before :/ Thanks for the catch.

migration/idempotent/001-base.sql
```diff
@@ -488,7 +488,7 @@ BEGIN
     --these indexes are logically on all series tables but they cannot be defined on the parent due to
     --dump/restore issues.
     EXECUTE format('CREATE INDEX series_labels_%s ON prom_data_series.%I USING GIN (labels)', NEW.id, NEW.table_name);
-    EXECUTE format('CREATE INDEX series_delete_epoch_id_%s ON prom_data_series.%I (delete_epoch, id) WHERE delete_epoch IS NOT NULL', NEW.id, NEW.table_name);
+    EXECUTE format('CREATE INDEX series_mark_for_deletion_epoch_id_%s ON prom_data_series.%I (mark_for_deletion_epoch, id) WHERE mark_for_deletion_epoch IS NOT NULL', NEW.id, NEW.table_name);
```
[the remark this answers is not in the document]
JamesGuthrie: TIL about `tablesample`. Thank you for sharing this analysis!

I also took a look at this: https://use-the-index-luke.com/blog/2019-04/include-columns-in-btree-indexes, which is an excellent explainer on the structural difference between `INCLUDE` and not.

I've made this change.

migration/idempotent/001-base.sql
```diff
@@ -474,7 +474,7 @@ BEGIN
             id bigint NOT NULL,
             metric_id int NOT NULL,
             labels prom_api.label_array NOT NULL,
-            delete_epoch BIGINT NULL DEFAULT NULL,
+            mark_for_deletion_epoch TIMESTAMPTZ NULL DEFAULT NULL,
```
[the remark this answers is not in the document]
JamesGuthrie: It is a little awkward. It is the epoch in which the row was _marked_.

migration/incremental/032-logical-epoch-to-time-epoch.sql
```diff
@@ -0,0 +1,96 @@
...
+            WHERE table_schema = 'prom_data'
+        )
+        LOOP
+            -- Add partial index on mark_for_deletion epoch for all existing partitions of _prom_catalog.series
+            EXECUTE format($$
+                CREATE INDEX IF NOT EXISTS
+                    series_mark_for_deletion_epoch_id_%s
+                ON prom_data_series.%I (mark_for_deletion_epoch, id)
+                WHERE mark_for_deletion_epoch IS NOT NULL
+                $$, _rec.id, _rec.table_name);
```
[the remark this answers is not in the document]
JamesGuthrie: 👍

migration/incremental/032-logical-epoch-to-time-epoch.sql
JamesGuthrie: Thank you for the analysis!

migration/incremental/032-logical-epoch-to-time-epoch.sql
JamesGuthrie: I have adopted this suggestion as well. Let's see if the build likes it.

migration/idempotent/001-base.sql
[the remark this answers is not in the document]
jgpruitt: ok. i can't think of a better name so never mind

migration/idempotent/001-base.sql
```diff
@@ -2182,19 +2174,21 @@ BEGIN
...
-        SELECT ARRAY(SELECT DISTINCT unnest(labels) as label_id
-            FROM deleted_series)
-    $query$, metric_schema, metric_table, metric_series_table, deletion_epoch) INTO label_array;
-
+        SELECT
+          array_agg(DISTINCT labels.label) as label_id
+        , max(mark_for_deletion_epoch) as max_deletion_time
+        FROM deleted_series d,
+        LATERAL unnest(d.labels) as labels(label)
+    $query$, metric_schema, metric_table, metric_series_table, deletion_time) INTO label_array, max_deletion_time;
```
sumerman, line 2191: Where did `max_deletion_time` come from? I don't remember it appearing either in specs or in Mat's proof.

migration/idempotent/001-base.sql
```diff
@@ -2233,16 +2227,15 @@ BEGIN
         SET LOCAL jit = DEFAULT;
     END IF;
 
-    UPDATE _prom_catalog.ids_epoch
-        SET (current_epoch, last_update_time) = (next_epoch, now())
-        WHERE current_epoch < next_epoch;
+    UPDATE _prom_catalog.global_epoch e
```
sumerman, line 2230: I realize this isn't the final protocol implementation, nonetheless I want to point out that this update happens at a different point and with different locking in the proven protocol.

migration/idempotent/011-maintenance.sql
```diff
@@ -170,8 +190,8 @@ BEGIN
         IF log_verbose THEN
             RAISE LOG 'promscale maintenance: data retention: metric %: done dropping chunks in %', metric_name, pg_catalog.clock_timestamp() OPERATOR(pg_catalog.-) lastT;
         END IF;
-        SELECT current_epoch, last_update_time INTO present_epoch, last_updated FROM
-            _prom_catalog.ids_epoch LIMIT 1;
+        SELECT e.current_epoch, e.delete_epoch INTO present_epoch, delete_epoch FROM
```
sumerman, line 193: I'm not sure this line is even necessary now, when 183 doesn't reference it.

JamesGuthrie:
This PR has been superseded by #529

PR #515, opened 2022-09-20
Make timescaledb mandatory

JamesGuthrie:
## Description

This change adds a `requires` entry to the extension control file,
making the `timescaledb` extension mandatory for installation.

We omit the `requires` entry when running `cargo pgx test`, because of
the (unnecessary) burden that it imposes on developer setup.

## Merge requirements

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -24,6 +24,7 @@ struct IdempotentWrapper<'a> {
 #[derive(Template)]
 #[template(path = "promscale.control", escape = "none")]
 struct ControlFile {
+    requires_timescaledb: bool,
     is_pg_12: bool,
 }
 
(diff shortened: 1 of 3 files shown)


build.rs
```diff
@@ -77,15 +78,23 @@ fn generate_promscale_000_sql(manifest_dir: &str) {
...
+///  - When we run `cargo pgx test`, we don't want a dependency on the timescaledb extension.
 fn generate_control_file(manifest_dir: &str) {
     let features: std::collections::HashSet<String> = env::vars()
         .filter(|(k, _)| k.starts_with("CARGO_FEATURE_"))
         .map(|(k, _)| k.replace("CARGO_FEATURE_", "").to_ascii_lowercase())
         .collect();
     let is_pg_12 = features.contains(&"pg12".to_string());
-    let control_file_contents = ControlFile { is_pg_12 }.render().unwrap();
+    let is_pg_test = features.contains(&"pg_test".to_string());
+    let control_file_contents = ControlFile {
```
sumerman, line 92: This is extremely unreliable because `build.rs` isn't always executed. This "works" for pg12 primarily because we don't switch to it very often. 

I would rather have `requires = 'timescaledb'` only when releasing.

JamesGuthrie: > This is extremely unreliable because `build.rs` isn't always executed.

I just went down the fun path of (re?) understanding this statement. You are right, but we do have the following workaround in `build.rs`:

```Rust
    // Forces build.rs to run every time and generate control file anew.
    // Otherwise it runs only once per feature combination, making it
    // difficult to switch back and forth between PG versions.
    // We should remove this as soon as we drop PG 12 support.
    println!("cargo:rerun-if-changed={}", CONTROL_FILE_NAME);
```

As long as we have that workaround, I think that we should be fine.

I would like to make the (admittedly lazy) proposal to push fixing this out until we drop PG12 support. Then we will no longer be generating the .control file depending on the postgres version, and can adopt your suggestion of adding the dependency at install time. WDYT?

.github/workflows/docker.yaml
```diff
@@ -159,7 +159,8 @@ jobs:
       - docker
       - pick-connector-branch
     if: needs.docker.result == 'success'
-    uses: timescale/promscale/.github/workflows/go-e2e.yml@master
+    # TODO (james): using jg/make-timescaledb-mandatory necessary to get this to work
+    uses: timescale/promscale/.github/workflows/go-e2e.yml@jg/make-timescaledb-mandatory
```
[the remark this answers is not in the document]
JamesGuthrie: Yes.

build.rs
sumerman: It doesn't work in 100% of cases, even with the aforementioned workaround, and I don't know exactly why. Hence `promscale.control` make target. You are probably right, and we should leave it to `build.rs` for now, yet I would still suggest to switch based on more than just `pg_test` feature. (E.g. "not pg_test and release". ) For instance, I don't think `cargo pgx run` would pass `pg_test` yet it's still useful for local development.

JamesGuthrie: Thanks, you're right! I've reworked this again. Please take another look.

sumerman (approved):

PR #520, opened 2022-09-27
Drop non empty span name check constraint from `_ps_trace.operation` table

arajkumar:
## Description

Our tracing implementation is based on OTLP spec, though OTLP spec doesn't mandate span name to be non empty[1], we have a constraint in the _ps_trace.operation table which raises error on empty span name. On the other hand, Jaeger span name can be empty and their storage integration tests validates the same[2]. Jaeger author even suggested to remove[3] the constraint as it shouldn't trouble any part of the system.

[1] https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L110-L121 
[2] https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/integration/fixtures/traces/default.json#L6 
[3] https://github.com/jaegertracing/jaeger/pull/3922#issuecomment-1256505785

Fixes https://github.com/timescale/promscale/issues/1621#issuecomment-1239182178

## Merge requirements

### DIFF
--- a/migration/incremental/032-remove-non-empty-span-name-constraint.sql
+++ b/migration/incremental/032-remove-non-empty-span-name-constraint.sql
@@ -1,0 +1,10 @@
+-- Our tracing implementation is based on OTLP spec, though OTLP spec
+-- doesn't mandate span name to be non empty[1], we have a constraint
+-- in the _ps_trace.operation table which raises error on empty span name.
+-- On the other hand, Jaeger span name can be empty and their storage
+-- integration tests validates the same[2]. Jaeger author even suggested to
+-- remove[3] the constraint as it shouldn't trouble any part of the system.
+-- [1] https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L110-L121
+-- [2] https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/integration/fixtures/traces/default.json#L6
+-- [3] https://github.com/jaegertracing/jaeger/pull/3922#issuecomment-1256505785
+ALTER TABLE _ps_trace.operation DROP CONSTRAINT operation_span_name_check;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ We use the following categories for changes:
 
 - Correctly remove all expired series [#521]
 
+### Changed
+
+- Drop non empty span name check constraint from `_ps_trace.operation` table [#520]
+
 ## [0.6.0] - 2022-08-25
 
 ### Added


migration/incremental/032-remove-non-empty-span-name-constraint.sql
```diff
@@ -0,0 +1,14 @@
...
+-- in the _ps_trace.operation table which raises error on empty span name.
+-- On the other hand, Jaeger span name can be empty and their storage
+-- integration tests validates the same[2]. Jaeger author even suggested to
+-- remove[3] the constraint as it shouldn't trouble any part of the system.
+-- [1] https://github.com/open-telemetry/opentelemetry-proto/blob/724e427879e3d2bae2edc0218fff06e37b9eb46e/opentelemetry/proto/trace/v1/trace.proto#L110-L121
+-- [2] https://github.com/jaegertracing/jaeger/blob/7872d1b07439c3f2d316065b1fd53e885b26a66f/plugin/storage/integration/fixtures/traces/default.json#L6
+-- [3] https://github.com/jaegertracing/jaeger/pull/3922#issuecomment-1256505785
+
+-- `_ps_trace.operation` table has been created with `span_name text NOT NULL CHECK (span_name != '')`, we just want to remove the `CHECK(span_name != '')`.
+-- Ideally we should have used 'ALTER TABLE <table> ALTER COLUMN <column> DROP NOT NULL', but it is not working with a column having 'CHECK' constraint, hence dropping the CONSTRAINT using implicit name and readding `NOT NULL` CONSTRAINT.
```
cevian, line 12: I am confused by the DROP NOT NULL comment. Don't we want the column def to be `span_name text NOT NULL` why do we need to DROP NOT NULL then? don't we just want to drop the check constraint but **keep** the not null constraint.

migration/incremental/032-remove-non-empty-span-name-constraint.sql
```diff
@@ -0,0 +1,14 @@
...
+ALTER TABLE _ps_trace.operation DROP CONSTRAINT operation_span_name_check;
+ALTER TABLE _ps_trace.operation ALTER COLUMN span_name SET NOT NULL;
```
cevian, line 14: why would the set not null be necessary. Isn't the column already NOT NULL?
```suggestion

```

migration/incremental/032-remove-non-empty-span-name-constraint.sql
arajkumar: Yes, we need `NOT NULL`, but I couldn't selectively drop only the `CHECK`. Thats why I removed the entire constraint which checks both `NOT NULL` and `CHECK..` and added later just for `NOT NULL`.

Ref: https://stackoverflow.com/questions/58407194/how-to-disable-check-constraint-in-postgresql

migration/incremental/032-remove-non-empty-span-name-constraint.sql
jgpruitt: image

It appears to me that dropping the check constraint does not impact the not null constraint. I think we should remove line 14.

jgpruitt:
Note: this means that ALL spans of a given service and span kind that have an empty span name will map to the same single operation id.

jgpruitt (changes requested):
Please add a note in the CHANGELOG about this

jgpruitt (approved):

cevian (approved):

PR #524, opened 2022-09-28
Adjust security checks

cevian:
Allow DB owners to opt out of security checks. This is necessary for privileged users on cloud to have the option of opt-ing out.

## Merge requirements


jgpruitt (approved):

JamesGuthrie (changes requested):
Can you add a CHANGELOG entry for this.

JamesGuthrie (approved):

PR #529, opened 2022-10-04
WIP: logical to time epoch round 2

JamesGuthrie:
## Description

TODO:

- validate that the logic is actually close to the spec
- clean up the control flow and organize it better

## Merge requirements


migration/idempotent/001-base.sql
```diff
@@ -866,7 +866,7 @@ AS
 $$
 BEGIN
     EXECUTE FORMAT(
-        'UPDATE prom_data_series.%1$I SET delete_epoch = current_epoch+1 FROM _prom_catalog.ids_epoch WHERE delete_epoch IS NULL AND id = ANY($1)',
+        'UPDATE prom_data_series.%1$I s SET delete_epoch = current_epoch FROM _prom_catalog.ids_epoch WHERE s.delete_epoch IS NULL AND id = ANY($1)',
```
sumerman, line 869: 1. Is it possible to rename the `s.delete_epoch` column? I would prefer to if there are no technical issues with it.
2. I don't think this is correct w.r.t our locking strategy. Same as in the mark function, this need at least a `SELECT ... FOR SHARE` on `ids_epoch`.

migration/idempotent/001-base.sql
```diff
@@ -2058,6 +2058,37 @@ data newer than `newer_than` in any metric table.
...
+    VOLATILE
+    SET search_path = pg_catalog, pg_temp
+AS $func$
+    DECLARE
+        epoch BIGINT;
+        rows_affected INTEGER;
+    BEGIN
+        epoch := EXTRACT (EPOCH FROM ran_at);
+        UPDATE _prom_catalog.ids_epoch e SET current_epoch = epoch WHERE e.current_epoch = 0;
+        GET DIAGNOSTICS rows_affected = ROW_COUNT;
```
sumerman, line 2074: - Why use `GET DIAGNOSTICS`? Is something wrong with `UPDATE ... RETURNING`? 
- I think always performing the `SELECT current_epoch` is safer.

migration/idempotent/001-base.sql
```diff
@@ -2099,6 +2131,8 @@ BEGIN
     SELECT drop_point OPERATOR(pg_catalog.+) pg_catalog.interval '1 hour'
     INTO check_time;
 
+    SELECT current_epoch FROM _prom_catalog.ids_epoch FOR SHARE INTO STRICT locally_observed_epoch;
```
sumerman, line 2134: Even though `FOR SHARE` is sufficient here and is what I suggested in the spec, keeping our mxid struggles in mind, I would perhaps upgrade it to `FOR UPDATE`. I don't expect a ton of concurrency on this lock, but it should be easier on IO and vacuum.

migration/idempotent/011-maintenance.sql
```diff
@@ -131,11 +129,10 @@ BEGIN
             ORDER BY range_end DESC
             LIMIT 1;
         END IF;
-        -- end this txn so we're not holding any locks on the catalog
-
-        SELECT current_epoch, last_update_time INTO present_epoch, last_updated FROM
-            _prom_catalog.ids_epoch LIMIT 1;
+        PERFORM _prom_catalog.initialize_current_epoch(ran_at);
```
sumerman, line 132: My understanding this is necessary because the first `_prom_catalog.delete_expired_series` is conditional making it possible for `_prom_catalog.mark_series_to_be_dropped_as_unused` to start on `current_epoch` 0. 

It would be easier for me to reason about this if `_prom_catalog.mark_series_to_be_dropped_as_unused` simply punted when `current_epoch` is 0, letting either the second `_prom_catalog.delete_expired_series` or the connector initialize it.

sumerman (changes requested):
`_prom_catalog.delete_series_catalog_row` is what concerns me the most, the rest is debatable or just questions.

I'm also not 100% sure about epoch initialization, I'll need to spend more time looking at the connector's code

migration/idempotent/001-base.sql
JamesGuthrie: I have opted to always perform `SELECT current_epoch`, as you suggested.

migration/idempotent/001-base.sql
JamesGuthrie: Done.

migration/idempotent/011-maintenance.sql
JamesGuthrie: I'm not sure if that really works.

Part of the problem is that there is no expired series cleanup process. If we don't mark a series as unused in `_prom_catalog.mark_series_to_be_dropped_as_unused`, but then remove all associated chunks for the series in the subsequent call to `_prom_catalog.drop_metric_chunk_data`, then the empty series will never be removed.

If we punt in `_prom_catalog.mark_series_to_be_dropped_as_unused` then this will happen. Granted, the likelihood is low, but it could still happen.

migration/idempotent/001-base.sql
```diff
@@ -2136,110 +2166,149 @@ This function is designed to be used in the context of dropping metric
...
-                --do the global check to confirm
-                SELECT label_id
-                FROM check_local_series
-                WHERE NOT EXISTS (
-                    SELECT 1
-                    FROM  _prom_catalog.series series_exists
-                    WHERE series_exists.labels && ARRAY[label_id]
-                    LIMIT 1
-                )
+    --jit interacts poorly why the multi-partition query below
```
antekresic, line 2270: Indentation seems to be wrong here.

migration/idempotent/001-base.sql
sumerman: 1. Yeah, that would be the best, in my opinion.
2. 👍🏼

migration/idempotent/001-base.sql
JamesGuthrie: done

migration/idempotent/001-base.sql
JamesGuthrie: @sumerman I've hit an issue: the upgrade tests are not happy, because the index definition is not the same on the "fresh install" and "update" paths. The reason is that the attribute which belongs to the index is not renamed. I could reach into pg_attribute and rename this, but I'm wondering if it's worth it. Can you take a look at what I have already and let me know.

sumerman (approved):
Just a couple of nits about comments. Otherwise LGTM. I've played a little with the model and now feel confident about the suggested epoch initialization mechanism.

migration/idempotent/001-base.sql
```diff
@@ -2136,59 +2169,106 @@ This function is designed to be used in the context of dropping metric
...
+    COMMIT;
+
+    -- reset search path after transaction end
+    SET LOCAL search_path = pg_catalog, pg_temp;
+
+    -- Now we recheck the delete conditions, and delete series.
+    -- This corresponds to the ActuallyDeleteTx in our model.
+    CALL _prom_catalog._actually_delete_series_and_labels(metric_schema, metric_table, metric_series_table, deletion_epoch);
+    -- Now we check if there are any labels which we can remove.
+    -- This is not reflected in the model.
```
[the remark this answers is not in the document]
JamesGuthrie: Indeed!

sumerman (changes requested):
I believe this PR needs to be against `feature/series-cache-invalidation` branch to be in sync with the connector repo

niksajakovljevic:
I didn't get into implementation but mostly reading the description. So if I understand correctly this implementation should enable us to only remove specific series ids that have been deleted? One of the bigger problems with existing implementation is that on epoch change we just reset the whole cache in the connector which is really bad from the memory perspective.

sumerman (approved):

sumerman (approved):

Issue #527, opened 2022-10-03
Release 0.7.0


JamesGuthrie:
## Pre-release
- [x] Create a git branch named `pre-release-x.x.x`
- [x] Ensure `upgradeable_from` in `templates/promscale.control` contains the previously released version.
- [x] Update `CHANGELOG.md` with release version and date, ensuring that all relevant changes are reflected.
- [x] Update the version in all places with `./update-version.sh <version-here>`
- [x] Freeze any new incremental sql scripts released in this version in `incremental_freeze_test`
- [x] Create a PR from the `pre-release-x.x.x` branch. Get it approved. Merge the PR to master.

## Release
- [x] Create a tag on the master branch `git tag <new version>`
- [x] Push the tag `git push origin <new version>`
- [x] CI will trigger and create a draft release with assets attached.
- [x] Prepare extension release notes, share with team asking for feedback.
- [x] Attach release notes to draft release created above.
- [x] Wait for CI to generate the packages files
- [x] Create a PR to update the HA image in [timescaledb-docker-ha](https://github.com/timescale/timescaledb-docker-ha). [EXAMPLE](https://github.com/timescale/timescaledb-docker-ha/pull/285/files)
- [x] In the timescaledb-docker-ha repo, update `TIMESCALE_PROMSCALE_EXTENSIONS` in the `Makefile` to include the version just released
- [x] Update the CHANGELOG entry in the timescaledb-docker-ha repo and wait for the CI to complete and request review from the Cloud team and merge it when approved.
- [x] Create a new PR in the timescaledb-docker-ha repo. Stamp the version in the CHANGELOG. Merge it with master and push the correct tag to trigger CI ([see instructions in the repo](https://github.com/timescale/timescaledb-docker-ha#release-process)) [EXAMPLE](https://github.com/timescale/timescaledb-docker-ha/pull/286/files)
- [x] Publish the GitHub release on the promscale_extension repo.

## Post-Release
- [x] Create a new git branch named `post-release-x.x.x`
- [x] Run `make post-release` which will generate the `sql/promscale--x.x.x.sql` file of the version just released and create all the upgrade path sql files.
- [x] Add and commit the newly created sql files to git. They are ignored by default. e.g. `git add sql/*--0.5.5.sql --force`
- [x] Determine the development version (determined by bumping the patch version and appending `-dev`)
- [x] Set the version in all places necessary with `./update-version.sh <develop-version>`
[...]

PR #535, opened 2022-10-11
Post-release 0.7.0

sumerman:
## Description

- Post-release SQL bits.
- Switching to the post-release dev version.
- Resolves https://github.com/timescale/promscale_extension/issues/527

## Merge requirements

Waiting for `timescaledev/promscale-extension:0.7.0-ts2.8.1-pg` images


sumerman:
In the interest of unblocking further merges in this repo, I suppose it'd be acceptable to merge even before we have `timescaledev/promscale-extension:0.7.0-ts2.8.1-pg` images ready. We can update the test config when they are.

INSTALL.md
```diff
@@ -104,7 +104,7 @@ For bare-metal installations, the full instructions for setting up PostgreSQL, T
     sudo apt-get install -y git
     git clone https://github.com/timescale/promscale_extension
     cd promscale_extension
-    git checkout 0.7.0
+    git checkout 0.7.1-dev
```
cevian, line 107: why are we changing the instructions to install a non-release? Also, I don't think thats an actual tag.

Cargo.toml
```diff
@@ -22,7 +22,7 @@ panic = "unwind"
 # If your tests fail with SIGSEGV try and use 1.64 or newer.
 # Sadly, we can't just drop it because it's a workaround for
 # another issue https://github.com/tcdi/pgx/pull/208
-lto = "thin"
+#lto = "thin"
```
cevian, line 25: why are we turning lto off?

cevian (changes requested):

Cargo.toml
sumerman: I keep accidentally committing it. Thanks for catching it!

INSTALL.md
sumerman: You are correct. `./update-version.sh` does it, but it needs to leave `INSTALL.md` alone on the post-release run. I'll fix it.

cevian (approved):

PR #538, opened 2022-10-17
Adds support for creating metric rollup based on metric types

Harkishen-Singh:

In relation with https://github.com/timescale/promscale/pull/1703 & https://github.com/timescale/o11y-team-applications/issues/228

## Description

This commit does 3 things on a high level.
1. create_metric_rollup is a straightforward that creates schema based on rollup name, prefixed with 'ps_'.

2. Functions to create metric rollup depending on metric type. This is done by
a function scan_for_new_rollups that runs every 30 mins. This function looks for
pending metrics (by peeking inside metric_with_rollup table) that need rollups
and creates them. The creation path is
scan_for_new_rollups -> create_metric_rollup_view (decides the metric type) -> create_rollup_for_{gauge|counter|summary}.
Counter and Histogram have same columns, hence handled by the same function.

3. Adds set of utility functions like counter_reset_sum and irate, which are required for rollup creation. These
are temporary functions and they MUST be removed before the MVP.

Note: These functions are just dummies, their behaviour is not 100% correct.
Example: These functions require ordered values array, but we cannot do ORDER BY in Caggs query.
Hence, these functions will be handled by SQL aggregate written in Rust.

## Merge requirements


migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,282 @@
+-- create_metric_rollup prepares the given rollup. The actual rollup views are created by _prom_catalog.scan_for_new_rollups()
+CREATE OR REPLACE PROCEDURE _prom_catalog.create_metric_rollup(name TEXT, resolution INTERVAL, retention INTERVAL) AS
+$$
+    DECLARE
+        schema_name TEXT := 'ps_' || name;
```
antekresic, line 5: `ps_` should be constified.

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,282 @@
...
+-- 3. Delete the actual Cagg views that represent the rollups. This is done by directing deleting the entire rollup schema
+CREATE OR REPLACE PROCEDURE _prom_catalog.delete_metric_rollup(rollup_name TEXT) AS
+$$
+    DECLARE
+        rollup_schema_name TEXT;
+
+    BEGIN
+        EXECUTE FORMAT('SELECT schema_name FROM _prom_catalog.rollup WHERE name = %L', rollup_name) INTO rollup_schema_name;
+        IF rollup_schema_name IS NULL THEN
+            RAISE EXCEPTION '[Rollup] Cannot delete metric rollup due to rollup_schema_name being NULL';
```
antekresic, line 33: This message might be a bit confusing when read outside of the context of this function (i.e. in a log file).

Maybe something like this is better:
`Cannot delete metric rollup [rollup name], missing schema [schema name]`

migration/idempotent/017-metric-rollup.sql
antekresic, line 5: Let's create a constant for the schema name prefix.

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,282 @@
...
+CREATE OR REPLACE PROCEDURE _prom_catalog.scan_for_new_rollups(job_id int, config jsonb) AS
+$$
+DECLARE
+    r RECORD;
+    m RECORD;
+    new_rollups_created INTEGER := 0;
+    rollup_view_created BOOLEAN;
+
+BEGIN
+    IF ( SELECT _prom_catalog.get_default_value('automatic_downsample') != 'true' ) THEN
```
antekresic, line 53: Why not use created helper function here? `get_automatic_downsample`

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,282 @@
...
+LANGUAGE SQL;
+COMMENT ON FUNCTION prom_api.set_downsample_resolution(TEXT)
+    IS 'Set downsample resolution for metric rollups. This must be a comma separated string of label:resolution:retention';
+GRANT EXECUTE ON FUNCTION prom_api.set_downsample_resolution(TEXT) TO prom_admin;
+
+CREATE OR REPLACE FUNCTION prom_api.get_downsample_resolution()
+    RETURNS TEXT
+    SET search_path = pg_catalog, pg_temp
+AS $func$
+    SELECT _prom_catalog.get_default_value('downsample_resolution')::TEXT;
```
antekresic, line 130: I don't see an initial default for this. Is that intended?

Harkishen-Singh: Yes, this is intended, but I am not confident that this is the correct pattern.

The idea behind this is, `downsample_resolution` (example value `short:5m:90d,long:1h:365d`) will be set by connector (by seeing the `dataset-config`). We need to store this somewhere in the DB, so that when the user changes the downsample (in `dataset-config`) to something like `short:5m:90d,very_long:1d:365d`, then connector deletes the `long` based downsample and creates the `very_long` downsample resolution. Hence, we need a place to store the current resolution state the DB contains.

There should not be any entry in `_prom_catalog.initial_defaults` view, since when Promscale starts, we do not have any downsampling. Creation of downsampling is done by connector.

Imagine, if we had a initial default for downsample_resolution, then Promscale starts and tries to delete the downsampling if the dataset-config is saying a different resolution than a default, and that creates an exception in the database.

I didn't wanna create a new table just to keep the state, hence used the defaults table.

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,290 @@
...
+AS
+$$
+BEGIN
+    EXECUTE FORMAT(
+        'CREATE MATERIALIZED VIEW %1$I.%2$I WITH (timescaledb.continuous, timescaledb.materialized_only=true) AS
+            SELECT
+                timezone(
+                    %3$L,
+                    time_bucket(%4$L, time) AT TIME ZONE %3$L + %4$L
+                ) as time,
```
jgpruitt, line 233: can you explain what is happening here with timezones? I'm not following what's supposed to happen

migration/incremental/033-metric-rollup.sql
```diff
@@ -0,0 +1,18 @@
...
+    schema_name    TEXT NOT NULL,
+    resolution     INTERVAL NOT NULL,
+    retention      INTERVAL NOT NULL
+);
+GRANT SELECT ON TABLE _prom_catalog.rollup TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _prom_catalog.rollup TO prom_writer;
+
+CREATE TABLE _prom_catalog.metric_with_rollup (
+    rollup_schema   TEXT,
+    metric_name     TEXT,
```
jgpruitt, line 12: why not use `metric_id` and `rollup_id`?

migration/incremental/033-metric-rollup.sql
```diff
@@ -0,0 +1,18 @@
...
+    table_name      TEXT,
```
jgpruitt, line 13: should these columns be NOT NULL?

jgpruitt (changes requested):
Good work!

If you would, it would be nice to define the functions/procedures above where they are first referenced. That way, if they are switched from plpgsql to SQL they work without having to be rearranged.

migration/idempotent/017-metric-rollup.sql
Harkishen-Singh: > ach downsample resolution could be a separate row instead of using a comma-separated string

Ah, now I relate that i am already creating a `_prom_catalog.rollup` table that can be used for doing exactly this. It has all the information that the downsample resolution string needs i.e., label, resolution, retention. So, we should not store any of this in defaults table.

Thank you for the pointer.

migration/idempotent/017-metric-rollup.sql
Harkishen-Singh: Sure, I have tried to keep the semantics similar to what's done [here](https://docs.timescale.com/promscale/latest/downsample-data/caggs/#example-of-a-promscale-continuous-aggregate) (See the note).

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,290 @@
...
+    END LOOP;
+
+    COMMIT;
+
+    -- Refresh the newly added rollups for entire duration of data.
+    FOR r IN
+        SELECT rollup_schema, table_name FROM _prom_catalog.metric_with_rollup WHERE refresh_pending = TRUE
+    LOOP
+        CALL refresh_continuous_aggregate(r.rollup_schema || '.' || r.table_name, NULL, NULL);
+        UPDATE _prom_catalog.metric_with_rollup SET refresh_pending = FALSE WHERE rollup_schema = r.rollup_schema AND table_name = r.table_name;
```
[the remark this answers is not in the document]
Harkishen-Singh: Yes, I was thinking in that direction when I was programing this. My reasoning was that this will do frequent flushing (so flush batch sizes won't be huge) and avoid too much buffering in memory. Let's do `COMMIT` after every refresh.

jgpruitt:
Hey, can we get some tests added for the new stuff, too, please?

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,274 @@
...
+AS
+$$
+BEGIN
+    EXECUTE FORMAT(
+            'CREATE MATERIALIZED VIEW %1$I.%2$I WITH (timescaledb.continuous, timescaledb.materialized_only=true) AS
+                SELECT
+                    timezone(
+                        %3$L,
+                        time_bucket(%4$L, time) AT TIME ZONE %3$L + %4$L
+                    ) as time,
```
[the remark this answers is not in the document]
Harkishen-Singh: Does that mean we don't need this conversion, since we are storing TIMESTAMPTZ by default?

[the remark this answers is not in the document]
jgpruitt: I'm not sure. That's what I'm thinking. I didn't get a chance to test it yesterday. I'll try again today.

migration/idempotent/017-metric-rollup.sql
jgpruitt: okay, here's what i tested:

```

create table x (i int, t timestamptz);

insert into x (i, t) 
select 
    x,
    'today'::timestamptz + (interval '1 minute' * x) 
from generate_series(0, 60 * 24) x
;

select 
    time_bucket('1 hour'::interval, t) as hr,
    count(i) as nbr
from x
group by 1
order by 1
;


-- does time_bucket return a timestamptz when passed a timestamptz?
select 
    pg_typeof(time_bucket('1 hour'::interval, t)) as typeof,
    count(i) as nbr
from x
group by 1
order by 1
;
/*
┌──────────────────────────┬──────┐
│          typeof          │ nbr  │
├──────────────────────────┼──────┤
│ timestamp with time zone │ 1441 │
└──────────────────────────┴──────┘
*/


-- is it still a timestamptz if we pass in a timezone?
select 
    pg_typeof(time_bucket('1 hour'::interval, t, 'UTC')) as typeof,
    count(i) as nbr
from x
group by 1
order by 1
;
/*
┌──────────────────────────┬──────┐
│          typeof          │ nbr  │
├──────────────────────────┼──────┤
│ timestamp with time zone │ 1441 │
└──────────────────────────┴──────┘
*/
```

jgpruitt: so, if you pass in a `timestamptz` you will be returned a `timestamptz`. No need to add the `AT TIME ZONE` or wrap in the `timezone(...)` function

jgpruitt: it will by default be rendered in the `show timezone` timezone

jgpruitt: additionally, at the moment you are hardcoding the `timezone` parameter to the `time_bucket` function to `'UTC'`

- this is the default, so there's no need to pass it
- should we hardcode it to `'UTC'`, or should we use the database's `timezone` setting? does it matter? I'm not sure

Harkishen-Singh: Thank you for looking into this.

I think all the `AT TIME ZONE` and wrapping with `timezone()` was done basically to ensure that we can correctly add the `+'1 hour'` (done [here](https://docs.timescale.com/promscale/latest/downsample-data/caggs/#example-of-a-promscale-continuous-aggregate)) for this reason `This is done to match the PromQL semantics of representing a bucket with the timestamp at the end of the bucket instead of the start of the bucket.`.

Now, from the above analysis, I think we can simply change
```postgresql
timezone('UTC', 
          time_bucket('1 hour', time) AT TIME ZONE 'UTC' +'1 hour')  
            as time
```
to
```postgresql
time_bucket('1 hour', time) + '1 hour' as time
```

Am I understanding this right?

Harkishen-Singh: Okay, so when I made the change and again tested the rollup creation, I got this error

```
```

Harkishen-Singh: Maybe thats all the reason for having the `timezone()` thing, @antekresic if you can take a look and share the idea behind using timezone [here](https://docs.timescale.com/promscale/latest/downsample-data/caggs/#example-of-a-promscale-continuous-aggregate).

Harkishen-Singh:
Added `sql-tests` for metric rollups.

migration/idempotent/017-metric-rollup.sql
jgpruitt: that's weird. `time_bucket` IS immutable. I wonder if `+` is not immutable...
image

jgpruitt: It's the addition :(
image

jgpruitt: https://www.postgresql.org/message-id/CANWftzJNW1hUZmRN9=tWn=K=bQrUhAOxkcogVNC72STp5-_czQ@mail.gmail.com

jgpruitt: In light of this, I think the original pattern (though unfortunate) makes sense. Sorry to sidetrack you for a bit.

Harkishen-Singh: Not at all, this was an interesting investigation. But happy that we understood the root cause. Thank you!

migration/incremental/033-metric-rollup.sql
```diff
@@ -0,0 +1,19 @@
+CREATE TABLE _prom_catalog.rollup (
+    id             SERIAL,
```
jgpruitt, line 2: The id should be the primary key, and the name should have a unique constraint.

migration/incremental/033-metric-rollup.sql
```diff
@@ -0,0 +1,19 @@
...
+    name           TEXT NOT NULL PRIMARY KEY,
+    schema_name    TEXT NOT NULL,
+    resolution     INTERVAL NOT NULL,
+    retention      INTERVAL NOT NULL
+);
+GRANT SELECT ON TABLE _prom_catalog.rollup TO prom_reader;
+GRANT SELECT, INSERT, UPDATE, DELETE ON TABLE _prom_catalog.rollup TO prom_writer;
+
+CREATE TABLE _prom_catalog.metric_rollup (
+    id              SERIAL,
```
jgpruitt, line 12: The id should be the primary key, and the (rollup_id, metric_id) should have a unique constraint.

migration/incremental/033-metric-rollup.sql
```diff
@@ -0,0 +1,19 @@
...
+    rollup_id       INTEGER NOT NULL,
```
jgpruitt, line 13: should there be a foreign key constraint on rollup_id to rollup.id?

sql-tests/tests/snapshots/tests__testdata__chunks_to_freeze.sql.snap
```diff
@@ -14,7 +14,7 @@ expression: query_result
 
  add_compression_policy 
 ------------------------
-                   1005
+                   1006
```
jgpruitt, line 17: do you know why this number changed?

sql-tests/tests/snapshots/tests__testdata__chunks_to_freeze.sql.snap
```diff
@@ -24,7 +24,7 @@ expression: query_result
 
  add_compression_policy 
 ------------------------
-                   1006
+                   1007
```
jgpruitt, line 27: do you know why this number changed?

jgpruitt (changes requested):
super close. just a few minor things

jgpruitt:
Oh, and we need an entry in the CHANGELOG

migration/idempotent/017-metric-rollup.sql
```diff
@@ -0,0 +1,277 @@
...
+        _rollup_id INTEGER;
+        _rollup_schema_name TEXT;
+
+    BEGIN
+        SELECT id, schema_name INTO _rollup_id, _rollup_schema_name FROM _prom_catalog.rollup WHERE name = _rollup_name;
+        IF _rollup_id IS NULL THEN
+            RAISE EXCEPTION '% rollup not found', _rollup_name;
+        END IF;
+        DELETE FROM _prom_catalog.rollup WHERE id = _rollup_id;
+        DELETE FROM _prom_catalog.metric_rollup WHERE rollup_id = _rollup_id;
```
jgpruitt, line 190: now that there's a foreign key constraint, you need to swap the order in which these are executed. delete from `metric_rollup` first.

jgpruitt (changes requested):
pgspot has identified a number of issues that need to be resolved as well

JamesGuthrie:
> `pgspot` is happy when I add `SET search_path = pg_catalog, pg_temp` to all the functions that i introduced. But, that creates another problem, because then I cannot do `COMMIT;` (SET does not allow COMMIT in procedure) in `_prom_catalog.scan_for_new_rollups()` which is important for Caggs to be created in a loop.

This is a "known issue", with a known workaround. It is covered in point 5 of the migration [readme](https://github.com/timescale/promscale_extension/blob/master/migration/README.md), but on re-reading it I realise that it might not be explicit enough. You'll notice that some functions do `SET LOCAL search_path ...` _in the body_ of the function, e.g.:

```
    -- Note: We cannot use SET in the procedure declaration because we do transaction control
    -- and we can _only_ use SET LOCAL in a procedure which _does_ transaction control
    SET LOCAL search_path = pg_catalog, pg_temp;
```

You will need to re-set the search path after each `COMMIT` in the body of the procedure.

jgpruitt (approved):

Issue #495, opened 2022-08-30
Implement a data buffer for internal metrics and logs data


sumerman:
Implement support for collecting data from rolled-back transactions in Rust using shared memory.

I suggest implementing an API in Rust that would allow jobs to post metrics data and the connector to collect it via SQL function calls. Internally the data could be stored in a shared memory buffer. A similar technique could work for logs as well. But it would require allocating a ring buffer, and logs may demand more memory than we would be comfortable allocating in shmem.

sumerman:
See https://github.com/timescale/promscale_extension/pull/555 for details

PR #552, opened 2022-11-02
A data buffer for job metrics and logs

sumerman:
## Description

This PR introduces a Rust-based escape hatch for backend metrics (and, in the future, logs) data `backend_telemetry_buffer`. 

Addresses #495 

## Merge requirements


sumerman:
I validated the absence of serious memory leaks by running:

```SQL
explain analyse SELECT _prom_ext.push_rec(ROW(now(), x)::_prom_ext.backend_telemetry_rec) FROM generate_series(1, 200000) x;
SELECT COUNT(*) FROM _prom_ext.pop_recs();
```

After executing the first line, the RAM consumption always spiked by 15 MB and reliably returned to the original value by the end of the test.

sumerman:
`CALL run_job(:job_id);` doesn't seem to be using the actual maintenance job infrastructure. In order to prove the buffer works inside an actual job worker I repeated the integration test on a regular instance.

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,169 @@
...
+            let remaining = self.buffer.list.len() - self.pos;
+            (remaining, Some(remaining))
+        }
+    }
+
+    impl<'a> Drop for BufferIter<'a> {
+        fn drop(&mut self) {
+            // Drop on the list calls list_free(). For it to work without segfaults we first 
+            // need to replace the list (triggering the Drop) and only then call mem_ctx.reset()
+            self.buffer.list = PgMemoryContexts::CacheMemoryContext.switch_to(|_| PgList::new());
```
cevian, line 63: can't this be an assignment to NIL instead?

sumerman: `PgList::new()` is the `NIL`

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,169 @@
...
+    ) -> bool {
+        let buffer_mutex = init_buffer();
+        let r_datum = pg_getarg_datum_raw(fcinfo, 0);
+        {
+            let mut q_guard = buffer_mutex.lock().unwrap();
+            let prev_ctx = q_guard.mem_ctx.set_as_current();
+            let detoasted = unsafe {
+                pg_sys::pg_detoast_datum_copy(r_datum.cast_mut_ptr() as *mut pg_sys::varlena)
+            };
+            q_guard.list.push(detoasted.cast());
```
cevian, line 99: 2 questions:
- shouldn't this go through a push function in Buffer. Not sure I love accessing the internal list element out here. 
- should we limit the size of the buffer to last X items. They way this is done, the list can grow indefinitely. I always imagined this was a ring buffer

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,169 @@
...
+        "#,
+        name = "backend_telemetry_rec_decl"
+    );
+
+    const COMPOSITE_TYPE_NAME: &str = "_prom_ext.backend_telemetry_rec";
+
+    type BufferItem = composite_type!(COMPOSITE_TYPE_NAME);
+    struct Buffer {
+        mem_ctx: PgMemoryContexts,
+        list: PgList<BufferItem>,
```
cevian, line 26: are we sure PgList is better than a Rust list type? Do we need a pg type here?

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,169 @@
...
+    #[pg_extern(volatile, strict, create_or_replace, requires = ["backend_telemetry_rec_decl"])]
+    pub fn push_rec(
+        _r: pgx::composite_type!(COMPOSITE_TYPE_NAME),
+        fcinfo: pg_sys::FunctionCallInfo,
+    ) -> bool {
+        let buffer_mutex = init_buffer();
+        let r_datum = pg_getarg_datum_raw(fcinfo, 0);
+        {
+            let mut q_guard = buffer_mutex.lock().unwrap();
+            let prev_ctx = q_guard.mem_ctx.set_as_current();
```
cevian, line 95: should we use switch_to(|_| ... ) instead?

cevian (commented):
Some questions

src/backend_telemetry_buffer.rs
sumerman: It won't work because a guard built into that function wouldn't let `q_guard` cross the unwind boundary. No matter, it's gone now :)

src/backend_telemetry_buffer.rs
sumerman: I was trying to keep everything allocated inside the pg's memory contexts. It matters less with a long-lived ring buffer, but since list cells are dynamically allocated, I opted for something that would be 100% gone at the context reset.

Now I switched to a statically allocated ring buffer of Datum.

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,268 @@
...
+
+        /// Empties the buffer and frees memory occupied by contained items.
+        fn reset(&mut self) {
+            self.next_idx = 0;
+            self.mem_ctx.reset();
+        }
+
+        /// Creates an iterator, traversing all initialized items
+        /// in the buffer. Calls `reset()` when dropped.
+        fn consume_as_iter(&mut self) -> BufferIter {
```
JamesGuthrie, line 97: You could replace this method with the following for a little more 🦀 -idiomism:

```rust
    impl<'a> IntoIterator for &'a mut Buffer {
        type Item = BufferItem;
        type IntoIter = BufferIter<'a>;

        #[inline]
        fn into_iter(self) -> BufferIter<'a> {
            BufferIter {
                buffer: self,
                pos: 0,
            }
        }
    }
```

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,268 @@
...
+            // We are not deallocaing anything here because Drop does it wholesale.
+            let cur_pos = self.pos;
+            self.pos += 1;
+            self.buffer
+                .inner_buffer
+                .get(cur_pos)
+                // guard against accessing uninitialized parts of the buffer
+                .filter(|_| cur_pos < self.buffer.next_idx)
+                // SAFETY:
+                // - append is expected to be the only funciton writing into the buffer,
```
JamesGuthrie, line 118:
```suggestion
                // - append is expected to be the only function writing into the buffer,
```

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,268 @@
...
+                //   therefore all elements are BufferItem
+                // - the if above ensures we don't access unintialized parts of the buffer
```
JamesGuthrie, line 120:
```suggestion
                // - the if above ensures we don't access uninitialized parts of the buffer
```

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,268 @@
...
+                .get_mut(self.next_idx % BUFFER_SIZE)
+                .and_then(|place| {
+                    copied_opt.map(|copied| {
+                        if fully_initialized {
+                            unsafe { pg_sys::pfree(place.cast_mut_ptr()) }
+                        }
+                        *place = copied;
+                    })
+                })
+                .map(|_| self.next_idx += 1)
```
JamesGuthrie, line 85: So this is _probably_ being paranoid, but what about overflow of `next_idx`?

JamesGuthrie (changes requested):
Overall I'm loving this! I have a couple of smallish comments/suggestions.

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,261 @@
...
+    use pgx::{prelude::*, *};
+    use std::{
+        mem::MaybeUninit,
+        sync::{Mutex, Once},
+    };
+
+    // Using an SQL-defined composite type allows future modifications
+    // to be carried out in an incremental migration via ALTER TYPE,
+    // without touching Rust code.
+    extension_sql!(
```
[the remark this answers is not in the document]
sumerman: Right, thanks for catching this! I intended to apply this exact treatment but forgot about it.

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,268 @@
...
+        buffer: &'b mut Buffer,
+        pos: usize,
+    }
+
+    impl Buffer {
+        fn new() -> Self {
+            let mem_ctx = PgMemoryContexts::CacheMemoryContext
+                .switch_to(|_| PgMemoryContexts::new("backend_telemetry_buffer_context"));
+            let inner_buffer = PgBox::<[Datum; BUFFER_SIZE], AllocatedByRust>::alloc0_in_context(
+                PgMemoryContexts::CacheMemoryContext,
```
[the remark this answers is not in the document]
sumerman: We never need to de-allocate `inner_buffer`. If I were to create it inside `mem_ctx` I'd have to re-allocate it every time `mem_ctx.reset()` is called.

src/backend_telemetry_buffer.rs
```diff
@@ -0,0 +1,268 @@
...
+                inner_buffer,
+                next_idx: 0,
+            }
+        }
+
+        /// Copies [`BufferItem`] into the memory context associated
+        /// with the [`Buffer`]. Frees an older item if it needs
+        /// to be overwritten due to ring buffer overflow.
+        fn append(&mut self, item: BufferItem) -> bool {
+            let copied_opt = self.mem_ctx.switch_to(|_| item.into_composite_datum());
```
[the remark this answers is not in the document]
sumerman: `.into_datum()` for `for PgHeapTuple` is `into_composite_datum()`. I assume pgx folks wanted to keep similar signatures.

src/backend_telemetry_buffer.rs
sumerman: Thanks! Will do!

src/backend_telemetry_buffer.rs
sumerman: Yes, you are 😛 

Firstly, the intended use (see `custom_func` in the SQL tests) for the buffer is unlikely to require even the 10k elements I'm allocating. 

Secondly, let's entertain the possibility of the overflow: 
- items added after the overflow will be placed at position 0 instead of an expected placement. Sure, this will slightly break the ring semantics of the buffer, but if no one bothered reading metrics data for that long, they probably wouldn't mind that extra bit of staleness 😅 
- Next, `BUFFER_SIZE` items won't be properly `pfree`'d until `mem_ctx.reset()` is called. Temporarily leaking roughly 5-10 megs of RAM is unpleasant but not critical.

sumerman: I guess adding an extra `mem_ctx.reset()` when the overflow happens isn't too complex and would prevent the potential memory leak.

migration/idempotent/017-backend-telemetry-buffer.sql
```diff
@@ -0,0 +1,22 @@
+-- The contents of this file was auto-generated by the pgx extension, but was
```
JamesGuthrie, line 1: I don't think that this file is necessary, because these definitions will be emitted by PGX.

src/backend_telemetry_buffer.rs
```diff
@@ -167,7 +173,7 @@ mod _prom_ext {
     }
 
     /// Adds a `backend_telemetry_rec` record to a backend-local ring buffer.
-    #[pg_extern(volatile, strict, create_or_replace, requires = ["backend_telemetry_rec_decl"])]
+    #[pg_extern(volatile, strict, create_or_replace, sql = false)]
```
JamesGuthrie, line 176: Sorry, my feedback in the previous review may have been confusing, I didn't mean that we should set `sql = false` here.

JamesGuthrie (changes requested):

src/backend_telemetry_buffer.rs
sumerman: It doesn't work otherwise. Generated DDL references `backend_telemetry_rec` that is now defined at a later point.

JamesGuthrie: Of course! Sorry I missed that. Hmm... That's somehow quite annoying...

JamesGuthrie (approved):

sumerman:
Closing in favor of https://github.com/timescale/promscale_extension/pull/555

PR #557, opened 2022-11-11
Add build artifact for RHEL 8

arajkumar:
## Description

So far we were build rpms for RHEL 7, lets add RHEL 8 into the list as it has been around for so long[1]. Also one of our community member has also requested an artifact for the RHEL 8.

[1] https://access.redhat.com/articles/3078#RHEL8

## Merge requirements


arajkumar:
Looks like it is not straightforward, let me dig around the dockerfile :)

JamesGuthrie:
I've addressed this in #564. Closing.

PR #559, opened 2022-11-14
Fix column "series_id" specified more than once

### LINKED ISSUE #531: Cannot create metric view when there's a tag with key `series`
When creating the view for a metric the following function is called with the `label_key` as argument:

https://github.com/timescale/promscale_extension/blob/9db8b900e8e070729e1a77c34f1fbed97ba68347/migration/idempotent/001-base.sql#L2425

The `label_key` is checked against some reserved keys, if there's a match then the key is prepended with `label_`:

https://github.com/timescale/promscale_extension/blob/9db8b900e8e070729e1a77c34f1fbed97ba68347/migration/idempotent/001-base.sql#L2433-L2438

If the key is `series` it won't match `series_id` and pass the check. The problem is that if the key is an `id` (second parameter of the function) we call:

https://github.com/timescale/promscale_extension/blob/9db8b900e8e070729e1a77c34f1fbed97ba68347/migration/idempotent/001-base.sql#L2441

Which returns `series_id` and creates the conflict:

https://github.com/timescale/promscale_extension/blob/9db8b900e8e070729e1a77c34f1fbed97ba68347/migration/idempotent/001-base.sql#L615

alejandrodnm:
## Description

When creating a metric view, the column names for the labels are created by the `_prom_catalog.get_label_key_column_name_for_view` function. If a given label matches one of the reserved keys, then it's sanitized by prefixing the returned column name with `label_`, thus avoiding a conflict.

The `_prom_catalog.get_label_key_column_name_for_view` function also accepts a second argument, it's a boolean that if true means that the `id_column_name` should be used. The `id_column_name` is created as the label name with the `_id` suffix.

One of the reserved keys that's checked by
`_prom_catalog.get_label_key_column_name_for_view` is `series_id`. If the value `series` is given as argument the served key validation passes since `series != series_id`, and for `id=false` this continues to be correct, but if `id=true` the resulting column name will be `series_id` which matches the reserved key and causes the conflict.

To avoid the conflict `series` is added to the reserved keys if `id=true`.

Fixes #531

Fixes https://github.com/timescale/promscale/issues/1685

## Merge requirements

### DIFF
--- a/sql-tests/testdata/get_label_key_column_name_for_view.sql
+++ b/sql-tests/testdata/get_label_key_column_name_for_view.sql
@@ -1,0 +1,77 @@
+\unset ECHO
+\set QUIET 1
+\i 'testdata/scripts/pgtap-1.2.0.sql'
+
+SELECT * FROM plan(20);
+
+SELECT
+  is(
+    _prom_catalog.get_label_key_column_name_for_view(key_, true),
+    format('label_%s_id', key_)::NAME,
+    format('%s is a restricted keyword and sanitized. When id=true the column name is returned suffixed with `_id`', key_)
+  )
+FROM (
+  VALUES
+    ('time'),
+    ('value'),
+    ('series_id'),
+    ('labels'),
+    ('series')
+  ) as reserved_keys (key_);
+
+SELECT is(
+  _prom_catalog.get_label_key_column_name_for_view('my-key', true),
+  'my-key_id'::NAME,
+  'when id=true the column name is returned suffixed with `_id`'
+);
+
+-- Entries in _prom_catalog.label_key are created
+SELECT is(
+  k.*,
+  ROW(id, test.key_, test.key_, format('%s_id', test.key_))::_prom_catalog.label_key,
+  format('_prom_catalog.label_key entry for %s exists', test.key_)
+)
+FROM (
+  VALUES
+    ('label_time'),
+    ('label_value'),
+    ('label_series_id'),
+    ('label_labels'),
+    ('label_series'),
+    ('my-key')
+  ) as test (key_)
+LEFT JOIN _prom_catalog.label_key k on (k.key = test.key_);
+
+SELECT is(
+  count(*),
+  6::BIGINT,
+  '6 entries were created for _prom_catalog.label_key'
+) FROM _prom_catalog.label_key;
+
+SELECT is(
+  _prom_catalog.get_label_key_column_name_for_view(key_, false),
+  format('label_%s', key_)::NAME,
+  format('%s is a restricted keyword and sanitized', key_)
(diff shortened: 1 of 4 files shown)


jgpruitt (approved):

sql-tests/testdata/get_label_key_column_name_for_view.sql
```diff
@@ -0,0 +1,77 @@
...
+SELECT is(
+  count(*),
+  6::BIGINT,
+  '6 entries were created for _prom_catalog.label_key'
+) FROM _prom_catalog.label_key;
+
+SELECT is(
+  _prom_catalog.get_label_key_column_name_for_view(key_, false),
+  format('label_%s', key_)::NAME,
+  format('%s is a restricted keyword and sanitizied', key_)
```
antekresic, line 54:
```suggestion
  format('%s is a restricted keyword and sanitized', key_)
```

antekresic (approved):

jgpruitt:
`_prom_catalog.get_label_key_column_name_for_view(label_key text, id BOOLEAN)` is called by this function with `id = false`: `_prom_catalog.create_series_view(metric_name text)` and by this function with `id = true`: `_prom_catalog.create_metric_view(metric_name text)`. In both instances it is used to determine the column names in the view that is created.

I don’t see `_prom_catalog.get_label_key_column_name_for_view(label_key text, id BOOLEAN)` called anywhere else.

* when `id = true` (metric view) the `id_column_name` is used
* when `id = false` (series view) the `value_column_name` is used

```SQL
  IF id THEN
    RETURN (_prom_catalog.get_or_create_label_key(label_key)).id_column_name;
  ELSE
    RETURN (_prom_catalog.get_or_create_label_key(label_key)).value_column_name;
  END IF;
```

* `value_column_name` = `key_name`
* `id_column_name` = `key_name` + `_id`

unless the key name is long (> 62 chars) in which case `value_column_name` is shortened and `_<key_id>` is appended and `id_column_name` is shortened to `key_name` + `_id` + `_` + `key_id` + `_id`

jgpruitt:
In short, I believe the solution you've provided is correct as is.

alejandrodnm:
Thanks @jgpruitt then the only thing is, are we ok with having for `series`:

- `label_series_id` in the case of metric view
- `series` in the case of series view

Or should we just add `series` as always reserved regardless of `id=true` to have:

- `label_series_id` in the case of metric view
- `labels_series` in the case of series view

jgpruitt:
I don't have a huge preference either way. I think maybe it would be more clear to have `series` always reserved. So, I guess I lean that way a bit.

jgpruitt (approved):

PR #561, opened 2022-11-15
Update GitHub Actions

JamesGuthrie:
## Description

GitHub deprecated some patterns which we use in our own, and in dependent GitHub Actions.

## Merge requirements

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -18,11 +18,8 @@ jobs:
       AWS_SECRET_ACCESS_KEY: ${{ secrets.PROMSCALE_EXTENSION_SCCACHE_AWS_SECRET_ACCESS_KEY }}
     steps:
       - uses: actions/checkout@v3
-      - name: Install latest nightly
-        uses: actions-rs/toolchain@v1
-        with:
-            toolchain: stable
-            override: true
+      - name: Install rust
+        uses: dtolnay/rust-toolchain@1.64.0
 
       - name: Setup sccache
         run: |
@@ -31,7 +28,7 @@ jobs:
           sudo mv sccache-*/sccache /usr/local/bin/sccache
           sccache --show-stats
 
-      - uses: Swatinem/rust-cache@v1
+      - uses: Swatinem/rust-cache@v2
 
       - name: Install cargo-pgx
         run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yaml
```diff
@@ -18,11 +18,10 @@ jobs:
       AWS_SECRET_ACCESS_KEY: ${{ secrets.PROMSCALE_EXTENSION_SCCACHE_AWS_SECRET_ACCESS_KEY }}
     steps:
       - uses: actions/checkout@v3
-      - name: Install latest nightly
-        uses: actions-rs/toolchain@v1
+      - name: Install rust
+        uses: dtolnay/rust-toolchain@master
```
sumerman, line 22: I don't feel good about depending on someone's `master`. This specific repo has a `stable` branch.

JamesGuthrie: The way that the actions in `dtolnay/rust-toolchain` work is that you either select the toolchain like this:

```
      - name: Install rust
        uses: dtolnay/rust-toolchain@master
        with:
            toolchain: 1.64.0
```

or like this:

```
      - name: Install rust
        uses: dtolnay/rust-toolchain@1.64.0
```

I figured that using `@master` and `toolchain: <version>` was easier to understand than `@<version>`. I would prefer not to use `@stable`, because that's always whatever the current rust stable version is, so it's a moving target.

I don't have a strong preference. If you would prefer not to use `master`, I would change this to:

```
      - name: Install rust
        uses: dtolnay/rust-toolchain@1.64.0
```

sumerman: can we fix it at a certain commit of their master?

JamesGuthrie: I suppose we could, but that would end up with something like:

```
  - name: Install rust
    uses: dtolnay/rust-toolchain@55c7845fad90d0ae8b2e83715cb900e5e861e8cb
    with:
        toolchain: 1.64.0
```

which seems less ideal.

I would prefer the `@1.64.0` syntax.

sumerman: That works for me 👍

JamesGuthrie: done

sumerman (approved):

PR #568, opened 2022-11-22
Use pgtap parser to interpret test results where possible

JamesGuthrie:
## Description

Until now, we have used the snapshot mechanism for pgTAP tests. This
works, but adds additional steps to creating or modifying a test.

This change adds a minimal pgTAP test output parser, and applies it to
the output of our sql tests, falling back to the snapshot mechanism if
the test output is not identified as containing pgTAP output.

## Merge requirements


sumerman (approved):
❤️ 💜 🧡

PR #575, opened 2022-12-02
WIP arm64

JamesGuthrie:
## Merge requirements


JamesGuthrie:
I opened this change as a PR to the `promscale_extension_private` repo as well: https://github.com/timescale/promscale_extension_private/pull/1, which illustrates that the build works there with these changes. Note: The idea is not that we will open PRs against that repo for builds, but that by pushing a tag to that repository, we can get the package built for a tag. We could potentially automate this step in the future.

JamesGuthrie:
It would also be nice to automate the tag push from `promscale_extension` to the `promscale_extension_private` repo. There is even a github action which does this: https://github.com/cpina/github-action-push-to-another-repository.

PR #576, opened 2022-12-02
Compression Job Improvements

jgpruitt:
## Description

Adds a view named metric_chunks_that_need_to_be_compressed which only returns results for metrics that have chunks which need to be compressed. It returns one row per metric with a jsonb array of objects listing the chunks to be compressed.

The execute_compression_policy previously looped over EVERY metric, grabbed an advisory lock on it, and then checked for chunks to be compressed for that metric. This new version uses the view above. In this way, it only loops over and locks metrics which have chunks to compress. Further, it does already has the list of chunks to compress. This reduces locking, transactions, and queries.

## Merge requirements


migration/idempotent/001-base.sql
```diff
@@ -3384,73 +3384,193 @@ $$
...
+            AND c.dropped IS FALSE
+            AND (c.status OPERATOR(pg_catalog.&) 1) OPERATOR(pg_catalog.!=) 1 -- not yet compressed
+        ) x
+    ) c ON (c.chunks_to_compress OPERATOR(pg_catalog.!=) '[]'::jsonb) -- don't bother with metrics that have no chunks to compress
+    ;
+    GRANT SELECT ON _prom_catalog.metric_chunks_that_need_to_be_compressed TO prom_maintenance;
+END;
+$block$;
+
+CREATE OR REPLACE FUNCTION _prom_catalog.compress_chunk(uncompressed_chunk regclass, if_not_compressed boolean DEFAULT false)
```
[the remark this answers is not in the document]
jgpruitt: ok. in that case, I'll just use `_prom_catalog.compress_chunk_for_hypertable`. I was hoping to avoid the extra query of the catalog, but it should not be a big deal.

migration/idempotent/001-base.sql
```diff
@@ -3384,73 +3384,193 @@ $$
 LANGUAGE PLPGSQL STABLE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.get_metrics_that_need_compression() TO prom_maintenance;
 
+DO $block$
+BEGIN
+    IF NOT _prom_catalog.is_timescaledb_installed() THEN
```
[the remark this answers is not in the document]
jgpruitt: No. The way the devenv is setup prevents this apparently. It's highly annoying IMO.

migration/idempotent/001-base.sql
```diff
@@ -3384,73 +3384,193 @@ $$
...
 
+        _metrics_compressed = _metrics_compressed OPERATOR(pg_catalog.+) 1;
         COMMIT;
         -- reset search path after transaction end
         SET LOCAL search_path = pg_catalog, pg_temp;
     END LOOP;
+
+    _job_duration_min = extract(epoch from pg_catalog.clock_timestamp() OPERATOR(pg_catalog.-) _job_start)::double precision / 60.0;
+    IF log_verbose AND _job_duration_min != 0.0 THEN
+        RAISE LOG 'promscale maintenance: compression: finished in %s minutes. %s metrics | %s chunks | %s metrics per minute | %s chunks per minute',
```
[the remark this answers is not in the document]
jgpruitt: switched to INFO

migration/idempotent/001-base.sql
```diff
@@ -3384,73 +3384,181 @@ $$
...
+                    'table_name', x.table_name,
+                    'range_end', x.range_end,
+                    'chunk_num', x.chunk_num
+                )
+                ORDER BY x.chunk_num DESC -- oldest chunks first
+            )
+            FILTER
+            (
+                WHERE x.chunk_num OPERATOR(pg_catalog.>) 1 -- don't compress the most recent chunk
+                AND x.range_end OPERATOR(pg_catalog.<=) now() - interval '1 hour' -- don't compress chunks that aren't at least 1 hour old
```
cevian, line 3432: hmm if this was a function and not a view the 1 hour interval could be a parameter. We might need this feature

jgpruitt:
On a test system with the below stats, the query in `_prom_catalog.metric_chunks_that_need_to_be_compressed` generates the following explain plan

2,000 metrics
200,000 series
44,010 chunks

https://explain.dalibo.com/plan/a36861e6g71c741d

jgpruitt:
On a test system with the below stats, the query in _prom_catalog.metric_chunks_that_need_to_be_compressed generates the following explain plan

2,500 metrics
250,000 series
1,496,370 chunks

https://explain.dalibo.com/plan/hd818fca8d52c9c8

migration/idempotent/001-base.sql
jgpruitt: As a view, I can use the view's type to declare variable in the execute_compression_policy.
As far as i know, there is no way to declare variables of a type referring to the return type of a function.
I can use record for a variable type, but a `record[]` is not allowed. Thus, I have to use a temp table to accumulate the metrics that we cannot lock without waiting.

jgpruitt: I guess I **could use a variable of type jsonb... that feels kinda gross though. WDYT?

migration/idempotent/001-base.sql
```diff
@@ -3384,73 +3384,203 @@ $$
 LANGUAGE PLPGSQL STABLE;
 GRANT EXECUTE ON FUNCTION _prom_catalog.get_metrics_that_need_compression() TO prom_maintenance;
```
sumerman, line 3385: Shouldn't we get rid of it?

sumerman (approved):

dev.Dockerfile
```diff
@@ -27,7 +27,7 @@ RUN chmod a+w /usr/share/postgresql/*/extension /usr/lib/postgresql/*/lib
 USER ubuntu
 
 # Install rust
-ENV RUST_VERSION=1.62.0
+ENV RUST_VERSION=1.64.0
```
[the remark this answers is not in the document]
jgpruitt: yeah, weird. fixed

migration/idempotent/001-base.sql
jgpruitt: let me see if i can tell if it's used anywhere else...

jgpruitt: removing...

PR #581, opened 2022-12-06
Don't cache target dir

jgpruitt:
## Description

Tell Swatinem/rust-cache not to cache the target directory. It's causing issues in CI. All credit to @JamesGuthrie 

## Merge requirements


sumerman:
Could you elaborate on why is this necessary in addition to #577 ?

PR #584, opened 2022-12-08
Make cargo use 2 jobs in devenv

jgpruitt:
## Description

We have identified issues with `make devenv` and `make docker-image-14` failing due to a signal 9 when cargo is building the `pgx-pg-sys` package. We believe this is due to RAM usage exceeding what docker is allowed. This PR limits cargo to using 2 jobs which in turn ought to reduce the RAM used.

## Merge requirements

### DIFF
--- a/dev.Dockerfile
+++ b/dev.Dockerfile
@@ -51,7 +51,10 @@ RUN sudo apt-get install -y vim lld
 RUN mkdir -p ~/.cargo
 # Make cargo put compile artifacts in non-bind-mounted directory
 # To re-use compiled artifacts, mount a docker volume to /tmp/target
-RUN echo -e '[build]\ntarget-dir="/tmp/target"' > ~/.cargo/config.toml
+# We have seen issues with the docker container running out of memory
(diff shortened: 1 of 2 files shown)


sumerman (approved):

dev.Dockerfile
```diff
@@ -51,7 +51,7 @@ RUN sudo apt-get install -y vim lld
 RUN mkdir -p ~/.cargo
 # Make cargo put compile artifacts in non-bind-mounted directory
 # To re-use compiled artifacts, mount a docker volume to /tmp/target
-RUN echo -e '[build]\ntarget-dir="/tmp/target"' > ~/.cargo/config.toml
+RUN echo -e '[build]\ntarget-dir="/tmp/target"\njobs=2' > ~/.cargo/config.toml
```
cevian, line 54: can we add a comment above for why jobs =2

cevian (approved):

dev.Dockerfile
jgpruitt: done.

PR #587, opened 2022-12-09
Clean generated files on devenv (re)builds

### LINKED ISSUE #585: Devenv fails to use changes from source
Say you are working on 0.7.1-dev. If you have run `make devenv`, and have run something on the host machine like `make build` that generates `sql/promscale--0.7.1-dev.sql`, then the devenv starts using the sql file from the host's sql dir instead of the one that is generated from source. When this happens, you make changes to the source code, but it does not get updated in the devenv despite rebuilds happening. This can be incredibly confusing/frustrating. What can we do here to make this footgun go away?

sumerman:
## Description

Addressing a devenv footgun documented in https://github.com/timescale/promscale_extension/issues/585
- Firstly, I don't see a reason for make build to generate a versioned SQL
  file in the first place. This action normally takes place during
  the packaging step.
- Secondly, make clean and make devenv targets will be removing generated
  files moving forward.
  
I didn't test this very exhaustively but it seems to help and I expect it to resolve #585
  
## Merge requirements

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,6 +7,7 @@ H := \#
 # not available for sudo, so we use `command -v` to only run them if they are available
 EXT_VERSION ?= $(shell command -v cargo >/dev/null && ./extract-extension-version.sh | tr -d '\n')
 RUST_VERSION ?= $(shell command -v rustc >/dev/null && rustc --version | cut -d' ' -f2)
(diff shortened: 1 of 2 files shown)


cevian (approved):
This seems right to me but would want some more eyes on it

jgpruitt (approved):
Gave it a spin. Seems to remove the foot-guns.

PR #588, opened 2022-12-09
Increase the default number of compression jobs and prevent a "thundering herd" scenario

sumerman:
## Description

It seems like 1 metric compression job is almost never enough and our benchmarks show that 3 should be sufficient for thousands of metrics with 30min chunks.

In addition, this PR addresses a thundering herd scenario when reconfiguring maintenance jobs.

## Merge requirements


jgpruitt (approved):
Took me a minute to decode what all the random() math does ;). Might be worth some comments, but it LGTM!

PR #589, opened 2022-12-13
Support running devenv without entr

jgpruitt:
## Description

There may be situations when manually triggering rebuilds and reinstalls of the extension in the devenv container is preferred. This provides a `make devenv-no-entr` target to disable the automatic rebuilds/reinstalls.

The new `make dev-build` is used to trigger the rebuild/reinstall manually.

`make dev-bash` is handy to get a bash shell in the devenv container.

`make dev-psql` gets a psql shell into the devenv container.

## Merge requirements

### DIFF
--- a/devenv.sh
+++ b/devenv.sh
@@ -37,13 +37,20 @@ for db in template1 postgres; do
 done
 createdb -h localhost -p "${PGPORT}" "$(whoami)"
 
-# This allows entr to work correctly on docker for mac
-export ENTR_INOTIFY_WORKAROUND=true
+if [ "$DEVENV_ENTR" -eq 1 ]; then
+  echo "entr enabled. rebuilds will be triggered on source file changes"
+  # This allows entr to work correctly on docker for mac
+  export ENTR_INOTIFY_WORKAROUND=true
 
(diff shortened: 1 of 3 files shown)


devenv.sh
```diff
@@ -37,13 +37,20 @@ for db in template1 postgres; do
...
+  # This allows entr to work correctly on docker for mac
+  export ENTR_INOTIFY_WORKAROUND=true
+
+  # Note: this is not a comprehensive list of source files, if you think one is missing, add it
+  SOURCE_FILES="src migration"
+  find ${SOURCE_FILES} | entr make devenv-internal-build-install > "${HOME}/compile.log" &
+
+  tail -f "${HOME}/.pgx/${DEVENV_PG_VERSION}.log" "${HOME}/compile.log" &
+else
+  echo "entr disabled. you must trigger rebuilds manually"
```
sumerman, line 51: I'd mention `make dev-build`

sumerman (approved):

devenv.sh
jgpruitt: good idea

PR #593, opened 2022-12-15
Analyze chunks after compression

jgpruitt:
## Description

We need to analyze chunks right after we compress them. If we don't, `n_live_tup` will be zero, and the autovacuum engine will ignore the chunk.

## Merge requirements


sumerman (approved):

jgpruitt:
Upon further investigation, I don't think this will actually fix the issue we are trying to fix. AFAICT:

- if the stats are missing, the table won’t be vacuumed until we get into anti-wraparound territory
- if the stats are missing, running analyze won’t help unless there is subsequent DML to trigger a vacuum

